PR #17, opened 2014-12-30
Fixed "all" Task to allow plugin baking.

DSchalla:
Calling "bake" directly in the Tasks skips the $this->_getName(); function call, which separates Plugin from Name. Fix is to call main(); instead of bake();.

### DIFF
--- a/tests/comparisons/Controller/testBakeWithPlugin.php
+++ b/tests/comparisons/Controller/testBakeWithPlugin.php
@@ -22,6 +22,7 @@ class BakeArticlesController extends AppController
             'contain' => ['BakeUsers']
         ];
         $this->set('bakeArticles', $this->paginate($this->BakeArticles));
+        $this->set('_serialize', ['bakeArticles']);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


tests/comparisons/Controller/testBakeWithPlugin.php
```diff
@@ -58,6 +60,7 @@ public function add()
         $bakeUsers = $this->BakeArticles->BakeUsers->find('list', ['limit' => 200]);
         $bakeTags = $this->BakeArticles->BakeTags->find('list', ['limit' => 200]);
         $this->set(compact('bakeArticle', 'bakeUsers', 'bakeTags'));
+        $this->set('_serialize', ['bakeArticle', 'bakeUsers', 'bakeTags']);
```
lorenzo: I don't think associations should be serialized

DSchalla:
Will fix that the next days, what you would recommend instead in the task? Not baking the compact but only the main stuff?

lorenzo:
Thanks!

PR #25, opened 2015-01-06
Add empty entry on nullable fields that will render as select inputs

michha:
It was painfull to manually add the empty option to the baked views. So I added code that would check the schema and if the field is nullable, the empty option is added.

### DIFF
--- a/src/Template/Bake/Element/form.ctp
+++ b/src/Template/Bake/Element/form.ctp
@@ -59,12 +59,11 @@ $fields = collection($fields)
             }
             if (isset($keyFields[$field])) {
                 $fieldData = $schema->column($field);
-                if (array_key_exists('null', $fieldData) && $fieldData['null'] == 1) {
+                if (!empty($fieldData['null'])) {
 %>
             echo $this->Form->input('<%= $field %>', ['options' => $<%= $keyFields[$field] %>, 'empty' => true]);
 <%
-                }
-                else {
+                } else {
 %>
             echo $this->Form->input('<%= $field %>', ['options' => $<%= $keyFields[$field] %>]);
 <%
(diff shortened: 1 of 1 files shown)


dereuromark:
Sounds reasonable to me.

src/Template/Bake/Element/form.ctp
```diff
@@ -58,9 +58,17 @@ $fields = collection($fields)
                 continue;
             }
             if (isset($keyFields[$field])) {
+                $fieldData = $schema->column($field);
+                if (array_key_exists('null', $fieldData) && $fieldData['null'] == 1) {
```
markstory: You can use !empty() instead here.

src/Template/Bake/Element/form.ctp
```diff
@@ -58,9 +58,17 @@ $fields = collection($fields)
...
+%>
+            echo $this->Form->input('<%= $field %>', ['options' => $<%= $keyFields[$field] %>, 'empty' => true]);
+<%
+                }
+                else {
```
markstory: Should be `} else {`.

michha:
Fixed both issues

lorenzo:
Good job, thanks!

dereuromark:
Actually, even for not NULL fields it is bad to pre-select. Instead the text should be slightly different.
I have always been using this for the last years: https://github.com/dereuromark/cakephp-setup/blob/master/Console/Templates/setup/views/form.ctp#L52-L55

jhli:
I do agree. Two ways to go:
- Don't pre-select and blank for NULL and NOT NULL
- Don't pre-select and different messages for NULL and NOT NULL. For NOT NULL it should say something like "Please select..." and for not null something like "You may select..." 

I am fine with both but pre-selecting NULL foreign keys is bad from a utility point of view and pre-selecting NULL and NOT NULL values is bad from an ergonomics point of view (users get confused if they already chose something or it it was pre-selected)

PR #26, opened 2015-01-10
Fixed Plugin Task to deny option of ROOT/plugins and fixed Autoloading Generation

DSchalla:
PR related to the change: https://github.com/cakephp/app/pull/206

The first commit is filtering the ROOT/plugins path from the options and exits with an error if no other path is available, I am unsure if exit; was the correct choice in a Shell Task, I would like to get some feedback regarding that.

The second commit fixed the Autoloading Generation to allow other than the default path options to be used as plugin path, the chdir adaption was required since the old version could not deal with anything but the first child of ROOT.

Any comments are appreciated.

### DIFF
--- a/src/Shell/Task/PluginTask.php
+++ b/src/Shell/Task/PluginTask.php
@@ -215,9 +215,11 @@ class PluginTask extends BakeTask
             return false;
         }
 
+        $autoloadPath = str_replace(ROOT, '.', $this->path);
+
         $config = json_decode(file_get_contents($file), true);
-        $config['autoload']['psr-4'][$plugin . '\\'] = "./plugins/$plugin/src";
-        $config['autoload-dev']['psr-4'][$plugin . '\\Test\\'] = "./plugins/$plugin/tests";
+        $config['autoload']['psr-4'][$plugin . '\\'] = $autoloadPath . $plugin . "/src";
+        $config['autoload-dev']['psr-4'][$plugin . '\\Test\\'] = $autoloadPath . $plugin . "/tests";
 
         $this->out('<info>Modifying composer autoloader</info>');
 
@@ -235,7 +237,7 @@ class PluginTask extends BakeTask
             $cwd = getcwd();
 
             // Windows makes running multiple commands at once hard.
-            chdir(dirname($path));
+            chdir(ROOT);
             $command = 'php ' . escapeshellarg($composer) . ' dump-autoload';
             $this->callProcess($command);
 
@@ -284,7 +286,7 @@ class PluginTask extends BakeTask
         }
 
         if ($max === 1) {
-            $this->path=$pathOptions[0];
+            $this->path = $pathOptions[0];
             return;
         }
 
--- a/src/Shell/Task/PluginTask.php
+++ b/src/Shell/Task/PluginTask.php
@@ -273,7 +273,7 @@ class PluginTask extends BakeTask
     {
         $valid = false;
         foreach ($pathOptions as $i => $path) {
-            if (!is_dir($path) || $path == ROOT . DS . 'plugins' . DS) {
+            if (!is_dir($path)) {
                 unset($pathOptions[$i]);
             }
         }
--- a/src/Shell/Task/PluginTask.php
+++ b/src/Shell/Task/PluginTask.php
@@ -237,7 +237,7 @@ class PluginTask extends BakeTask
             $cwd = getcwd();
 
             // Windows makes running multiple commands at once hard.
-            chdir(ROOT);
+            chdir(dirname($this->_rootComposerFilePath()));
             $command = 'php ' . escapeshellarg($composer) . ' dump-autoload';
             $this->callProcess($command);
 
--- a/src/Shell/Task/PluginTask.php
+++ b/src/Shell/Task/PluginTask.php
@@ -282,7 +282,7 @@ class PluginTask extends BakeTask
 
         if ($max == 0) {
             $this->err('No valid path found!');
-            exit;
+            throw new \RuntimeException();
         }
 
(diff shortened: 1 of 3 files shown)


DSchalla:
I am unsure how to handle the test failures, could it be that the ROOT constant is not defined in the Tests which lead to the failure? Is the usage of the ROOT variable not fine or do I simply have to rename the paths in the testcases?

dakota:
I don't think it's a good idea to deny baking of plugins into `ROOT/plugins`. I often create plugins with bake which I then publish via packagist or a local satis install.

I don't see what's wrong with baking plugins into `ROOT/plugins`?

DSchalla:
@dakota Did you read the PR linked at top?

dakota:
I did, I still don't think it's a good idea blocking baking into `ROOT/plugins`. Perhaps giving a warning would be the solution?

DSchalla:
A warning or maybe a option flag could be used, like --composer or something like that.

ADmad:
> I often create plugins with bake which I then publish via packagist or a local satis install.

@dakota IMHO you should maintain separate folders for developing / published and using in an app.

dakota:
@ADmad you're right, I'm must being lazy :)

jadb:
@dakota you can still be lazy, this PR won't stop you ;) 

And to suit your laziness, here's the new path ready for you to copy/paste:

```
cd src/Plugin
```

markstory:
I don't think preventing people from putting code where they want is a good idea. What if there was an output-dir option that defaulted to the new path? This would let people output plugin skeletons for standalone plugins as well.

lorenzo:
I like that idea @markstory

DSchalla:
Then we can just remove the filter again and the app.default.php has to be adapted, since this controls how the paths are ordered when multiple exist. Or did I misunderstood something?

@lorenzo, any hints regarding the tests?

src/Shell/Task/PluginTask.php
```diff
@@ -235,7 +237,7 @@ protected function _modifyAutoloader($plugin, $path)
             $cwd = getcwd();
 
             // Windows makes running multiple commands at once hard.
-            chdir(dirname($path));
+            chdir(ROOT);
```
lorenzo: why did this change?

DSchalla:
When using a different path than ROOT/plugins you would not switch back correct to the root folder. As example, ROOT/src/Plugin would switch back to ROOT/src, leading to a failure for the composer call.

src/Shell/Task/PluginTask.php
AD7six: It looks to be logically correct to make this change; `dirname($path)` means `dirname('plugins/`)` - it's implicitly assuming the plugins dir is one level below the root. 

However, I'd rather it was:

    $file = $this->_rootComposerFilePath(); # <-
    ...
    chdir(dirname($file));

There's no need to use a constant.

DSchalla: Indeed, the changes was related to that issue. Regarding _rootComposerFilePath(), that would be  an unnecessary function call since the function is also just:
"return ROOT . DS . 'composer.json';"

However, if requested I can ofc change it to use the function call. dirname($file) is sufficient, since we do not want to step out of the root, hu?

AD7six: Please consider testing - you _don't_ want to be modifying `ROOT . DS . 'composer.json` when tests run and that's why it's in a separate function (and why generally I purged constants from the code recently).

DSchalla: That's totally fine, I will update the code to use the function tomorrow. Thanks for your advice @AD7six

josegonzalez:
@lorenzo bump

tests/comparisons/Plugin/testMainUpdateComposer.json
```diff
@@ -1,12 +1,12 @@
...
     "autoload": {
         "psr-4": {
-            "ComposerExample\\": "./plugins/ComposerExample/src"
+            "ComposerExample\\": "/tmp/tests/BakedPlugins/ComposerExample/src"
         }
     },
     "autoload-dev": {
         "psr-4": {
-            "ComposerExample\\Test\\": "./plugins/ComposerExample/tests"
+            "ComposerExample\\Test\\": "/tmp/tests/BakedPlugins/ComposerExample/tests"
```
lorenzo: Paths should be relative as when you deploy the path will probably not be the same.

DSchalla: The path used above is related to the fact that the plugin path in the Test Instance is set to /tmp/tests/BakedPlugins, on a productive environment it would look like:
'"Test6\\Test\\": "./src/Plugin/Test6/tests"

lorenzo:
Thanks a lot for the effort @DSchalla, the team decided to go another way with this. Instead of disallowing baked plugins in ROOT, we decided to put them only there and have composer install plugins in vendors. That way there will be a clear separation between both, all thrid-party libraries will be in the same place and no gitignore will be required for in-app plugins.

I'm sorry you had to work on this and not get this merged, you win 100 internets from me :muscle:

lorenzo:
Reopening as this PR was transformed to fix a couple limitations on how plugins are baked. The problem in tests still need to be solved

DSchalla:
The issues with the tests should be fixed after https://github.com/DSchalla/bake/commit/48a284bb4fe6211e0d0665b03a544f78844c9e63 and https://github.com/DSchalla/bake/commit/475d5f3c71c957474dd633e39da8397edc91f6c7.

markstory:
Thanks :+1:

DSchalla:
Thanks @markstory for the merge!

PR #31, opened 2015-01-25
Do not attempt to retrieve the template file name twice

josegonzalez:


src/View/BakeView.php
```diff
@@ -187,12 +187,11 @@ protected function _evaluate($viewFile, $dataForView)
...
      * Get the contents of the template file
      *
-     * @param string $name A template name or a Bake template string
+     * @param string $filename A template filename
      * @return string Bake template to evaluate
      */
-    protected function _getViewFileContents($name)
+    protected function _getViewFileContents($filename)
     {
-        $filename = $this->_getViewFileName($name);
```
markstory: Any tests that could be updated, or were they failing previously?

josegonzalez: Thats the thing, tests ran fine with/without this change, but without this change, it seemed that we were grabbing the name twice. I haven't thought of a thoughtful reproduction yet :(

markstory:
Ok, hopefully we can get figure out a way to catch this issue in the future.

src/View/BakeView.php
AD7six: If I ever git-bisect to this commit I will have no real idea why this commit was made.

Was `_getViewFileName` returning the passed input when it's a filepath, except in some circumstances or something? I don't pretend the previous code was right - but knowing in what way the previous code caused a problem would help understand "the problem".

josegonzalez: The method always returns a full path - or appeared to in the migrations plugin. It is first called in `BakeView::render()` - same as in `View::render()` - and this name is eventually passed into `BakeView::_evaluate()`. `View::_evaluate()` will simply use this line as is to get the file contents, while `BakeView` was using it to *again* interpolate the filename before inclusion.

As soon as it is passed in the second time, it includes the extension, so the second time around it throws an exception because it'll be something like `file.ctp.ctp`, which won't exist.

I'm actually not really sure how this worked at all before, but removing the second attempt and making that act as `View` seems to work fine. Like I said, I dug into what was happening, but not why tests were failing, hence why I left it open as is and din't immediately merge.

Issue #37, opened 2015-02-08
problem with hierarchy of namespace in plugin name


zoghal:
when use problem with hierarchy of namespace in plugin name, baker generating incorrect any class of need used namespace
for example: plugin name is Rita/Users, baker generate entity class Similar this :
```
<?php
/* /Plugin/Rita/Users/src/Model/Entity/User.php */
namespace Rita/Users\Model\Entity;

use Cake\ORM\Entity;
```

markstory:
I can take a look at this.

DSchalla:
https://github.com/cakephp/bake/blob/master/src/Shell/Task/ModelTask.php#L642

The Namespace must be adjusted in case its a Plugin with vendor prefix, at least that is what seems to be the issue. A str_replace should fix that.

markstory:
Closing as #38 has been merged.

PR #38, opened 2015-02-09
fix issuse #37 

zoghal:
fix issuse #37

### DIFF
--- a/src/Shell/Task/TestTask.php
+++ b/src/Shell/Task/TestTask.php
@@ -201,7 +201,7 @@ class TestTask extends BakeTask
 
         $baseNamespace = Configure::read('App.namespace');
         if ($this->plugin) {
-            $baseNamespace = str_replace('/', '\\', $this->plugin);
+            $baseNamespace = $this->_pluginNamespace($this->plugin);            
         }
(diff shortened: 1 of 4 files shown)


src/Shell/Task/TestTask.php
```diff
@@ -201,7 +201,7 @@ public function bake($type, $className)
 
         $baseNamespace = Configure::read('App.namespace');
         if ($this->plugin) {
-            $baseNamespace = str_replace('/', '\\', $this->plugin);
+            $baseNamespace = $this->_pluginNamespace($this->plugin);
```
bcrowe: Trailing whitespace.

antograssiot:
not related but could you also update the composer.json like this so tests passes : 
```
"cakephp/cakephp-codesniffer": "dev-master"
```

I thought the 2.x tag existed at some time...

markstory:
@antograssiot No we never did make that tag. I'll add one now for 2.0.0 :smile:

src/Shell/Task/TestTask.php
markstory: Fixed in eb9ff7daa9401e0ee6f95c060b817153ad217994

PR #41, opened 2015-02-11
The app folder name is not a namespace

AD7six:
There is no restriction on what a project root folder can be named


antograssiot:
:+1:

PR #42, opened 2015-02-11
Don't assume there will always be a bootstrap file

ADmad:
If plugin does not have a bootstrap file load the core tests bootstrap.


jadb:
:+1: 

nice one!

PR #43, opened 2015-02-11
RFC: Remove the project task

AD7six:
When bake was part of cakephp/cakephp - it arguably made sense. But now
that it's a seperate plugin I don't think so.

To use the project task, users would need to EITHER:

    composer create-project cakephp/app whatever
    cd whatever
    bin/cake bake project /another/project

OR

    download a zipball
    cd whatever
    bin/cake bake project /another/project

Instead of just:

    composer create-project cakephp/app /another/project

the project task wouldn't work anyway if composer wasn't installed, so
it doesn't provide that crutch to users anyway.


antograssiot:
:+1: for removing this

ADmad:
:+1:

rchavik:
:+1:

dakota:
:+1:

DSchalla:
Makes perfect sense.

PR #45, opened 2015-02-11
add  .gitattributes file

antograssiot:


.editorconfig
```diff
@@ -10,3 +10,7 @@ indent_style = space
 indent_size = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.yml]
+indent_style = space
+indent_size = 2
```
jadb: Is there a reason for repeating the `indent_style`?

Also, why not using something like `[*.{json,yml}]` since I believe it applies to both?

dereuromark: @jadb only .yml require 2 space. Json is 4 space as normal.

antograssiot: composer.json  fo example have 4 spaces I think...

jadb: I see 2 spaces being used almost everywhere for json:

https://docs.npmjs.com/misc/coding-style

For example, `bower.json` is 2 spaces as well. 

I noticed noticed now composer init creates it with 4 spaces, not sure why though.

ADmad: @jadb PHP's json_encode() with pretty print option indents using 4 spaces.

markstory:
Looks good to me. We can always fiddle with the different file types in future pull requests.

Issue #2, opened 2014-12-12
Restructuring


AD7six:
From https://github.com/cakephp/cakephp/issues/5319, here's a list of things that aught to be done before a stable release:

 - [x] Convert to a plugin
 - [x] PSR2
 - [ ] Allow baking anything
 - [ ] Restructure to be not tasks, but event based
 - [x] Normalize view/template terminology
 - [x] Baking a plugin should point at a template folder (in a bake theme probably) and recursively render files
 - [ ] Rewrite/improve templates, using helper methods elements etc.

davidyell:
:+1: For event driven bake

josegonzalez:
@AD7six can this be closed?

PR #46, opened 2015-02-12
Rename view usage/references to template

AD7six:
The current Template task, is renamed BakeTemplate; the current view task is renamed Template task.

Ref: https://github.com/cakephp/bake/issues/2


AD7six:
@josegonzalez Need to coordinate merging this one ot prevent migrations breaking.

markstory:
It would be nice to leave a stub ViewTask so people can find out what the new command name is.

AD7six:
I'm not sure that's a good idea.

You can't right now bake a view class, and unless the syntax changes, that would block adding that possibility.

AD7six:
I can do this if we do want to point users in the right direction:

```
diff --git src/Shell/BakeShell.php src/Shell/BakeShell.php
index ecc6845..d35de60 100644
--- src/Shell/BakeShell.php
+++ src/Shell/BakeShell.php
@@ -72,6 +72,16 @@ class BakeShell extends Shell
      */  
     public function main()
     {   
+        if ($this->args && $this->args[0] === 'view') {
+            $this->out('<error>The view command has been renamed</error>');
+            $this->out('To create template files, use the template task:', 2);
+            $args = $this->args;
+            array_shift($args);
+            $args = implode($args, ' ');
+            $this->out(sprintf('    bin/cake bake template %s', $args), 2);
+            return false;
+        }
+
         $connections = ConnectionManager::configured();
         if (empty($connections)) {
             $this->out('Your database configuration was not found.');
```

markstory:
@AD7six I think that warning is enough.

antograssiot:
:+1: 
I've done some docs update proposal regarding this modification.

AD7six:
@antograssiot awesome, thanks!

PR #48, opened 2015-02-17
- include 'empty' => true, 'default' => '' for date DEFAULT NULL fields

renardrouge:
Passed all current tests (11) under /bake

### DIFF
--- a/src/Template/Bake/Element/form.ctp
+++ b/src/Template/Bake/Element/form.ctp
@@ -71,9 +71,17 @@ $fields = collection($fields)
                 continue;
             }
             if (!in_array($field, ['created', 'modified', 'updated'])) {
+                $fieldData = $schema->column($field);
+                if (($fieldData['type'] === 'date') && (!empty($fieldData['null']))) {
+%>
+            echo $this->Form->input('<%= $field %>', array('empty' => true, 'default' => ''));
+<%
+                } else {
 %>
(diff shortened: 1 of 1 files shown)


src/Template/Bake/Element/form.ctp
```diff
@@ -71,9 +71,17 @@ $fields = collection($fields)
...
+                if (($fieldData['type'] === 'date') && (!empty($fieldData['null']))) {
+%>
+            echo $this->Form->input('<%= $field %>', array('empty' => true, 'default' => ''));
+<%
+                } else {
 %>
             echo $this->Form->input('<%= $field %>');
 <%
+                }
+                continue;
```
markstory: Why continue?

renardrouge: Modeled after block just previous, beginning    if (isset($keyFields[$field])) {

markstory: ok.

renardrouge: Hey Mark !
Something is not right.  That block of code in your "Why continue" question is not the same as the code in github.com/renardrouge/bake .

renardrouge: Oops !

I don't know what just happened.  I refreshed the screen, now the quoted code is the same as mine.

dereuromark:
This seems to be missing for datetime, though.

PR #50, opened 2015-02-25
Allow overriding plugin bake tasks within an app

josegonzalez:
**This PR is strictly for discussion. I'm not convinced this is the best way to perform what I want to do.**

This gives developers the ability to sub-class bake shells to give them new abilities.

As an example, a developer may wish to subclass the FixtureTask to disable generating certain tables, or use a custom method to generate data for the fixtures. Without the above change, they would be forced to use a different name for the task.

### DIFF
--- a/src/Shell/BakeShell.php
+++ b/src/Shell/BakeShell.php
@@ -127,7 +127,6 @@ class BakeShell extends Shell
     {
         $tasks = [];
 
-        $tasks = $this->_findTasks($tasks, APP, Configure::read('App.namespace'));
         foreach (Plugin::loaded() as $plugin) {
             $tasks = $this->_findTasks(
                 $tasks,
@@ -136,6 +135,7 @@ class BakeShell extends Shell
                 $plugin
             );
         }
+        $tasks = $this->_findTasks($tasks, APP, Configure::read('App.namespace'));
 
         $this->tasks = array_values($tasks);
         parent::loadTasks();


lorenzo:
With this, how would you use your own Fixture task? Just by having the class in your App ?

markstory:
This seems pretty reasonable to me.

josegonzalez:
Yeah, basically. One thing I noticed is that this wouldn't affect other bake tasks that load Fixture. So for instance, it doesn't override fixture baking that is done when you bake a model. That would need a similar alteration in BakeTask::loadTasks() I think?

markstory:
@josegonzalez You'd have to do some change as currently the various tasks in Bake use `Bake.Fixture` for example.

josegonzalez:
Right, just asking if that is okay with everyone. I'd love to be able to shadow plugin/core tasks now that we have namespaces, which makes the handling pretty easy to do, but want to be sure everyone else is on board.

jadb:
:star2:

markstory:
Before bake was a plugin shadowing was super easy, now it's a bit harder.

lorenzo:
Does this work or anything else in needed to be done?

josegonzalez:
This doesn't work for sub-tasks that load their own tasks, with my Model->Fixture example being a case where it does not. As well, I'm not sure how I'd write the tests for this.

burzum:
Why is this still open and not merged?

markstory:
@burzum Sounds like there were still issues with tasks loaded inside other tasks.

burzum:
@markstory so we should either fix it or close the PR this has been sleeping for almost two years without any feedback.

josegonzalez:
If someone would like to take over the pull request, feel free to. Otherwise, please close.

dereuromark:
Seems like this is not going to happen anymore.

Issue #51, opened 2015-03-03
Bake mistakenly thinks composer dump-autoload failed


sitedyno:
Baking a plugin I see this at the end:
> Found composer executable in /usr/local/bin
> Running php '/usr/local/bin/composer' dump-autoload
> Error: Could not run `composer dump-autoload`: Generating autoload files

Composer does write to stderr: https://github.com/composer/composer/blob/master/src/Composer/Command/DumpAutoloadCommand.php#L60

The autoloader is updated as I can see the namespaces added to vendor/composer/autoload_psr4.php

markstory:
Writes to STDERR, everything OK :cry:

ceeram:
Recently changed in composer: https://github.com/composer/composer/pull/3715

<info> output should not go to stderr :(

2015-03-03 23:34 GMT+01:00 Mark Story <notifications@github.com>:

sitedyno:
The choice to send info to stderr aside, shouldn't the return value of proc_close() be used to determine success or failure?

dzuelke:
Welcome to Unix, where the success or failure of an operation has been indicated through an application's exit code since the 1960s. Output on the standard error stream does not mean an error occurred.

josegonzalez:
@dzuelke Like I said in the composer thread, we're definitely going to change this to not depend upon stderr (it was a bug to do so) but I still think composer should not just arbitrarily write status messages to stderr without some notion of debug/warning/info/etc log levels.

dzuelke:
@josegonzalez Composer is a console app... If you want more verbosity, use `-v`, if you want less, `-q`, that's what most of those do.

IMO it might be useful for "composer install" to write the installed packages to `stdout` at the end as a summary of some kind, but that's another story.

markstory:
@sitedyno @dzuelke Yes, I should have used the sub-processes' exit code. That's the change I was planning on making :smile:

markstory:
Closing as a pull request is up now.

PR #53, opened 2015-03-05
Only fail subcommands when they exit non-zero.

markstory:
Instead of looking in stderr, the exit code of a process should be used to determine whether or not it worked.

Refs #51

### DIFF
--- a/src/Shell/Task/BakeTask.php
+++ b/src/Shell/Task/BakeTask.php
@@ -141,9 +141,9 @@ class BakeTask extends Shell
 
         $error = stream_get_contents($pipes[2]);
         fclose($pipes[2]);
-        proc_close($process);
+        $exit = proc_close($process);
 
-        if ($error) {
+        if ($exit != 0) {
             throw new \RuntimeException($error);
         }
 


src/Shell/Task/BakeTask.php
```diff
@@ -141,9 +141,9 @@ public function callProcess($command)
 
         $error = stream_get_contents($pipes[2]);
         fclose($pipes[2]);
-        proc_close($process);
+        $exit = proc_close($process);
 
-        if ($error) {
+        if ($exit != 0) {
```
dereuromark: Does this need to be a strict check?

AD7six: Can be less strict i.e. `if ($exit)`

dereuromark: I am thinking about false positives when returning a string like "foobar", which would then be true with `!= 0`, while false with strict checks.

AD7six: The return value of proc_close [is an int](http://php.net/manual/en/function.proc-close.php)

dereuromark: Right, so the strict check would be best regarding the guideline "be as as strict as possible".

AD7six: well, it's academic really. We know it's a number and to me this is clearer:

    if ($exit) {

instead of:

    if ($exit !== 0) {

For me that involves mental shuffling whereas "if exit go here" does not.

PR #55, opened 2015-03-13
Filter HasMany associations if already exist as BelongsToMany

steinkel:
Get rid of duplicated HasMany associations already existing as BelongsToMany. Example: ArticlesTags would be removed from the list of associations used for code generation if a Articles BelongsToMany Tags association exists.

### DIFF
--- a/src/Shell/Task/TemplateTask.php
+++ b/src/Shell/Task/TemplateTask.php
@@ -433,7 +433,7 @@ class TemplateTask extends BakeTask
      * Get filtered associations
      * To be mocked...
      *
-     * @param Table $model
+     * @param Table $model Table
      */
     protected function _filteredAssociations(Table $model)
     {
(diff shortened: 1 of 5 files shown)


src/Shell/Task/TemplateTask.php
```diff
@@ -422,46 +423,12 @@ public function getOptionParser()
...
-                    'displayField' => $target->displayField(),
-                    'foreignKey' => $assoc->foreignKey(),
-                    'alias' => $alias,
-                    'controller' => $className,
-                    'fields' => $target->schema()->columns(),
-                ];
-            }
-        }
-        return $associations;
+        return AssociationFilter::filterAssociations($model);
```
markstory: Why a static method? Couldn't this be an instance? Static methods have generally been painful in the past as they cannot be mocked if required.

steinkel:
Static calls were refactored, and some CS fixes applied

PR #56, opened 2015-03-16
Add ability to bake all tests for a given type

dakota:
Allow running on `bin/cake bake tests <type> --all` for generating tests for all classes of the given type. Reason for using a option rather than a argument (`name == 'all'), is to prevent problems with classes called `All`.

### DIFF
--- a/src/Shell/Task/TestTask.php
+++ b/src/Shell/Task/TestTask.php
@@ -94,6 +94,10 @@ class TestTask extends BakeTask
             return $this->outputTypeChoices();
         }
 
+        if ($this->param('all')) {
+            return $this->_bakeAll($type);
+        }
+
         if (empty($name)) {
             return $this->outputClassChoices($type);
         }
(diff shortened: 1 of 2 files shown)


src/Shell/Task/TestTask.php
```diff
@@ -147,6 +151,26 @@ public function outputClassChoices($type)
     }
 
     /**
+     * @param string $type The typename to get bake all classes for.
+     * @return void
+     */
+    protected function _bakeAll($type)
+    {
+        $mapped_type = $this->mapType($type);
```
ADmad: Should be `$mappedTyped`

markstory:
Other than PHPCS errors, this looks good to me. :+1:

src/Shell/Task/TestTask.php
dakota: Fixed now.

Issue #63, opened 2015-03-31
When baking (bin\cake bake all) with "--connection" parameter it throws error when it comes to the Templates (Views)


icaroscherma:
When I separated an App in two databases, if I try to bake something from a db other than the default, it is supposed to run smoothly if I use `--connection CONNECTION_NAME`, but it doesn't, when it reaches in the Templates (Views), it says that:

```
Exception: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'cakeapp_db.timelines' doesn't exist in [C:\cakeapp\vendor\cakephp\cakephp\src\Database\Schema\Collection.php, line 130]

2015-03-31 18:34:07 Error: [Cake\Database\Exception] SQLSTATE[42S02]: Base table or view not found: 1146 Table 'cakeapp_db.timelines' doesn't exist
Stack Trace:
#0 C:\cakeapp\vendor\cakephp\cakephp\src\Database\Schema\Collection.php(96): Cake\Database\Schema\Collection->_reflect('Column', 'timelines', Array, Object(Cake\Database\Schema\Table))
#1 C:\cakeapp\vendor\cakephp\cakephp\src\Database\Schema\CachedCollection.php(65): Cake\Database\Schema\Collection->describe('timelines', Array)

#2 C:\cakeapp\vendor\cakephp\cakephp\src\ORM\Table.php(422): Cake\Database\Schema\CachedCollection->describe('timelines')

#3 C:\cakeapp\vendor\cakephp\bake\src\Shell\Task\TemplateTask.php(273): Cake\ORM\Table->schema()

#4 C:\cakeapp\vendor\cakephp\bake\src\Shell\Task\TemplateTask.php(139): Bake\Shell\Task\TemplateTask->_loadController()

#5 C:\cakeapp\vendor\cakephp\bake\src\Shell\BakeShell.php(243): Bake\Shell\Task\TemplateTask->main('Timelines')

#6 [internal function]: Bake\Shell\BakeShell->all('timelines')

#7 C:\cakeapp\vendor\cakephp\cakephp\src\Console\Shell.php(376): call_user_func_array(Array, Array)

#8 C:\cakeapp\vendor\cakephp\cakephp\src\Console\ShellDispatcher.php(200): Cake\Console\Shell->runCommand(Array, true)

#9 C:\cakeapp\vendor\cakephp\cakephp\src\Console\ShellDispatcher.php(171): Cake\Console\ShellDispatcher->_dispatch()

[...]

PR #64, opened 2015-04-01
Bake defaultConnectionName method for tables.

markstory:
Doing this fixes baking controllers, and templates that use models that connect to non-default sources.

Refs #63


antograssiot:
:+1:

PR #66, opened 2015-04-04
Using inner joins when the belongs to association is mandatory

jrbasso:
If the field is not null and it is an association the join type can be inner. It gives a better performance on database.


lorenzo:
I had that as the default in the association class but was not very welcomed by many. I'm ok to have it in bake since at least it makes it obvious to change.

jrbasso:
@lorenzo Interesting, do you know why people didn't like it?

markstory:
@jrbasso It was because of optional belongsTo associations. You've covered that use case by checking whether or not the column is nullable.

PR #54, opened 2015-03-07
Modified ModelTask::findBelongsTo() to check for table referenced by id field

djamesfar:
Fix for issue #6004

PR #67, opened 2015-04-05
Re-implement finding belongsTo using constraints

markstory:
Re-do the changes from #54 as github can't figure things out anymore, also add tests.


jrbasso:
@markstory It is giving some error in postgres. You probably need to rebase to include the `joinType` change too.

PR #75, opened 2015-04-15
Fix fixtures' _options indentation

bcrowe:


src/Shell/Task/FixtureTask.php
```diff
@@ -265,7 +265,10 @@ protected function _generateSchema(Table $table)
             $content .= "        '_constraints' => [\n" . implode("\n", $constraints) . "\n        ],\n";
         }
         if (!empty($options)) {
-            $content .= "        '_options' => [\n" . implode(', ', $options) . "\n        ],\n";
+            foreach ($options as &$option) {
+                $option = '            ' . $option;
+            }
+            $content .= "        '_options' => [\n" . implode(",\n", $options) . "\n        ],\n";
```
ceeram: Could we not add the additional spaces to the implode? imploding by `"\n            "`?

ceeram:
There are no existing tests for this it seems?

src/Shell/Task/FixtureTask.php
bcrowe: Nope, because the first key/value pair wouldn't get indented.

PR #82, opened 2015-04-30
Validation rule code formatting

davidyell:
**TL;DR**
[This](http://i.imgur.com/7ZVEuYK.png) to [this](http://i.imgur.com/NlQwEm5.png).

**What?**
This change will break up the validation rules from a single large, visually messy, block of code into lines grouped by field.

**Why?**
Trying to visually see which rules apply to which field was very frustrating, because I'm lazy.

### DIFF
--- a/src/Template/Bake/Model/table.ctp
+++ b/src/Template/Bake/Model/table.ctp
@@ -82,7 +82,12 @@ class <%= $name %>Table extends Table
         $validator
 <% $validationMethods = []; %>
 <%
+$firstField = true;
 foreach ($validation as $field => $rules):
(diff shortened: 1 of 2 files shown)


ADmad:
:+1:

lorenzo:
You nee to update the tests. I also have a suggestion. What about making it look like this:

```php
$validator
    ->allowEmpty('title', false)
    ->add('title',...);

$validator
    ->allowEmpty('price', false)
    ->add('price',...);
```

davidyell:
Ah yes, I forgot about the tests!

Issue #57, opened 2015-03-18
CakePHP 3.0.0RC2 when baking to a plugin, bake controller doesn't find model's associations.


djamesfar:
To reproduce:
bake a plugin:
```
bin\cake bake plugin Pluggy
```
bake a model in the plugin, where the table has belongsTo or hasMany associations (ie Roles):
```
bin\cake bake model --plugin Pluggy --no-test --force roles
bin\cake bake model --plugin Pluggy --no-test --force users
```
now bake the UsersController:
```
bin\cake bake controller --plugin Pluggy --no-test --force users
```
Check the UsersController::add() method and you won't see the find for the associated belongsTo table:
```
$roles = $this->Users->Roles->find('list',...etc)     <--- *this is missing*
```
so it doesn't set(compact('roles'..  for the view.

If you take out the _--plugin Pluggy_,  it will work correctly, creating the controller in _src/controller_, of course, not in _plugins/Pluggy/src/controller_.

I dug around a little in the add.ctp element template used by the controller template and inserted some debug logging after the $associations array is populated using Bake->aliasExtractor (around line 36), and it shows the array empty when baking to a plugin, filled if not.  I didn't dig into the source for this aliasExtractor, but somewhere I expect a _$plugin_ variable got left out.
DJ

imagra:
Just stepped on the same rake. And can confirm that baking the same tables for App (without --plugin parameter) works well. Also as Djamesfar found that shit happens somewhere before function aliasExtractor – the function gets empty list of associations.

djamesfar:
@imagra, since a fix for this looks like it keeps getting pushed down the road, I created a somewhat hairy workaround: we have an app running as a plugin with over 120 tables in the model.  I bake everything but the controller to the plugin, but the controller I bake without the --plugin option so the output goes to /src/Controller.  I wrote a script that modifies the 'uses' statement to include the plugin namespace in each controller file and then copy it to the /plugins/<plugin>/src/Controller directory.

markstory:
@djamesfar Is this still an issue. I tried to reproduce based on your steps above (which were quite detailed thank you) and couldn't. The controller that was generated included the find(list) calls as it should.

imagra:
Checked. Can be reproduced on clean install:
```
CREATE DATABASE IF NOT EXISTS `cake-test3` DEFAULT CHARACTER SET utf32 COLLATE utf32_unicode_ci;
USE `cake-test3`;
DROP TABLE IF EXISTS `roles`;
[...]

PR #83, opened 2015-05-05
Fix issue where plugin controllers were missing associations.

markstory:
Add missing plugin name when getting table object for a name.

Refs #57


lorenzo:
Good job!

PR #85, opened 2015-05-07
Allow for singular table names when baking fixtures

dakota:
Encountered a use case where I am baking fixtures for all my tables. I have a few tables where the table name is singular. This PR adds in the ability to fallback to a singular table if the plural version does not exist.

### DIFF
--- a/src/Shell/Task/FixtureTask.php
+++ b/src/Shell/Task/FixtureTask.php
@@ -16,6 +16,7 @@ namespace Bake\Shell\Task;
 
 use Cake\Console\Shell;
 use Cake\Core\Configure;
+use Cake\Database\Exception;
 use Cake\Database\Schema\Table;
 use Cake\Datasource\ConnectionManager;
 use Cake\ORM\TableRegistry;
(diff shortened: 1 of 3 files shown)


src/Shell/Task/FixtureTask.php
```diff
@@ -175,7 +176,13 @@ public function bake($model, $useTable = null)
             );
         }
         $schemaCollection = $connection->schemaCollection();
-        $data = $schemaCollection->describe($useTable);
+        try {
+            $data = $schemaCollection->describe($useTable);
+        } catch (Exception $e) {
```
lorenzo: You should catch the specific exception

adayth:
I had the same issues with recommended i18n table.

src/Shell/Task/FixtureTask.php
dakota: I am. `Schema/Collection::describe()` throws a `Cake\Database\Exception`, which is what I'm catching here.

lorenzo: Ah ok

lorenzo:
Is it possible to add a test for this?

dakota:
Test done :)

markstory:
Any reason you didn't use the `--table` option?

adayth:
Maybe he tried to generate fixtures with ```bake fixtures all``` and found issue #86

dakota:
@markstory, I was running `bake fixtures all` on order to refresh all fixture files at once.

markstory:
Ok, with `all` you can't really use the option, so trying one other table name makes sense.

Issue #34, opened 2015-02-01
CakePHP3.0.0 RC1: While baking the model Form Validation should applied as per the rules defined at Database  end.


ib-anurag-mishra:
Hi,

if a column has defined as  'Unique', then 'validateUnique' rule should automatically applied while bake.
Please make it dynamic. It should  not only applied to fixed column names  'username', 'email' etc.

Also when  more than one column declared as  unique , then respective validations must also be applied while baking.

For example  

if I have declared collectively 'Column 1' and 'Column 2' and 'Column 3' as Unique, so respective validation should be automatically applied while bake.

ib-anurag-mishra:
Hi,
Please refer:  https://github.com/cakephp/cakephp/issues/5802

Use RulesChecker::isUnique() which takes array of fields as argument.

anuid28:
What if multiple fields collectively marked as 'Unique', For Example : 'first_name'  and  'last_name' fields may be marked as unique key at database end

dereuromark:
Then there should be a unique rule across those together automatically.

othercorey:
The validateUnique configuration still needs work for this I think.

PR #91, opened 2015-05-12
Add unique validation rule to fields which have a unique constraint

josegonzalez:
Note that because of how the built-in `validateUnique` method works, we cannot attach this validation rule to multiple fields.

Refs cakephp/bake#34

> Making this PR here as I don't have a way to run tests atm.

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -16,6 +16,7 @@ namespace Bake\Shell\Task;
 
 use Cake\Console\Shell;
 use Cake\Core\Configure;
+use Cake\Database\Schema\Table as SchemaTable;
 use Cake\Datasource\ConnectionManager;
 use Cake\ORM\Table;
 use Cake\ORM\TableRegistry;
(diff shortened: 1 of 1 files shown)


src/Shell/Task/ModelTask.php
```diff
@@ -569,6 +570,17 @@ public function fieldValidation($fieldName, array $metaData, $primaryKey)
             ]
         ];
 
+        foreach ($schema->constraints() as $constraint) {
+            $constraint = $schema->constraint($constraint);
+            if (!in_array($fieldName, $constraint['columns']) || count($constraint['columns']) > 1) {
+                continue;
+            }
+
+            if ($constraint['type'] == 'unique') {
```
dereuromark: ===

markstory: We have constants for this as well I think.

src/Shell/Task/ModelTask.php
```diff
@@ -569,6 +571,17 @@ public function fieldValidation($fieldName, array $metaData, $primaryKey)
             ]
         ];
 
+        foreach ($schema->constraints() as $constraint) {
+            $constraint = $schema->constraint($constraint);
+            if (!in_array($fieldName, $constraint['columns']) || count($constraint['columns']) > 1) {
+                continue;
+            }
+
```
markstory: Why only unique indexex with 1 column?

josegonzalez: It looks like the isUnique rule doesn't work for multiple columns.

lorenzo: I does work with multiple columns, but extra columns need to be passed as `scope`

markstory:
Merging as this is an incremental improvement over what came before.

PR #92, opened 2015-05-17
Dispatch custom events on a per-template basis during render process

josegonzalez:
This class allows users to implement custom event listeners in their applications, allowing the following kind of usage:

```php
EventManager::instance()->attach(function (Event $event) {
   (new VersionListener($event))->__invoke();
}, 'Bake.beforeRender');
```

This would be useful for users attempting to modify bake output by injecting/modifying view variables.


josegonzalez:
You apparently cannot do the following:

```
EventManager::instance()->attach(function (Event $event) {
   (new VersionListener($event))();
}, 'Bake.beforeRender');
```

Boo. Anyone have any ideas to make this better?

josegonzalez:
Switched to:

```
EventManager::instance()->attach(function (Event $event) {
   (new VersionListener($event))->execute();
}, 'Bake.beforeRender');

lorenzo:
Wouldn't it be better to specifically trigger the events you are trying to mimic here?

markstory:
I agree that separate events triggered per template type would be a bit simpler. With separate events we wouldn't need custom event listeners, and instead the custom view class could be a bit more custom :smile:

josegonzalez:
@lorenzo not sure what you mean. What I'm trying to solve with this PR is not needing to "detect" what type of file is being baked, which can be a manual process.

lorenzo:
@josegonzalez I meant doing the type check in the BakeView class and triggering specific events for each type

markstory:
@josegonzalez I think what @lorenzo and I are getting at is that in addition to the generic `Bake.beforeRender` event, we could also trigger `Bake.beforeRender.testCase` as well.

josegonzalez:
Yeah that makes sense. I'll adjust the PR to do that.

markstory:
Ping.

simkimsia:
:+1: to EventListener for Bake. I am happy to write more about the documentation for this. Need some primer. Will be available on #FriendsOfCake or #cakephp as kimstacks

josegonzalez:
@markstory fixered. I replaced `DS` with a period so we don't have to worry about directory separators on different platforms.

markstory:
Code looks good, just needs a test.

josegonzalez:
I added a simple test.

lorenzo:
@josegonzalez Would you mind adding some simple docs to the book? I'm not sure how this should be used, so I cannot write it myself

PR #95, opened 2015-05-18
Wrapped baked flash messages with i18n function. Closes #88

### LINKED ISSUE #88: Some baked strings are not wrapped with __();
This hardens localisation.
[Example strings](https://github.com/cakephp/bake/search?utf8=%E2%9C%93&q=%24this-%3EFlash-%3E&type=Code)

dmromanov:
Closes #88


dereuromark:
Looks good to me.

markstory:
We have had and not had this feature a few times now. If comes back it won't be leaving again.

dereuromark:
At least they don't hurt. Who doesn't use it can just ignore them. No real overhead here.

markstory:
Agreed. If people really hate them, they can always go the custom template route as well.

Issue #98, opened 2015-05-24
Records with datetime break Fixture


sean-nicholas:
When using records from database in FixtureTask with a datetime field it will create following code:

```php
 public $records = [
        [
            ...
            'last_login' => Cake\I18n\Time::__set_state(array(
   'date' => '2015-02-05 12:58:15',
   'timezone_type' => 3,
   'timezone' => 'Europe/Berlin',
)),
...
```

Which results in 
> Parse error: syntax error, unexpected '(', expecting ']'

Maybe the records should be set in 
```php
public function init() {}
```
But the second problem is the namespace:

> Fatal error: Class 'App\Test\Fixture\Cake\I18n\Time' not found in tests\Fixture\UsersFixture.php on line 69

There is missing a blackslash before Cake\I18n\Time

markstory:
Pull request open now.

Issue #97, opened 2015-05-24
Record properties don't match fields in FixtureTask


sean-nicholas:
FixtureTask will create entities when using records from database: https://github.com/cakephp/bake/blob/master/src/Shell/Task/FixtureTask.php#L423-426

So $_hidden and $_virtual will be parsed and you get a record whose properties don't match the fields: 
> SQLSTATE[21S01]: Insert value list does not match column list: 1136 Column count doesn't match value count at row 1

Maybe using hydrate(false) will help?

lorenzo:
@markstory I created https://github.com/cakephp/cakephp/issues/6627

markstory:
I'm too slow :smile:

markstory:
Reopening as I think the code in bake needs to change too.

markstory:
Pull request open now.

PR #99, opened 2015-05-24
Fix datetime field generation, and entity usage.

markstory:
Fixtures should be generated with string datetime values, not object instances. Using strings matches the format people would normally author fixtures in. Using hydrate(false) means fixture generation will not be impacted by any virtual, hidden, or accessors on the fixtures.

Refs #97
Refs #98

### DIFF
--- a/src/Shell/Task/FixtureTask.php
+++ b/src/Shell/Task/FixtureTask.php
@@ -22,6 +22,7 @@ use Cake\Datasource\ConnectionManager;
 use Cake\ORM\TableRegistry;
 use Cake\Utility\Inflector;
 use Cake\Utility\Text;
+use DateTime;
 
 /**
  * Task class for creating and updating fixtures files.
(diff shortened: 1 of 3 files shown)


lorenzo:
Good job!

src/Shell/Task/FixtureTask.php
```diff
@@ -386,6 +387,9 @@ protected function _makeRecordString($records)
         foreach ($records as $record) {
             $values = [];
             foreach ($record as $field => $value) {
+                if ($value instanceof DateTime) {
```
mattmemmesheimer: This is still an issue when using immutable time objects (specified in bootstrap.php).  `$value` will be an instance of `Cake\I18n\FrozenTime`.

 Should this `if` statement check if `$value` is an instance of `DateTimeInterface` instead of `DateTime`?

This resolved the issue for me.

lorenzo: A pull request with the proposed change would be nice.

mattmemmesheimer: https://github.com/cakephp/bake/pull/257

PR #101, opened 2015-05-28
Add Bake task for shell helpers

markstory:
With Shell Helpers mostly complete, I thought it would be good to have bake tasks for them as well.

### DIFF
--- a/src/Template/Bake/Shell/helper.ctp
+++ b/src/Template/Bake/Shell/helper.ctp
@@ -1,0 +1,36 @@
+<%
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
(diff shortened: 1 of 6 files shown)


src/Template/Bake/Shell/helper.ctp
```diff
@@ -0,0 +1,36 @@
...
+{
+
+    /**
+     * Output method.
+     *
+     * Generate the output for this shell helper.
+     *
+     * @return void
+     */
+    public function output($args) {
```
antograssiot: PSR2 maybe ? :stuck_out_tongue:

markstory: Right...

lorenzo:
:clap:

PR #103, opened 2015-05-30
Add possibility to bake modeless form

antograssiot:
and tests for it

### DIFF
--- a/src/Shell/Task/TestTask.php
+++ b/src/Shell/Task/TestTask.php
@@ -56,6 +56,7 @@ class TestTask extends BakeTask
         'Shell' => 'Shell',
         'Shell_helper' => 'Shell\Helper',
         'Cell' => 'View\Cell',
+        'Form' => 'Form'
     ];
 
     /**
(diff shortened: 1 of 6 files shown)


lorenzo:
This could probably just go to master

josegonzalez:
yay

src/Shell/Task/TestTask.php
```diff
@@ -601,7 +603,9 @@ public function getOptionParser()
                 'Component', 'component',
                 'Behavior', 'behavior',
                 'Shell', 'shell',
+                'shell_helper',
```
markstory: Thanks :+1:

markstory:
Looks good :+1:

antograssiot:
I could target master, then I'll have to update 1.1 for shell helper but I'm ok with it if others also think it's worth it

markstory:
@antograssiot It means we can ship form baking faster. I can fix the mistake on `1.1` if its going to be a hassle.

antograssiot:
@markstory no no pb, I'll get this done

PR #104, opened 2015-05-30
New design generators

mtkocak:
I think I have to update tests too...

### DIFF
--- a/src/Template/Bake/Element/form.ctp
+++ b/src/Template/Bake/Element/form.ctp
@@ -19,35 +19,35 @@ $fields = collection($fields)
         return $schema->columnType($field) !== 'binary';
     });
 %>
-<div class="actions columns large-2 medium-3">
-    <h3><?= __('Actions') ?></h3>
-    <ul class="side-nav">
-<% if (strpos($action, 'add') === false): %>
-        <li><?= $this->Form->postLink(
-                __('Delete'),
-                ['action' => 'delete', $<%= $singularVar %>-><%= $primaryKey[0] %>],
-                ['confirm' => __('Are you sure you want to delete # {0}?', $<%= $singularVar %>-><%= $primaryKey[0] %>)]
-            )
-        ?></li>
-<% endif; %>
-        <li><?= $this->Html->link(__('List <%= $pluralHumanName %>'), ['action' => 'index']) ?></li>
-<%
-        $done = [];
-        foreach ($associations as $type => $data) {
-            foreach ($data as $alias => $details) {
-                if ($details['controller'] != $this->name && !in_array($details['controller'], $done)) {
-%>
-        <li><?= $this->Html->link(__('List <%= $this->_pluralHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'index']) %></li>
-        <li><?= $this->Html->link(__('New <%= $this->_singularHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'add']) %></li>
-<%
-                    $done[] = $details['controller'];
-                }
-            }
-        }
-%>
-    </ul>
-</div>
-<div class="<%= $pluralVar %> form large-10 medium-9 columns">
+    <nav class="large-3 medium-4 columns" id="actions-sidebar">
+        <ul class="side-nav">
+           <li class="heading"><?= __('Actions') ?></li>
+           <% if (strpos($action, 'add') === false): %>
+                   <li><?= $this->Form->postLink(
+                           __('Delete'),
+                           ['action' => 'delete', $<%= $singularVar %>-><%= $primaryKey[0] %>],
+                           ['confirm' => __('Are you sure you want to delete # {0}?', $<%= $singularVar %>-><%= $primaryKey[0] %>)]
+                       )
+                   ?></li>
+           <% endif; %>
+                   <li><?= $this->Html->link(__('List <%= $pluralHumanName %>'), ['action' => 'index']) ?></li>
+           <%
+                   $done = [];
+                   foreach ($associations as $type => $data) {
+                       foreach ($data as $alias => $details) {
+                           if ($details['controller'] != $this->name && !in_array($details['controller'], $done)) {
+           %>
+                   <li><?= $this->Html->link(__('List <%= $this->_pluralHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'index']) %></li>
+                   <li><?= $this->Html->link(__('New <%= $this->_singularHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'add']) %></li>
+           <%
+                               $done[] = $details['controller'];
+                           }
+                       }
+                   }
+           %>
+      </ul>
+   </nav>
+<div class="<%= $pluralVar %> form large-9 medium-8 columns" id="content">
     <?= $this->Form->create($<%= $singularVar %>) ?>
     <fieldset>
         <legend><?= __('<%= Inflector::humanize($action) %> <%= $singularHumanName %>') ?></legend>
--- a/src/Template/Bake/Element/form.ctp
+++ b/src/Template/Bake/Element/form.ctp
@@ -19,34 +19,34 @@ $fields = collection($fields)
         return $schema->columnType($field) !== 'binary';
     });
 %>
-    <nav class="large-3 medium-4 columns" id="actions-sidebar">
-        <ul class="side-nav">
-           <li class="heading"><?= __('Actions') ?></li>
-           <% if (strpos($action, 'add') === false): %>
-                   <li><?= $this->Form->postLink(
-                           __('Delete'),
-                           ['action' => 'delete', $<%= $singularVar %>-><%= $primaryKey[0] %>],
-                           ['confirm' => __('Are you sure you want to delete # {0}?', $<%= $singularVar %>-><%= $primaryKey[0] %>)]
-                       )
-                   ?></li>
-           <% endif; %>
-                   <li><?= $this->Html->link(__('List <%= $pluralHumanName %>'), ['action' => 'index']) ?></li>
-           <%
-                   $done = [];
-                   foreach ($associations as $type => $data) {
-                       foreach ($data as $alias => $details) {
-                           if ($details['controller'] != $this->name && !in_array($details['controller'], $done)) {
-           %>
-                   <li><?= $this->Html->link(__('List <%= $this->_pluralHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'index']) %></li>
-                   <li><?= $this->Html->link(__('New <%= $this->_singularHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'add']) %></li>
-           <%
-                               $done[] = $details['controller'];
-                           }
-                       }
-                   }
-           %>
-      </ul>
-   </nav>
+<nav class="large-3 medium-4 columns" id="actions-sidebar">
+    <ul class="side-nav">
+        <li class="heading"><?= __('Actions') ?></li>
+<% if (strpos($action, 'add') === false): %>
+        <li><?= $this->Form->postLink(
+                       __('Delete'),
+                       ['action' => 'delete', $<%= $singularVar %>-><%= $primaryKey[0] %>],
+                       ['confirm' => __('Are you sure you want to delete # {0}?', $<%= $singularVar %>-><%= $primaryKey[0] %>)]
+                   )
+        ?></li>
+<% endif; %>
+        <li><?= $this->Html->link(__('List <%= $pluralHumanName %>'), ['action' => 'index']) ?></li>
+<%
+        $done = [];
+        foreach ($associations as $type => $data) {
+            foreach ($data as $alias => $details) {
+                if ($details['controller'] != $this->name && !in_array($details['controller'], $done)) {
+%>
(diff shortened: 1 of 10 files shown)


markstory:
Looks like the tests do need to be updated. I would wait until the feedback rounds are done before updating the test to save some work.

AD7six:
You can auto-update the test expectations by temporarily modifying [the string compare trait](https://github.com/cakephp/cakephp/blob/master/src/TestSuite/StringCompareTrait.php#L46) like so:

```
    public function assertSameAsFile($path, $result)
    {
        $path = $this->_compareBasePath . $path;
        file_put_contents($path, $result); // <-
        $expected = file_get_contents($path);
        $this->assertTextEquals($expected, $result);
    }
```

Then running the affected/failing tests, committing and pushing the changes to the tests folder.

src/Template/Bake/Template/view.ctp
```diff
@@ -72,72 +72,75 @@ $pk = "\$$singularVar->{$primaryKey[0]}";
...
 <% foreach ($groupedFields['text'] as $field) : %>
-    <div class="row texts">
-        <div class="columns large-9">
-            <h6 class="subheader"><?= __('<%= Inflector::humanize($field) %>') ?></h6>
-            <?= $this->Text->autoParagraph(h($<%= $singularVar %>-><%= $field %>)) ?>
-        </div>
-    </div>
+<div class="row">
+            <h4><?= __('<%= Inflector::humanize($field) %>') ?></h4>
+            <?= $this->Text->autoParagraph(h($<%= $singularVar %>-><%= $field %>)); ?>
```
markstory: Indentation seems off.

mtkocak:
is there any cakephp 3 bundle for textmate?

markstory:
@mtkocak I'm not sure, I've not used textmate in years. However, you may be able to find a generic PHP bundle that supports PSR-2 formatting.

mtkocak:
I think I did it. I updated all test manually and tested using phpunit. by the way @markstory  what ide do you use?

mtkocak:
coverage/coveralls ? @AD7six

AD7six:
> Coverage decreased (-0.2%) 

You can ignore that @mtkocak - all the test files are missing a newline at the end of the file. Apart from that :+1: great work =).

mtkocak:
oh I see. I got coverage problem because I did not fetch 1.1 :P I am going to add new lines too

mtkocak:
I think we are done here. https://github.com/cakephp/bake/blob/master/tests/comparisons/Template/testBakeEdit.ctp does not have a new line at the end. do I have to check any other file?

markstory:
@mtkocak Just vim with [some config/plugins](https://github.com/markstory/vim-files).

mtkocak:
@markstory thanks for the advice, I will try that.

lorenzo:
Great job @mtkocak :clap:

tests/comparisons/Template/testBakeTemplate.ctp
```diff
@@ -8,33 +8,37 @@
         <li><?= $this->Html->link(__('List Articles'), ['controller' => 'Articles', 'action' => 'index']) ?> </li>
         <li><?= $this->Html->link(__('New Article'), ['controller' => 'Articles', 'action' => 'add']) ?> </li>
     </ul>
-</div>
-<div class="templateTaskComments view large-10 medium-9 columns">
-    <h2><?= h($templateTaskComment->id) ?></h2>
+</nav>
+<div class="templateTaskComments view large-9 medium-8 columns" id="content">
```
ADmad: I see that class `templateTaskComments` was already used earlier but it should be switched to dashed version for consistency.

jadb: :+1:

src/Template/Bake/Template/view.ctp
```diff
@@ -73,95 +73,91 @@ $pk = "\$$singularVar->{$primaryKey[0]}";
     }
 %>
     </ul>
-</div>
-<div class="<%= $pluralVar %> view large-10 medium-9 columns">
-    <h2><?= h($<%= $singularVar %>-><%= $displayField %>) ?></h2>
-    <div class="row">
+</nav>
+<div class="<%= $pluralVar %> view large-9 medium-8 columns" id="content">
```
dereuromark: Personally, I am not a fan of too many non-js-functional IDs in the HTML code layout. Especially such generic ones, as they need to be unique across the HTML body and there could easily be another content injected somewhere. Is there a reason those IDs have been added? Or could those just be a class instead?

src/Template/Bake/Element/form.ctp
```diff
@@ -19,9 +19,9 @@ $fields = collection($fields)
         return $schema->columnType($field) !== 'binary';
     });
 %>
-<div class="actions columns large-2 medium-3">
-    <h3><?= __('Actions') ?></h3>
+<nav class="large-3 medium-4 columns" id="actions-sidebar">
```
ionas: Are there ways to do away with the inline classes?

src/Template/Bake/Element/form.ctp
```diff
@@ -46,8 +46,8 @@ $fields = collection($fields)
         }
 %>
     </ul>
-</div>
-<div class="<%= $pluralVar %> form large-10 medium-9 columns">
+</nav>
+<div class="<%= $pluralVar %> form large-9 medium-8 columns" id="content">
```
ionas: Same as above.

src/Template/Bake/Template/index.ctp
```diff
@@ -20,9 +20,9 @@ $fields = collection($fields)
     })
     ->take(7);
 %>
-<div class="actions columns large-2 medium-3">
-    <h3><?= __('Actions') ?></h3>
+<nav class="large-3 medium-4 columns" id="actions-sidebar">
```
ionas: Same as above.

src/Template/Bake/Template/index.ctp
```diff
@@ -20,9 +20,9 @@ $fields = collection($fields)
...
     <ul class="side-nav">
+        <li class="heading"><?= __('Actions') ?></li>
```
ionas: Same as above.

src/Template/Bake/Template/index.ctp
```diff
@@ -39,20 +39,21 @@ $fields = collection($fields)
     endforeach;
 %>
     </ul>
-</div>
-<div class="<%= $pluralVar %> index large-10 medium-9 columns">
+</nav>
+<div class="<%= $pluralVar %> index large-9 medium-8 columns" id="content">
```
ionas: Same as above.

markstory: What do you mean by inline classes? All classes are 'inline'.

src/Template/Bake/Element/form.ctp
ADmad: I don't see how you can do away with classes for column widths.

jhli: By using css (or sass or scss or less).

ADmad: @jhli What's wrong with existing css classes used?

jhli: Not depending on whatever-framework-was used style.
being able to replace the presentation.
http://alistapart.com/article/separationdilemma
http://alistapart.com/column/no-good-can-come-of-bad-code

In this case the way without a preprocessor would be to copy the class attribute to css as a comment also refering to the franework utilized AND below create the very few selectors and include their properties.

I am very happy with the inprovement though as now it looks good!

src/Template/Bake/Template/index.ctp
mtkocak: the semantic reason for an id and class usage in an app is to have a line
between repeating objects and unique ones. Actually I would love to delete
all the css and write my own but that would not be backwards compatible.
due to prevent clashes and having discrete semantic elements I had to use
id, so each page had its own unique discrete sections too.

11 Haziran 2015 Perşembe tarihinde, Jonas <notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> yazdı:

mtkocak: another thing I did not use any javascript because foundation's js files
were not included in the package. I avoided anything that would increase
package size nor preprocessors or etc

11 Haziran 2015 Perşembe tarihinde, Midori Koçak <mtkocak@gmail.com> yazdı:

ADmad: @mtkocak The benefit of basing the css off a standard css lib (zurb foundation in this case) is that people can easily add new elements as per their app needs.

mtkocak: I am really missing point here. I think foundation zurb is a good choice.
yes having these classes is a common problem for separating presentation
and code but a person could easily disable the css and can use ids instead
at any time. so I added these ids on purpose. also I don't know ehat would
be the alternative or advantage of not using a css library. for me the
foundation was a huge improvement at using cake because I had routunes
from1.2 times changing standard ugly css to bootstrap and foundation. I
even wrote some js libraries to avoid these routines at that times.  the
problems talked here is another issue and common for css libraries. but the
frontend development is going somewhere different using schema.org or
structured data, machine readable information without any presentation. but
as a php fw for a base this is not our problem to solve. actually I would
love to create a new generator using only semantic elements without any
decoration and that would be a future improvement. I am open to all ideas
though.

11 Haziran 2015 Perşembe tarihinde, ADmad <notifications@github.com> yazdı:

ionas: There is no semantic reason usually for classes (stuff like http://en.wikipedia.org/wiki/Microformat aside).

As @mtkocak already pointed out: Foundation is not bundled and classes (and mixins) may miss so @ADmad there is no easy way to add new elements as required by the app.

If I get to it I will purpose a fix that simply uses css selectors for the same set of properties instead of classes.

src/Template/Bake/Template/index.ctp
```diff
@@ -39,20 +39,21 @@ $fields = collection($fields)
...
+    <h3><?= __('<%= $pluralHumanName %>') ?></h3>
     <table cellpadding="0" cellspacing="0">
```
ionas: Cellspacing and cellpadding can be removed. See http://stackoverflow.com/a/3209434

src/Template/Bake/Template/index.ctp
ionas: @mtkocak I support to create a pure semantic generator. As I pointed out in another ticket in the long run I'd either want to see a bake result that does good practice without exception or one that does as little as possible, e.g. maybe just do html5 + normalize.css and be done with it.

src/Template/Bake/Template/view.ctp
ionas: One could utilize <main>

tests/comparisons/Template/testBakeEdit.ctp
```diff
@@ -1,6 +1,6 @@
-<div class="actions columns large-2 medium-3">
-    <h3><?= __('Actions') ?></h3>
+<nav class="large-3 medium-4 columns" id="actions-sidebar">
     <ul class="side-nav">
+        <li class="heading"><?= __('Actions') ?></li>
```
ionas: I am not sure having a list of links with the first one being a description of the rest of the list is such a good idea.

src/Template/Bake/Template/index.ctp
markstory: I completely disagree that class names need to be semantic. From working on a few large applications, 'semantic' class names create css sprawl. While grid classes look ugly they offer far simpler mechanics for code reuse and I've found them much easier to maintain in the context of larger applications.

ionas: And on those 9 large applications I can speak of whenever there is a mess of tags and presentational classes I can tell that those are hard to maintain, hard to extend, hard to do SEO and speed optimization on, slow on clients and demanding a lot of CPU and RAM as well as bandwith usually even having a ton of unnecessary http requests (because with the laziness usually comes including tons of plugins), being error prone and hard to integrate content management in an easy way and last but not least being vendor locked into whatever frontend framework used (in this case into Foundation, for us into Bootstrap). In my experience working on ecommerce and point of sale systems since 1995:

It is true that table-layouts / grid classes are simple to start with and use but hard to maintain and bad practice entirely. They are mostly, an excuse for not understanding DOM and CSS selectors. See https://www.ostraining.com/blog/coding/bootstrap-right-way/ for instance.

markstory: How do classnames impact SEO? I've not worked much on e-commerce applications so I don't know how CSS class name conventions and SEO could be linked.

ionas: Doing stuff semantically is like a 90% automation to SEO optimization of HTML. In rare cases you want something to "look" like ```<h2>``` but actually be just a ```<p>```. Having a lot of class names and non-semantic wrappers (```<div>```s) make it hard to create html for layouts, and view template files as well as content that is responsive and semantic/search-engine optimized.

In my defense / pro your argument: The way I did my last huge PRs on this matter. It was one big and hard to maintain file. This doesn't have to be the case and I was not focusing on maintainability (but should have).

AD7six: > In rare cases you want something to "look" like <h2> but actually be just a <p>

Giving the user one impression of the markup, yet giving a html parser another is a shady practice - also not particularly relevant here.

ionas: The example with something looking like h2 but not being h2 was an example for a rare exception to my argument that creating semantic markup paves the way to do SEO easily, aside that it is not relevant and I don't like that practice either.

mtkocak:
I am totally lost.

ionas:
Don't worry. It got merged, it is tons better now. Thank you :+1: for making me love cake bake output visually again! :)

PR #105, opened 2015-05-30
Add possibility to bake modeless form

antograssiot:
and tests

(we should create milestone for 1.0.9)

PR #106, opened 2015-05-30
follows #101 and #105

antograssiot:


tests/TestCase/Shell/BakeShellTest.php
```diff
@@ -106,7 +106,7 @@ public function testMain()
             ->method('out')
             ->with($this->stringContains('The following commands'));
 
-        $this->Shell->expects($this->exactly(17))
+        $this->Shell->expects($this->exactly(18))
```
AD7six: I think it's about time this test was changed to be actually useful "this method calls out 18 times" is of little to no value.

markstory: I agree this is getting to be diminishing returns at this point.

AD7six: Addessed (hopefully) in https://github.com/cakephp/bake/pull/107

antograssiot:
closing, I'll reopen an other one

PR #107, opened 2015-05-30
Get rid of expects(big number)

AD7six:
And instead, assert the text looks like the output.

Depends upon https://github.com/cakephp/cakephp/pull/6694

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.4",
-        "cakephp/cakephp": "~3.0"
+        "cakephp/cakephp": "dev-master"
     },
     "require-dev": {
         "cakephp/cakephp-codesniffer": "dev-master",
--- a/tests/TestCase/Shell/BakeShellTest.php
+++ b/tests/TestCase/Shell/BakeShellTest.php
@@ -15,9 +15,11 @@
 namespace Bake\Test\TestCase\Shell;
 
 use Bake\Test\TestCase\TestCase;
+use Cake\Console\ConsoleIo;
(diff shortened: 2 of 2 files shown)


markstory:
Much nicer :clap:

AD7six:
["Oh"](https://github.com/cakephp/bake/blob/master/.travis.yml#L30)

composer.json
```diff
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.4",
-        "cakephp/cakephp": "~3.0"
+        "cakephp/cakephp": "dev-master"
```
josegonzalez: Why this change? Just curious.

markstory: I think this was a mistake. I'll revert it.

markstory: Hmm without relying on dev-master the tests fail because the stub console output is not yet in a packaged release.

josegonzalez: Well sounds like this depends on 3.1 then?

josegonzalez: Or at least can't be 3.x.

lorenzo: I think setting this to dev-master is a mistake, can we revert it?

AD7six: It is a temporary change so that the tests pass on travis - there is no release with the stub console output yet.

markstory: I'll probably do 3.0.7 in a week or so. We'll have to not release bake until after that :smile:

jeremyspencer:
screenshot_2015-10-07-04-21-18

jeremyspencer:
https://plus.google.com/109951829811741231558

PR #108, opened 2015-05-31
Feature/bake everything

steinkel:
Added a new "bake all everything" option.

The bake unit tests were in a funny state, so I'll push the additional unit test fixes once green is back in town...

### DIFF
--- a/src/Shell/BakeShell.php
+++ b/src/Shell/BakeShell.php
@@ -284,6 +284,8 @@ class BakeShell extends Shell
             ' are using command line arguments.'
         )->addSubcommand('all', [
             'help' => 'Bake a complete MVC skeleton.',
+        ])->addSubcommand('all everything', [
+            'help' => 'Bake a complete MVC skeleton, using all the available tables.',
         ])->addOption('connection', [
             'help' => 'Database connection to use in conjunction with `bake all`.',
             'short' => 'c',
--- a/src/Shell/BakeShell.php
+++ b/src/Shell/BakeShell.php
@@ -222,10 +222,9 @@ class BakeShell extends Shell
             $this->connection = $this->params['connection'];
         }
 
-        if (empty($name)) {
+        if (empty($name) && !$this->params['everything']) {
             $this->Model->connection = $this->connection;
             $this->out('Possible model names based on your database:');
-            $this->out('- everything');
             foreach ($this->Model->listAll() as $table) {
                 $this->out('- ' . $table);
             }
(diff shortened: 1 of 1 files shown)


src/Shell/BakeShell.php
```diff
@@ -225,6 +225,7 @@ public function all($name = null)
         if (empty($name)) {
             $this->Model->connection = $this->connection;
             $this->out('Possible model names based on your database:');
+            $this->out('- everything');
```
ADmad: There should be some indication that this is a special name.

steinkel: Should I add it as a class constant?

steinkel: Could be an option instead too... like "--everything"

ADmad: An option would be better.

markstory:
The tests [seem to be failing](https://travis-ci.org/cakephp/bake/jobs/64825354) because of your change.

src/Shell/BakeShell.php
```diff
@@ -222,7 +222,7 @@ public function all($name = null)
             $this->connection = $this->params['connection'];
         }
 
-        if (empty($name)) {
+        if (empty($name) && !$this->params['everything']) {
```
markstory: If you use `param()` you can avoid the notice errors.

src/Shell/BakeShell.php
```diff
@@ -232,15 +232,28 @@ public function all($name = null)
...
-            $this->{$task}->connection = $this->connection;
+        $allTables = collection([$name]);
+        $filteredTables = $allTables;
+
+        if ($this->params['everything']) {
+            $this->Model->connection = $this->connection;
+            $allTables = collection($this->Model->listAll());
+            $filteredTables = $allTables->reject(function ($tableName) {
+                return $tableName === 'phinxlog' || $tableName === 'users_phinxlog';
+            });
```
markstory: Do we want to toss out `i18n` and `cake_sessions` two?

src/Shell/BakeShell.php
```diff
@@ -270,6 +283,10 @@ public function getOptionParser()
             ' are using command line arguments.'
         )->addSubcommand('all', [
             'help' => 'Bake a complete MVC skeleton.',
+        ])->addOption('everything', [
+            'help' => 'Bake a complete MVC skeleton, using all the available tables. ' .
+            'Usage: "bake all --everything true"',
```
markstory: You don't need to set true. You can set `'boolean' => true`.

lorenzo:
Great job @steinkel !

PR #109, opened 2015-06-01
Add type hints for associated tables.

### LINKED ISSUE #59: Automatically add @property annotations to Tables
It would be nice if baked Table classes would have `@property` annotations for the associations they have, like this:
```php
/**
 * Entries Model
 *
 * @property \Cake\ORM\Association Users
 * @property \Cake\ORM\Association Jobs
 */
```
(The code above was in fact added by a PhpStorm inspection and is really usefull)

AFAIK this is already done for Controller so it should not be so hard to implement.

ndm2:
Resolves #59

### DIFF
--- a/src/Template/Bake/Model/table.ctp
+++ b/src/Template/Bake/Model/table.ctp
@@ -32,6 +32,14 @@ echo implode("\n", $uses);
 
 /**
  * <%= $name %> Model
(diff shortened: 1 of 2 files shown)


markstory:
AppVeyor fails seem unrelated, I'll fix those separately.

Issue #110, opened 2015-06-02
Bake `--force` option is broken


lorenzo:
It does not even show in the help text for bake

antograssiot:
@lorenzo In wich way is it broken ? i used it successfully with almost every command.
I have a PR ready to add the option to the help text but I can look into it with more information

lorenzo:
It is broken as it does not work with `cake bake all users -f` for example

markstory:
The all command is missing the `force` option. Other tasks have it though.

markstory:
I think my pull request solves this.

PR #112, opened 2015-06-02
Add force option to `bake all`.

markstory:
Refs #110


antograssiot:
looks good :+1: , I had the same ready :stuck_out_tongue:

Issue #114, opened 2015-06-05
Baking views for models omits self associations


markstory:
1. Create a model with a self-association. For example, links between two users, or a tree structure.
2. Bake views for this model with a self-association. The parent/child associations will be omitted from the views.

It looks like this issue was introduced in #4440 when navigation links were cleaned up. However, I think the related entity blocks should still be included in the view page. Perhaps we need another flag in the association data that indicates whether or not the association should be in the navigation as omitting self associations entirely is not great.

PR #115, opened 2015-06-10
Fix self associations not showing up in the related sections.

markstory:
Self associations like child categories were not showing up in the related sections. Previously these associations was filtered out entirely to clean up the navigation. This had the unfortunate side effect of eliding them from the view page as well.

This introduces a new template variable that indicates whether or not an association should be in the nav. Self-associations are marked false. This also fixes an issue where multi word self-associations would have the wrong heading.

Refs #114


jadb:
Nice!

PR #117, opened 2015-06-11
Try and get AppVeyor passing

markstory:
Update tests and fixtures. I hope that this solves the failing tests in AppVeyor as local tests passed against SQLServer 2008.


lorenzo:
You rock :100:

PR #118, opened 2015-06-11
Don't use common id selectors.

markstory:
We can get the same CSS to apply with classnames. This will match up with a pull request in cakephp/app


ADmad:
Tests need updating :)

markstory:
@ADmad I remembered seconds after opening the pull request :cry:

Issue #120, opened 2015-06-12
Table::initialize() is not calling parent in baked table classes


burzum:
The title says it all.

Check the discussion here https://github.com/FriendsOfCake/bootstrap-ui/issues/63

AD7six:
What the title does not say is in what way that's a problem - the parent class method [is empty](https://github.com/cakephp/cakephp/blob/master/src/ORM/Table.php#L304-L324)

burzum:
@AD7six today it's empty, what if it changes? What if I extend my custom Table object like AppTable?

But feel free to close it if you disagree.

dereuromark:
I don't think it would hurt to add the parent call - for consistency at least.

AD7six:
> what if it changes? 

I don't think that's likely.

I also don't think a parent call would hurt but there was no clarification as to whether it's a problem or other.

PR #124, opened 2015-06-19
Add missing parent call for Table::initialize().

dereuromark:
Closes https://github.com/cakephp/bake/issues/120

### DIFF
--- a/src/Template/Bake/Model/table.ctp
+++ b/src/Template/Bake/Model/table.ctp
@@ -52,6 +52,8 @@ class <%= $name %>Table extends Table
      */
     public function initialize(array $config)
     {
+        parent::initialize($config);
+
 <% if (!empty($table)): %>
         $this->table('<%= $table %>');
 <% endif %>
--- a/src/Template/Bake/Model/table.ctp
+++ b/src/Template/Bake/Model/table.ctp
@@ -89,15 +89,9 @@ class <%= $name %>Table extends Table
      */
     public function validationDefault(Validator $validator)
     {
-        $validator
-<% $validationMethods = []; %>
 <%
-$firstField = true;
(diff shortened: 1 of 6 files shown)


dereuromark:
Is there a way to remove that whitespace noise in the bake templates when generating?

AD7six:
It fails because bakes output contains extra whitespace - the correct fix is to fix the output, looks to me like [this](https://github.com/cakephp/bake/blob/master-table/src/Template/Bake/Model/table.ctp#L147) is responsible:

    <%= "            " . implode("\n            ", $validationMethods) %>

There shouldn't be any "echo whitespace" logic in these templates, only echo statements at the right indentation level.

dereuromark:
Or running a trim right on the generated code on a per row level.
Either way, the generated code shouldn't contain whitespace noise.

AD7six:
> Or running a trim right on the generated code on a per row level.

I don't want that kind of logic - the template is right, or it's not. not mostly-right with magic applied. There'll be cases where you want/expect trailing whitespace and that'd prevent it.

markstory:
Seems fine one the tests pass.

dereuromark:
Feel free to fix the unrelated whitespace issue either as commits on top or separate PR.
Then this one will be fine to merge.

AD7six:
@dereuromark it's changes in this pr that cause the test fails (it previously matched output, unrelated changes [to the test file in this PR](https://github.com/cakephp/bake/pull/124/files#diff-db1cb4ba7b04ad8856a4812ab4fe2a8aL35) make it fail).

As I've [mentioned before](https://github.com/cakephp/bake/pull/104#issuecomment-107087170) its easy to get the test output to match expectations - I can fix the now-apparent bad template later/tomorrow.

PR #133, opened 2015-07-09
Update gitattributes file

bcrowe:
We don't need any of this included during Composer installs, etc.


.gitattributes
```diff
@@ -2,3 +2,41 @@
...
+# Remove files for archives generated using `git archive`
+appveyor.yml export-ignore
+CONTRIBUTING.md export-ignore
+.editorconfig export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+Makefile export-ignore
+phpunit.xml.dist export-ignore
+.travis.yml export-ignore
+tests/test_app export-ignore
```
ADmad: Can just ignore `tests/`. In core we want the fixtures to be available to users for development hence we ignore these 2 dirs separately.

.gitattributes
```diff
@@ -2,3 +2,41 @@
 # Set default behaviour, in case users don't have core.autocrlf set.
 * text=auto
 * text eol=lf
+
+# Explicitly declare text files we want to always be normalized and converted
+# to native line endings on checkout.
```
ADmad: I would trim down this list to only file types this repo actually contains.

.gitattributes
markstory: I don't think we need fixtures is dist installs either.

ADmad: Non dist installs which pull in cake repo history are pretty slow and often timeout. That was a major annoyance during plugin development and testing. Keeping the fixtures in core's dist install allows using the dist install for development too and speed up testing.

markstory: Yes, I meant bake's fixtures shouldn't be included in dist installs of bake.

ADmad: Sorry, i misunderstood your comment. Reading it again i realize you were just agreeing with my comment :)

.gitattributes
bcrowe: Meh... it doesn't hurt leaving them around. It'll just be another thing to remember if any of these types of files start creeping into this repo at any point.

bcrowe:
Updated to export-ignore entire the tests directory.

PR #134, opened 2015-07-14
Only set primary key fields as non-accessible.

ADmad:
This change avoids having to update the accessible list when new fields are added to table.

This avoids some hair pulling for new users who don't realize the entity also needs to be updated if they change table fields. I have seen newbies often struggle to figure out why their new fields are not getting saved. e.g. http://stackoverflow.com/questions/31385268/cakephp-3-new-fields-wont-save-correctly

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -708,7 +708,7 @@ class ModelTask extends BakeTask
             'namespace' => $namespace,
             'plugin' => $this->plugin,
             'pluginPath' => $pluginPath,
-            'fields' => [],
+            'primaryKey' => [],
         ];
 
         $this->BakeTemplate->set($data);
--- a/src/Template/Bake/Model/entity.ctp
+++ b/src/Template/Bake/Model/entity.ctp
@@ -23,7 +23,7 @@ use Cake\ORM\Entity;
  */
 class <%= $name %> extends Entity
 {
-<% if (!empty($fields)): %>
+<% if (!empty($primaryKey)): %>
 
     /**
      * Fields that can be mass assigned using newEntity() or patchEntity().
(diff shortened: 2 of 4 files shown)


lorenzo:
Make me sad making the security feature less obvious

ADmad:
@lorenzo Life is full of compromises :smile:

markstory:
:cry: At least people who need more secure apps have better tools at their disposal.

GuidoHendriks:
:-1: I'd rather have security by default than ease of use for "newbies".

Spriz:
+1 to @GuidoHendriks :smile:

ADmad:
@GuidoHendriks @Spriz I am not particularly thrilled by this change but keeping things simpler for newbies is also important as initial impressions matter.

You can use custom bake template to get earlier behavior.

bcrowe:
Maybe we should add a note to the doc block?

ADmad:
@bcrowe Sounds good, can you please make PR with appropriate text?

ndm2:
The `fields` and `no-fields` options are now kinda superflous as they are not being used anymore. I'm wondering whether they should be removed, or maybe incorporated again, so that they override the default behavior in case they are being used? ie, if `no-fields` is used, don't created `$_accessible`, if `fields`is used, fill `$_accessible` with only those fields specified.

markstory:
I would expect `fields` would result in specific fields being added into the accessible list.

ndm2:
You mean added to what is currently being generated (the `*` and the primary key fields)?

markstory:
Well if there is an explicit lust of fields from the user only those fields should be marked accessible.

ndm2:
OK, that's what I was suggesting too, I misunderstood you then. I'll try to implement this before I continue with the entity property hints.

jhli:
Maybe instead mass assigning to a field that is not accessible could throw a warning on debug true?

PR #137, opened 2015-07-22
Enables Fixture Task To Process Fields Of Type UUID

### LINKED ISSUE #136: Baking a Fixture With Field Types of UUID Results in Empty Values in the Fixture Records
As an example this table:

```sql
CREATE TABLE `generic_joins` (
  `id` char(36) NOT NULL COMMENT 'UUID',
  `foreign_id` char(36) NOT NULL DEFAULT '',
  `created` datetime DEFAULT NULL,
  `creator_id` char(36) DEFAULT NULL,
  `modified` datetime DEFAULT NULL ,
  `modifier_id` char(36) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

Running the normal bake command:

```bash
$ ./bin/cake bake fixture GenericJoins
```

Generates this file: https://gist.github.com/jtyost2/c164dfbf898d18d05918

Which generates sample records that lack a populated entry for uuid fields.

This can be seen that they are indeed correctly registered as a `uuid` field type in the `$fields` property: https://gist.github.com/jtyost2/c164dfbf898d18d05918#file-genericjoinsfixture-php-L20 and that the field in the records are set as empty strings: https://gist.github.com/jtyost2/c164dfbf898d18d05918#file-genericjoinsfixture-php-L43 .

I think this is caused due to the `FixtureTask` not understanding how to process fields of type `uuid`: https://github.com/cakephp/bake/blob/master/src/Shell/Task/FixtureTask.php#L329

jtyost2:
Add Switch case to support uuid field types in FixtureTask
Adds Unit Test by modifying the Datatypes Fixture to include a uuid test case
Verifies both that the fields uses the uuid type and that the records array
generates a regexp verified uuid field

Closes #136

### DIFF
--- a/src/Shell/Task/FixtureTask.php
+++ b/src/Shell/Task/FixtureTask.php
@@ -367,6 +367,9 @@ class FixtureTask extends BakeTask
                         $insert .= " feugiat in taciti enim proin nibh, tempor dignissim, rhoncus";
                         $insert .= " duis vestibulum nunc mattis convallis.";
                         break;
+                    case 'uuid':
+                        $insert = Text::uuid();
+                        break;
                 }
                 $record[$field] = $insert;
             }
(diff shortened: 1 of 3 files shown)


markstory:
Thanks.

PR #138, opened 2015-07-25
Correct plugin generation for vendored plugins.

markstory:
Add additional template variables and templates to handle vendor plugins correctly.

Refs cakephp/cakephp#7062

### DIFF
--- a/src/Shell/Task/PluginTask.php
+++ b/src/Shell/Task/PluginTask.php
@@ -152,10 +152,21 @@ class PluginTask extends BakeTask
      */
     protected function _generateFiles($pluginName, $path)
     {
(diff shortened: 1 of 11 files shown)


antograssiot:
@markstory didn't tested it yet but it looks strange to me that the `_modifyAutoloader()` function are unchanged because the main application composer.json would need to be update to `Company\\Example\\ `?

markstory:
@antograssiot I missed that as its not in the test cases. I'll get that updated tonight.

markstory:
@antograssiot Fixed :smile:

antograssiot:
Looks good to me :+1:

lorenzo:
Great job!

PR #139, opened 2015-07-29
Visual break for behaviours

davidyell:
I think it is far easier to read with a visual break between the table setup, behaviour and associations.


bcrowe:
The test comparison templates need updated. Also, I'd only include the 2nd line break if `$associations` isn't empty.

davidyell:
Righto, will fix!

davidyell:
I haven't forgotten, just had a busy week.

bcrowe:
Awesome! Thanks! :+1:

PR #143, opened 2015-08-03
allow tests to be bake for mailer

antograssiot:


antograssiot:
the postgres failure is not related

PR #147, opened 2015-08-13
Insert fields in PhpDoc of the Entity

gsmit1996:


src/Template/Bake/Model/entity.ctp
```diff
@@ -20,6 +20,9 @@ use Cake\ORM\Entity;
 
 /**
  * <%= $name %> Entity.
+<% foreach ($fields as $field): %>
+ * @property $<%= $field %>
+<% endforeach; %>
```
markstory: There are tests that check the generated files. Those will need to be updated. The travis.ci build has the details of which comparison files need changes.

src/Template/Bake/Model/entity.ctp
```diff
@@ -20,6 +20,9 @@ use Cake\ORM\Entity;
 
 /**
  * <%= $name %> Entity.
+<% foreach ($fields as $field): %>
```
ndm2: Using `$fields` won't work properly as its purpose was to hold the mass assignable fields for the `$_accessible` property, it excludes `created`, `modified`, `updated` and primary key fields, and respects the (**[now kinda superflous](https://github.com/cakephp/bake/pull/134)** but still existent) [**`no-fields`**](https://github.com/cakephp/bake/blob/1.0.11/src/Shell/Task/ModelTask.php#L877) and [**`fields`**](https://github.com/cakephp/bake/blob/1.0.11/src/Shell/Task/ModelTask.php#L880) options, ie the `$fields` variable will not include all possible fields, and might even be empty.

I think either `$fields` should be used again to define the accessible fields, but this time only in case the `fields` or `no-fields` option was used, or the options should be removed and **[`$fields` should be changed](https://github.com/cakephp/bake/blob/1.0.11/src/Shell/Task/ModelTask.php#L438)** to hold _all_ existing fields.

markstory:
Any chance this could be updated so the tests pass?

ndm2:
I can take care of this whole thing in case @gsmit1996 isn't going to / cannot work on this further? I've started with entity property hints this some time ago, but put them on hold because I couldn't find a proper way to detect the possible datatypes for custom database types, and I was hell-bent on having either everything typed, or nothing.

markstory:
@ndm2 That would be great.

lorenzo:
Closing as there was another pull request taking care of this

Issue #146, opened 2015-08-13
Fixture default should be one demo record


dereuromark:
When running `cake bake model ...` it also creates the fixture with 1 demo entry.
After running again `cake bake fixture ...` at some later time it suddenly creates 10 records.

It should also default to 1 here for consistency until told otherwise.
The diff becomes huge without reason and we still only need a single record to do basic find operations for testing etc. No need for these huge demo fixture arrays defined.

markstory:
When calling FixtureTask through the CLI, the `count` option gets defaulted to 10, which is where the 10 generated records are coming from. Perhaps the default value should be 1 instead?

dereuromark:
I vote for that, yes.

markstory:
ok.

markstory:
Pull request up.

PR #148, opened 2015-08-14
Generate one row by default.

markstory:
Instead of 10 rows, we should generate 1. This makes the fixtures
generated by `bake model` and `bake fixture` consistent.

Refs #146


dereuromark:
Cool ;)
I would have done it this weekend, but you were faster - as usual.

PR #149, opened 2015-08-15
fix `bake controller all` and `bake template all`

antograssiot:
Actually the tables declared to be skipped are skipped only when bakin all models but still used when using the following command : 
```

bin/cake bake controller all
bin/cake bake template all
```

I also added the `phinxlog` table to the skipped table and also add PHP7.0 to the travis build.


markstory:
Nice work.

PR #151, opened 2015-08-19
Pass object to view for custom hooks

dereuromark:
For custom bake template hooks I need the current model object.
Since it already exists it would be easy to pass it down, as well.
One could also freshly create it in the template itself, but I think this little enhancement would not hurt.


ionas:
:+1: Helps with customization!

markstory:
Seems useful to me.

PR #152, opened 2015-08-19
Write bake's intermediary files inside "bake" subfolder.

### LINKED ISSUE #126: Generate temporary files in tmp/bake.
Currently bake generates it's intermediary/temporary files directly under `tmp` folder. Would be nice if it so under `tmp/bake` instead.

ADmad:
Closes #126

### DIFF
--- a/src/View/BakeView.php
+++ b/src/View/BakeView.php
@@ -63,6 +63,14 @@ class BakeView extends View
     ];
 
     /**
+     * Path where bake's intermediary files are written.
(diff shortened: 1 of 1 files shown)


markstory:
I'm not sure what is up with postgres. But tests pass locally with postgres 9.4.2.

antograssiot:
@markstory also passes locally for me. I've tried to force travis to use postgres 9.4 but it's still failling

PR #153, opened 2015-08-20
Fixtures generation is now using filtered table

antograssiot:
follows #149 

- fix the list generated by `bin/cake bake model`
- filter the tables to generate fixtures
- use the same table list to be skipped when using `bin/cake bake all --everything`


lorenzo:
Good job!

Issue #58, opened 2015-03-18
Integrate prefix option for 'bake all'


DSchalla:
The bake plugin is missing the 'prefix' Option for bake all, I will implement that in an PR.

Just note that down that this is already taken and WIP.

markstory:
Sounds great :+1:

andtxr:
Going to send a pr soon.

DSchalla:
@andtxr Thanks, since I had to pause the project where I needed this I totally forgot about it.

andtxr:
:+1:

dereuromark:
Closing as PR is now open.

PR #154, opened 2015-08-20
Enable prefix usage with `cake bake all`

andtxr:


andtxr:
Refers to #58

tests/TestCase/Shell/BakeShellTest.php
```diff
@@ -87,7 +87,7 @@ public function testAllWithModelName()
             ->with('Comments');
 
         $this->Shell->connection = '';
-        $this->Shell->params = [];
+        $this->Shell->params = ['prefix' => 'account'];
```
antograssiot: @andtxr The PR looks good but this doesn't look to really affect the actual test ? Am I missing something ?

andtxr: @antograssiot If the prefix param was passed before this PR, an exception would rise referring an invalid param. The new prefix option added to the parser covers it.

antograssiot: @andtxr well I'm pretty sure that I updated this line without the fix and it still works but nevermind

andtxr: So it is something to check again... This is not the behavior that I was expecting. Tonight I can check it better! Thanks @antograssiot   :+1:

markstory:
Looks good. Was all that was needed was to capture and forward the parameter?

antograssiot:
yep that's all

PR #159, opened 2015-09-04
Entity property hints

ndm2:
This is a first implementation of baking entity property type hints.

It includes types for known datbase-types and associations. Custom/Unknown database-types will be ignored for now, as I can't think of any realiable way to figure the possible types.

Properties that belong to `belongsTo` associations are being inserted after the corresponding foreign key, I thought that might be nice, but I'm not sure if it's really overly useful, since all other association properties will just be appended.

For the last two hours I've been stewing over where to put what, and I really couldn't decide, as everything seems to have its pros and cons. For now a property "schema" is set for the view, and a bunch of logic goes into the template to build a map of property names and docblock types. Theoretically some of the logic could be moved in the bake helper, this could reduce the code, like

```php
$properties = [];
if (!empty($propertySchema)) {
    foreach ($propertySchema as $property => $info) {
        switch ($info['kind']) {
            case 'column':
                $properties[$property] = $this->Bake->columnTypeToDocType($info['type']);
                break;

            case 'association':
                $type = $this->Bake->entityTypeToDocType($info['type'], $info['association']);
				if ($info['association']->type() === Association::MANY_TO_ONE) {
                    $property = $info['association']->property();
                    $index = array_search($info['association']->foreignKey(), array_keys($properties));
                    if ($index !== false) {
                        $properties = array_merge(
                            array_slice($properties, 0, $index + 1),
                            [$property => $type],
                            array_slice($properties, $index + 1, null)
                        );
                    } else {
                        $properties[$property] = $type;
                    }
                } else {
                    $properties[$property] = $type;
                }
                break;
        }
    }
}
```

and ditching the "insert after foreign key" feature for `belongsTo` associations would reduce the code to a minimum.

On the other hand this seems all very tightly entity docblock related, so having this in the global helper feels kinda akward. Moving everything into the task would also be an option, but again I'm not sure whether this would really be of any use outside of the entity view template...

The struggle is real  :cold_sweat:


src/Template/Bake/Model/entity.ctp
```diff
@@ -13,6 +13,77 @@
...
+                        );
+                    } else {
+                        $properties[$property] = $type;
+                    }
+                } else {
+                    $properties[$property] = $type;
+                }
+                break;
+        }
+    }
```
markstory: This seems like the kind of thing that could go into a helper. Having a helper would make it easier to use in custom templates that people might have.

ndm2: You mean the whole thing? Mmm, I'll see if I can come up with something that is ideally a little less entity specific.

markstory: Its ok if its entity specific and in the helper. The helper would just be to help reduce duplication in custom templates.

josegonzalez: We could have a custom `BakeModel` helper with model-layer specific stuff.

ndm2:
I've moved the logic into a new DocBlock helper. A helper dedicated to this doesn't seem too unreasonable. I have plans for more type hint related stuff in for example view templates, so I can see me adding more stuff in there.

Not sure whether this should have additional tests, as the helper is already being tested implicitly?

src/View/Helper/DocBlockHelper.php
```diff
@@ -0,0 +1,143 @@
...
+                return 'bool';
+
+            case 'binary':
+                return 'string|resource';
+
+            case 'date':
+            case 'datetime':
+            case 'time':
+            case 'timestamp':
+                return 'string|int|\DateTime|\Cake\I18n\Time';
```
lorenzo: I would be more comfortable if the typehint is just `\Cake\I18n\Time`

ndm2: @josegonzalez  But these are all accepted types for date/time based columns. Are there plans for restricting them, or what exactly do you mean by "comfortable"?

src/Template/Bake/Model/entity.ctp
```diff
@@ -13,6 +13,13 @@
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 
+use Cake\ORM\Association;
+
+$propertyHintMap = null;
```
markstory: This should default to an array. Then you wouldn't need the if block I don't think.

ndm2: @markstory That would result in an empty line after the description in case no properties are being available. Should be an edge case in userland I guess, but tests would need to adapt to this behavior. I can change it if you're OK with that?

markstory: Its not that important either way I guess.

src/View/Helper/DocBlockHelper.php
lorenzo: I mean, by default the ORM will only set `Time` objects in those properties, so it is easier to say that you can only find that there. When accepting data from a form, the value will also be converted to a `Time` object before it is set to the entity. So I would say that the cases where it is not a `Time` objects are very rare

ndm2: @lorenzo  That's true, I was just a little hesitant as the hints are only for these rare cases anyways. Setter wise they are only ever used when setting data manually, and in that case, using Integers, Strings or `DateTime` instances would "fail" (as in marked as errors). Personally I would be totally fine with always using `Time` instances when setting data, and I can see how it's less "confusing" for people that don't know about the internals, ie that by default on "read" it's always a `Time` instance, and that the other types are the ones that are accepted on "write" too.

src/Template/Bake/Model/entity.ctp
ndm2: Ok, then I'll leave it like that for now, just for the sake of having less to consider when creating tests, and for cleaner looking comparison files :)

lorenzo:
Great job! :clap:

PR #160, opened 2015-09-21
BakeHelper.php optionize indent string

schrolli:
Make the indent-string an option to give users, who want to use their own Theme, the ability to justify the indentation, e.g. indent by tab instead of four spaces.

### DIFF
--- a/src/View/Helper/BakeHelper.php
+++ b/src/View/Helper/BakeHelper.php
@@ -64,7 +64,7 @@ class BakeHelper extends Helper
         $options += [
             'indent' => 2,
             'tab' => '    ',
-            'commaAtEnd' => false,
+            'trailingComma' => false,
         ];
 
         if (!$list) {
@@ -87,7 +87,7 @@ class BakeHelper extends Helper
             $end = "\n" . str_repeat($options['tab'], $options['indent'] - 1);
         }
         
-        if ($options['commaAtEnd']) {
+        if ($options['trailingComma']) {
             $end = "," . $end;
         }
 
--- a/src/View/Helper/BakeHelper.php
+++ b/src/View/Helper/BakeHelper.php
@@ -62,7 +62,8 @@ class BakeHelper extends Helper
     public function stringifyList(array $list, array $options = [])
     {
         $options += [
-            'indent' => 2
+            'indent' => 2,
+            'tab' => '    ',
         ];
 
         if (!$list) {
@@ -80,9 +81,9 @@ class BakeHelper extends Helper
         $join = ', ';
         if ($options['indent']) {
             $join = ',';
-            $start = "\n" . str_repeat("    ", $options['indent']);
+            $start = "\n" . str_repeat($options['tab'], $options['indent']);
             $join .= $start;
-            $end = "\n" . str_repeat("    ", $options['indent'] - 1);
+            $end = "\n" . str_repeat($options['tab'], $options['indent'] - 1);
         }
 
         return $start . implode($join, $list) . $end;
--- a/src/View/Helper/BakeHelper.php
+++ b/src/View/Helper/BakeHelper.php
@@ -64,6 +64,7 @@ class BakeHelper extends Helper
         $options += [
             'indent' => 2,
             'tab' => '    ',
+            'comma_at_end' => false,
         ];
 
         if (!$list) {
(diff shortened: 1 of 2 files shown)


dereuromark:
:+1: Nice!

schrolli:
what are these checks? did i made an error?

markstory:
@schrolli The checks are from the various CI services we use. AppVeyor failed because of heap corruption. And the postgres tests have been failing for a while and I don't know why. I don't think your changes caused either of these builds to fail directly.

schrolli:
ah, ok. was just wondering;)

markstory:
@schrolli Could you add a test case for this? Having that would help ensure we don't accidentally break this feature in the future.

schrolli:
i added an another option and added some testcases to cover pretty much all the options

src/View/Helper/BakeHelper.php
```diff
@@ -62,7 +62,9 @@ public function arrayProperty($name, array $value = [], array $options = [])
     public function stringifyList(array $list, array $options = [])
     {
         $options += [
-            'indent' => 2
+            'indent' => 2,
+            'tab' => '    ',
+            'comma_at_end' => false,
```
markstory: Should be commaAtEnd.

src/View/Helper/BakeHelper.php
```diff
@@ -80,9 +82,13 @@ public function stringifyList(array $list, array $options = [])
...
             $join = ',';
-            $start = "\n" . str_repeat("    ", $options['indent']);
+            $start = "\n" . str_repeat($options['tab'], $options['indent']);
             $join .= $start;
-            $end = "\n" . str_repeat("    ", $options['indent'] - 1);
+            $end = "\n" . str_repeat($options['tab'], $options['indent'] - 1);
+        }
+        
+        if ($options['comma_at_end']) {
+            $end = ",".$end;
```
markstory: Spaces around operators.

tests/TestCase/View/Helper/BakeHelperTest.php
```diff
@@ -119,4 +119,105 @@ public function testAliasExtractorBelongsToMany()
...
+     */
+    public function testStringifyListDefaults()
+    {
+        $list = ['one' => 'foo', 'two' => 'bar', 'three'];
+        $result = $this->BakeHelper->stringifyList($list);
+        $spaces = '    ';
+        $expected = "\n".
+            $spaces.$spaces."'one' => 'foo',\n".
+            $spaces.$spaces."'two' => 'bar'\n".
+            $spaces.$spaces."'three'\n".
```
markstory: Spaces around operators improve readability and consistency with the rest of the project.

schrolli:
oh, i wasn't aware of concatenations and psr-2... now all dots should have spaces and the option is renamed;)

markstory:
Looking better, but [some of the tests are failing](https://travis-ci.org/cakephp/bake/jobs/81888566)

schrolli:
bump

AD7six:
@schrolli There are still phpcs fails - click on the red x.

schrolli:
ah, now the ci-job runs through, hadn't saw the error in the comments of the script...

src/View/Helper/BakeHelper.php
```diff
@@ -62,7 +62,9 @@ public function arrayProperty($name, array $value = [], array $options = [])
     public function stringifyList(array $list, array $options = [])
     {
         $options += [
-            'indent' => 2
+            'indent' => 2,
+            'tab' => '    ',
+            'commaAtEnd' => false,
```
markstory: `trailingComma` might be a better name for this.

schrolli:
so the appveyor-error is related to the core with some sql-errors and i can't do something about it...

the travis-ci test is passing

what i do not understand: the decreased coverage at coveralls. The decrease is located at the end of the default-config array. has it possibly something to do with the trailing comma in the line before? but is a trailing comma bad practice then?

dereuromark:
I wouldn't wry about any one of those two. Looks all good.

Personally, I think trailing comma can be a good practice if enforced automatically via cs (not manually), as it helps to keep merge diffs to a minimum without additional efforts for the developer. But that is out of scope of the actual PR here :P

PR #161, opened 2015-09-23
Attempt to get travis passing.

markstory:
Skip the postgres builds for now as I'm not able to reproduce the aliasing issues we're seeing on travisci.


ADmad:
The perpetually failing appveyor build is an annoyance too. GH's mobile UI doesn't allow merging a PR unless all webhooks return success status.

markstory:
Appveyor failing seems to be an issue with the PHP we're using. Perhaps we need to update?

antograssiot:
@markstory the failling tests is passing when filtered isolated. We could try to update php, it's currently a 5.4.x

PR #165, opened 2015-10-06
set $associations @property to \Cake\ORM\Table

ayman-alkom:
this is helpful for the code completion.
it is more likely if we can set the path to something like:    `\App\Model\Table\ModelTable`

### DIFF
--- a/src/Template/Bake/Model/table.ctp
+++ b/src/Template/Bake/Model/table.ctp
@@ -36,7 +36,7 @@ echo implode("\n", $uses);
  *
 <% foreach ($associations as $type => $assocs): %>
 <% foreach ($assocs as $assoc): %>
- * @property \Cake\ORM\Association\<%= Inflector::camelize($type) %> $<%= $assoc['alias'] %>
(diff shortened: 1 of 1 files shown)


src/Template/Bake/Model/table.ctp
```diff
@@ -36,7 +36,7 @@ echo implode("\n", $uses);
  *
 <% foreach ($associations as $type => $assocs): %>
 <% foreach ($assocs as $assoc): %>
- * @property \Cake\ORM\Association\<%= Inflector::camelize($type) %> $<%= $assoc['alias'] %>
+ * @property \Cake\ORM\Table $<%= $assoc['alias'] %>
```
ADmad: This is incorrect. Associations are instance of `Cake\ORM\Association` not `Cake\ORM\Table`.

markstory:
Closing as this makes the docs incorrect. @ADmad is right about the properties being association classes. These objects use __call to forward unknown methods to the connected table object.

PR #167, opened 2015-10-17
Replace incorrect tr with td in view template

chronon:
Replaces an incorrect closing `</tr>` with a correct `</td>`.


markstory:
Tests need to be updated, but I can do that post merge.

PR #168, opened 2015-10-21
Fix docblock

okinaka:
The delete() returns \Cake\Network\Response object or null.


markstory:
Test failures seem unrelated. I'll address those separately.

PR #171, opened 2015-10-30
doc return tag sync with code

luke83:
`bake()` and `bakeTest()` return type is a string (baked file contents), so doc tag return type should be string.

### DIFF
--- a/src/Shell/Task/SimpleBakeTask.php
+++ b/src/Shell/Task/SimpleBakeTask.php
@@ -90,7 +90,7 @@ abstract class SimpleBakeTask extends BakeTask
      * Generate a class stub
      *
      * @param string $name The classname to generate.
-     * @return void
+     * @return string
      */
     public function bake($name)
     {
@@ -109,7 +109,7 @@ abstract class SimpleBakeTask extends BakeTask
      * Generate a test case.
      *
      * @param string $className The class to bake a test for.
-     * @return void
+     * @return string
      */
     public function bakeTest($className)
     {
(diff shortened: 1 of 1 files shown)


src/Shell/Task/SimpleBakeTask.php
```diff
@@ -109,7 +109,7 @@ public function bake($name)
      * Generate a test case.
      *
      * @param string $className The class to bake a test for.
-     * @return void
+     * @return string
```
HavokInspiration: Actually, ``TestTask::bake()`` can return false as well.

luke83: Then it should be `@return string|bool`, right?

HavokInspiration: Yes

HavokInspiration: Which makes the case where the ``no-test`` param is true, the method should return ``false`` as well (line 117 below)

luke83: Fixed in 29e3173

src/Shell/Task/SimpleBakeTask.php
```diff
@@ -90,7 +90,7 @@ public function main($name = null)
      * Generate a class stub
      *
      * @param string $name The classname to generate.
-     * @return void
+     * @return string
```
dereuromark: this method has a `return;` => that should actually be `return null;` then and the doc block should also say `|null`.

markstory: $contents is a string generally.

dereuromark: Yeah i meant bakeTest()

PR #172, opened 2015-10-31
Implementation of Index Column Limitation (Closes #65)

### LINKED ISSUE #65: Cake 3.0 bake index template limits to first 7 fields
It took me hours to figure out why the eighth field of my users table wasn't showing up in index views.
Template/index.ctp  creates a Collection of fields, but chains on ->take(7) at the end to limit the list, presumably so index tables don't get too wide for screens.  Wouldn't it be better to show them all and offer an optional limit value, or better yet specify a list of fields to show/omit?  In 2.x you just went into the baked view and deleted any `<th>`'s and `<td>`'s you didn't want.  Now you have to pop the hood on the form template and discover and remove the hard-coded (7).  Good luck to newbies!
Either way, arbitrarily, and secretly, lopping off everything after the first 7 fields is not the friendliest Bake feature.

DSchalla:
As discussed in #65, this PR implements a flag to set the index number limitation to any value while keeping the default value of 7. 

I would welcome any feedback regarding the implementation + tests, I am uncertain whether its the "best practice".

### DIFF
--- a/src/Shell/Task/TemplateTask.php
+++ b/src/Shell/Task/TemplateTask.php
@@ -91,7 +91,7 @@ class TemplateTask extends BakeTask
     /**
      * Maximal Index Columns
      *
-     * @var integer
+     * @var int
      */
     public $maxIndexColumns = 7;
 
@@ -148,7 +148,7 @@ class TemplateTask extends BakeTask
             $action = $this->template;
         }
         if (!empty($this->params['index-columns'])) {
-            $this->maxIndexColumns = (int) $this->params['index-columns'];
+            $this->maxIndexColumns = (int)$this->params['index-columns'];
         }
         if ($action) {
             return $this->bake($action, true);
--- a/src/Shell/Task/TemplateTask.php
+++ b/src/Shell/Task/TemplateTask.php
@@ -402,8 +402,8 @@ class TemplateTask extends BakeTask
             return false;
         }
 
-        if ($action === "index" && !empty($this->params['indexColumns'])) {
-            $this->BakeTemplate->set('indexColumns', $this->params['indexColumns']);
+        if ($action === "index" && !empty($this->params['index-columns'])) {
+            $this->BakeTemplate->set('indexColumns', $this->params['index-columns']);
         }
 
         $this->BakeTemplate->set('action', $action);
--- a/src/Shell/Task/TemplateTask.php
+++ b/src/Shell/Task/TemplateTask.php
@@ -89,13 +89,6 @@ class TemplateTask extends BakeTask
     public $noTemplateActions = ['delete'];
 
     /**
-     * Maximal Index Columns
-     *
-     * @var int
-     */
-    public $maxIndexColumns = 7;
-
-    /**
      * AssociationFilter utility
      *
      * @var AssociationFilter
@@ -146,9 +139,6 @@ class TemplateTask extends BakeTask
         }
         if (!$action) {
             $action = $this->template;
-        }
-        if (!empty($this->params['index-columns'])) {
-            $this->maxIndexColumns = (int)$this->params['index-columns'];
         }
         if ($action) {
             return $this->bake($action, true);
@@ -412,8 +402,8 @@ class TemplateTask extends BakeTask
             return false;
         }
 
-        if ($action === "index") {
-            $this->BakeTemplate->set('indexColumns', $this->maxIndexColumns);
+        if ($action === "index" && !empty($this->params['indexColumns'])) {
+            $this->BakeTemplate->set('indexColumns', $this->params['indexColumns']);
         }
 
         $this->BakeTemplate->set('action', $action);
@@ -445,7 +435,8 @@ class TemplateTask extends BakeTask
         ])->addOption('prefix', [
             'help' => 'The routing prefix to generate views for.',
         ])->addOption('index-columns', [
-            'help' => 'Limit for the number of index columns (Default: 7)'
+            'help' => 'Limit for the number of index columns',
+            'default' => 0
         ])->addSubcommand('all', [
             'help' => '[optional] Bake all CRUD action views for all controllers. Requires models and controllers to exist.'
         ]);
--- a/src/Shell/Task/TemplateTask.php
+++ b/src/Shell/Task/TemplateTask.php
@@ -89,6 +89,13 @@ class TemplateTask extends BakeTask
     public $noTemplateActions = ['delete'];
 
     /**
+     * Maximal Index Columns
+     *
+     * @var integer
+     */
+    public $maxIndexColumns = 7;
+
+    /**
      * AssociationFilter utility
      *
      * @var AssociationFilter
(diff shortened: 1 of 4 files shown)


src/Shell/Task/TemplateTask.php
```diff
@@ -89,6 +89,13 @@ class TemplateTask extends BakeTask
     public $noTemplateActions = ['delete'];
 
     /**
+     * Maximal Index Columns
+     *
+     * @var integer
```
dereuromark: int

src/Shell/Task/TemplateTask.php
```diff
@@ -140,6 +147,9 @@ public function main($name = null, $template = null, $action = null)
         if (!$action) {
             $action = $this->template;
         }
+        if (!empty($this->params['index-columns'])) {
+            $this->maxIndexColumns = (int) $this->params['index-columns'];
```
dereuromark: no space after casts

dereuromark:
Personally I would like to have 0 as default (no limit).
But I guess people here love their 7 default, so we will have to keep it.

DSchalla:
I don't mind actually. I would personally also prefer 0. Thanks for the other feedback!

src/Shell/Task/TemplateTask.php
```diff
@@ -89,6 +89,13 @@ class TemplateTask extends BakeTask
     public $noTemplateActions = ['delete'];
 
     /**
+     * Maximal Index Columns
```
markstory: Maximum :bike:

src/Shell/Task/TemplateTask.php
```diff
@@ -140,6 +147,9 @@ public function main($name = null, $template = null, $action = null)
         if (!$action) {
             $action = $this->template;
         }
+        if (!empty($this->params['index-columns'])) {
```
markstory: You could use `$this->param()` instead here.

src/Shell/Task/TemplateTask.php
```diff
@@ -431,6 +446,8 @@ public function getOptionParser()
             'help' => 'The controller name if you have a controller that does not follow conventions.'
         ])->addOption('prefix', [
             'help' => 'The routing prefix to generate views for.',
+        ])->addOption('index-columns', [
+            'help' => 'Limit for the number of index columns (Default: 7)'
```
markstory: You can set the default in the option definition. Then the help formatter will output the default for you.

markstory: Also the option parser will set the default :smile:

src/Shell/Task/TemplateTask.php
```diff
@@ -140,6 +147,9 @@ public function main($name = null, $template = null, $action = null)
         if (!$action) {
             $action = $this->template;
         }
+        if (!empty($this->params['index-columns'])) {
+            $this->maxIndexColumns = (int)$this->params['index-columns'];
```
markstory: What does the property do that the param index does not? Seems like you could just use the param value.

ionas:
This issue can easily be solved by one line of CSS:
```css
tr > :nth-child(n+8) { display: none }
```
http://codepen.io/anon/pen/BobRdK

lorenzo:
@ionas what about the action column? The last one

ionas:
@lorenzo fixed that one (see above).

:not(.actions) will work too in case the last column has that classname.

lorenzo:
@ionas how would users opt out of that? It is quite difficult to see the reason the columns are hidden

ionas:
I prefer the cake bake templates to hide no columns in its HTML structure as it can easily be verified that all fields are existent. As the current implementation is a hack around user interface issues handling a undefined quantity of columns per table, the small CSS change can easily be altered to adapt the already backed output to your needs (either through more CSS or through removing those columns by hand that one deems superfluous).

dakota:
> through removing those columns by hand that one deems superfluous

Half the point of bake is to not need to add or remove things by hand.

I'm :+1: for this change, but with a default of 0. Then those who want the columns gone can set the appropriate option, and those who just want to hide them can set the appropriate CSS.

jhli:
Aggreed.

DSchalla:
Sorry for slacking on fixing the discussion points, I personally agree on changing the default to unlimited. Opinions?

dereuromark:
Having this option 0 as default makes the most sense.

josegonzalez:
Can we warn if the table is including more than N columns at least? That way a user can discover this feature and customize at will :)

src/Shell/Task/TemplateTask.php
ionas: Option should probably be named max-table-columns :)
I don't know if it does already but it should then also be respected by view:related tables.

dereuromark:
Can we continue and merge this please? :)

ionas:
:+1: @dereuromark ping @DSchalla ;-)

DSchalla:
@dereuromark I will have more time after the 15th Februrary and can implement the missing things after that. I hope it can wait till then.

DSchalla:
Seems both AppVeyor and Travis have atm some issues.

antograssiot:
@DSchalla the appveyor build should be fixed with #228.
I can't help with the github authentication issue cc @lorenzo @markstory

tests/TestCase/Shell/Task/TemplateTaskTest.php
```diff
@@ -446,6 +446,25 @@ public function testBakeIndex()
...
+     * test bake template with index limit overwrite
+     *
+     * @return void
+     */
+    public function testBakeIndexWithIndexLimit()
+    {
+        $this->Task->controllerName = 'TemplateTaskComments';
+        $this->Task->modelName = 'TemplateTaskComments';
+        $this->Task->controllerClass = __NAMESPACE__ . '\TemplateTaskCommentsController';
+        $this->Task->params['indexColumns'] = 3;
```
markstory: How does this inflection happen? The OptionParser won't automatically inflect `index-columns` into `indexColumns`

HavokInspiration: @markstory It happens here : https://github.com/DSchalla/bake/blob/index-columns/src/Shell/Task/TemplateTask.php#L405-L407

markstory: But even there, the `$this->params['indexColumns']` key will never be populated by the option parser.

HavokInspiration: Yes I completely agree with you.
I believe it's a just a mistake that was not caught because the tests pass since it's written the same both in the test and in the code. 
It should be ``index-columns`` and it will work as expected.

DSchalla: Thanks @markstory & @HavokInspiration, it was indeed a mistake after removing the parameter, I will unify the naming.

DSchalla:
Appveyor still broken sadly, Travis worked on a few runs, some failures are either related to a test not related to my PR or to Github denying access again....

markstory:
Looks good to me. Test failures are related to CI environments not your code.

DSchalla:
Thanks, sorry for the long delay with the changes.

markstory:
No problem. Thanks for taking the time to circle back on them.

jhli:
Is this now as well limiting associated table lists in simge views?

Thanks for doing this!

DSchalla:
Right now it limits only the index.ctp

PR #174, opened 2015-11-02
block doc return type

luke83:
All() method returns true (correct shell usage and execution) or false (wrong shell usage)


dereuromark:
Looks good!

Issue #155, opened 2015-08-29
Template baking with Tree behavior should omit lft/rght fields in form


dereuromark:
The lft/right fields in forms should not be displayed as the TreeBehavior manages them on its own.
When baking a model, the Tree behavior is automatically attached.
I would expect the view to also detect the desired tree behavior, or that the Tree behavior is attached, and then skips those two fields from being displayed.

Currently they show and will likely destroy the tree on submit.

It would also be useful to skip those fields for index, as they bear no meaning here, either.

markstory:
Pull request up now.

PR #176, opened 2015-11-04
Exclude lft/rght from baked output.

markstory:
When baking forms and list pages for tree models, we should exclude the lft/rght fields. The treebehavior manages these fields on add/update and they are not very useful on the list page.

Refs #155


antograssiot:
looks good :+1:

lorenzo:
Nice!

Issue #177, opened 2015-11-05
Baking with nested prefixes not working correctly


ndm2:
Using nested prefixes will fail/create unexpected results depending on the prefix format.

Using a format like `pre/fix` will result in a namespace like `Pre/fix` with a matching folder structure, ie a `Pre` folder containing `fix` folder.

Using a format like `pref\fix` ends up with a valid namespace separator, however this may fail on *nix systems unless `\` is being replaced with a valid directory separator somewhere.

Not time for a PR currently, but I can give it a try in the next few days in case no one else is going to take care of it in the mean time.

markstory:
I can take a look.

markstory:
Pull requests up now.

PR #178, opened 2015-11-07
Fix nested prefix handling

markstory:
Addresses how nested prefixes are handled in controller and template generation.

Refs #177


tests/test_app/App/Controller/Admin/Management/PostsController.php
```diff
@@ -0,0 +1,34 @@
...
+ * Posts Controller class.
+ *
+ * For testing prefix routing.
+ */
+class PostsController extends Controller
+{
+    /**
+     * index action
+     *
+     * @return void
```
dereuromark: We should probably correct those to allow for response returns. void would not be quite correct.

lorenzo:
Nice job :+1:

Issue #164, opened 2015-10-01
Is it possibile create a bake without add?


vnt:
I have created my temple for bake in plugins/MyBake/src/Template/Bake/Template/
Is there a way to create only:
- index
- edit
- view

without "add" in controller and templates?

ionas:
If it is not in the book, you should add that AFAIK:
- For any action in your controller bake is looking for templates
- Thus if you either remove actions in your controller before backing view action templates OR if you change the controller bake templates to only render/create certain actions, then you should only have those action view templates
- Don't forget that if the view templates are not index/edit/view etc, you might need to add those templates manually (in Console somewhere, check again the book)
If this solves your issue AND if the book does not mention all that (please check), then please add add a PR to cakephp/docs so that in future peeps have an easier time.

markstory:
No its not possible. Perhaps in the future it will be.

josegonzalez:
@markstory you can set an event that only sets the `index`, `edit` and `view` actions when baking a controller, no?

markstory:
@josegonzalez Yes/

vnt:
Hello josegonzalez,
you can give me an example?

josegonzalez:
You can do something like the following to set custom actions for your controllers:

```php
use Cake\Event\Event;
use Cake\Event\EventManager;
use Cake\Utility\Hash;
EventManager::instance()->attach(function (Event $event) {
    $view = $event->subject;
    $name = Hash::get($view->viewVars, 'name');
    $isController = strpos($event->data[0], 'Bake/Controller/controller.ctp') !== false;
    if ($isController && $name == 'Users') {
        $view->viewVars['actions'] = ['login', 'logout', 'index', 'view', 'register', 'edit', 'delete'];
    }
}, 'Bake.beforeRender');
```

simkimsia:
@josegonzalez where would we place this code?

ionas:
@simkimsia another way to do it is to copy the templates from vendor cakephp and make changes, 3.x example with 2 bake themes:

<img width="387" alt="screen shot 2015-11-29 at 17 38 06" src="https://cloud.githubusercontent.com/assets/106845/11458429/04fbdf0a-96c0-11e5-896e-f35cfbc60a94.png">

Enable in Boostrap:
```php
if (Configure::read('debug')) {
    Plugin::load('DebugKit', ['bootstrap' => true]);
    Plugin::load('ReadOnlyBakeTheme');
[...]

PR #188, opened 2015-11-29
Add authentication-related templates

josegonzalez:
- [x] Login action template
- [x] Login view template
- [x] Logout action template


josegonzalez:
Anyone have any thoughts on this? They are pretty basic, but essentially pulled straight from the book auth tutorial.

src/Template/Bake/Element/Controller/login.ctp
```diff
@@ -0,0 +1,31 @@
...
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+%>
+
+    /**
+     * Login method
+     *
+     * @return void
+     */
+    public function login() {
```
antograssiot: opening braket on new line maybe to please the PSR2 :trollface:

lorenzo:
How do you bake those only for one controller?

josegonzalez:
Pretty sure this line already *only* sets the actions to the existing ones, and will properly exclude these two.

https://github.com/cakephp/bake/blob/master/src/Shell/Task/ControllerTask.php#L94

lorenzo:
@josegonzalez ok, so how do you bake the controller code then?

markstory:
@lorenzo We could add a `--actions` CLI option to list out the specific actions people want included. That could also be used to address #164

lorenzo:
@markstory that would make sense

markstory:
@josegonzalez Do you want to add the `--actions` flag? If not I can take care of it.

simkimsia:
:+1: for auth related bake templates

josegonzalez:
@markstory I'd rather you did since I'll be missing the test :(

markstory:
@josegonzalez Cool. I will get to it this week then.

josegonzalez:
@lorenzo atm you can just set them via a bake event.

markstory:
I've merged these templates as #198 adds the CLI flag to whitelist actions into a controller.

PR #190, opened 2015-11-29
Remove unneeded templates.

ADmad:


ADmad:
@jhli  Yes they will. These are just left overs from earlier moving around of templates.

PR #191, opened 2015-12-02
Adding table prefix handling

burzum:
This works as long as _all_ tables have the prefix in the selected DB.


src/Shell/Task/ModelTask.php
```diff
@@ -85,6 +85,14 @@ public function main($name = null)
         parent::main();
         $name = $this->_getName($name);
 
+        if (!empty($this->connection)) {
+            $connection = ConnectionManager::get($this->connection);
+            $config = $connection->config();
+            if (isset($config['prefix'])) {
+                $this->tablePrefix = $config['prefix'];
+            }
+        }
```
lorenzo: I would much rather provide the prefix via a command line option, instead of the connection config.

burzum: I have a trait for the fixture and the table that use it based on the connection as well.

lorenzo: Ok, but that is something bake should not care about. It is much cleaner and easier to test if you provide it yourself from the cli

ionas:
command line prefix allows mix of non prefixed and prefixed tables?

lorenzo:
@ionas it allows whatever meaning we convey to it

ionas:
Great, the original statement at the top says otherwise, that's the reason I do ask so specifically.

burzum:
@ionas well, @lorenzo wanted me to change it.

ionas:
I hope you can work with it, chaining bake commands with && or ;; and --force should do magic.
Thanks for this contribution!

markstory:
I feel like we should have at least one test related to this feature :smile:

src/Shell/Task/ModelTask.php
```diff
@@ -299,6 +300,10 @@ public function findTableReferencedBy($schema, $keyField)
                 if (!isset($constraintInfo['references'])) {
                     continue;
                 }
+                $length = strlen($this->tablePrefix);
```
dereuromark: I can already see people crying because they *must* use umlauts or special char prefixes `äöü` ^^.

burzum: It's a valid point when you have to use Cake on a legacy DB with for example German table names.

dereuromark: Yeah, better playing it safe using mb_ IMO.

burzum: Agreed :+1:

dereuromark:
Once the strlen() issue is fixed we can probably finalize and ship this maybe?

lorenzo:
Should we close this?

burzum:
Whats missing? Just the mb_strlen and a test?

dereuromark:
And rebasing.

burzum:
S

Issue #192, opened 2015-12-03
Cake bake fixture generates wrong import property


wvdongen:
Example: generating a fixture for the companies table: 
<code>bin/cake bake fixture companies --schema --import-records</code>

Results in a fixture with:
<code>public $import = ['model' => 'Companies', 'records' => true];</code>

According to https://github.com/cakephp/cakephp/blob/master/src/TestSuite/Fixture/TestFixture.php#L67 

Both array keys ‘model’ and ‘records’ do not seem to be used by cake. The ‘model’ key should be ‘table’ and there doesn’t seem to be support for importing records.

markstory:
Your right! I'll get that fixed.

markstory:
Pull request.

PR #193, opened 2015-12-04
Remove dead options from fixture task.

markstory:
The `import-records` option doesn't actually do anything anymore, so we don't need it. The generated import property is now correct as well.

Refs #192


lorenzo:
Tests failing in windows in an odd way :S

markstory:
Tests are passing now :smile:

Issue #182, opened 2015-11-20
Autodetect SQL UNIQUE KEY and generate isUnique()


ionas:
At least for me unique constraint detection did not work. **Should it?**

```sql
CREATE TABLE `manage_manufacturers` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `vc24_manufacturer_id` bigint(20) NOT NULL,
  `is_visible` tinyint(1) unsigned NOT NULL DEFAULT '1',
  PRIMARY KEY (`id`),
  UNIQUE KEY `vc24_manufacturer_id` (`vc24_manufacturer_id`)
) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8;
```
Expected in ```buildRules()```:
```php
$rules->add($rules->isUnique(['vc24_manufacturer_id']));
```

markstory:
I'll take a look.

ionas:
```bin/cake bake all --prefix Admin --force manage_manufacturers;``` just in case there is some unlikely interaction with any of those params.

markstory:
Closing as a pull request is up now.

jhli:
Thank you!

PR #196, opened 2015-12-10
Generate isUnique rules for single column unique key indexes.

markstory:
While it is possible to add rules for multi-column unique keys, doing so would require changing the template data, which will likely break userland templates. For that reason, I'd rather not add support for multi-column unique keys at this point in time.

Refs #182

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -650,7 +650,8 @@ class ModelTask extends BakeTask
         if (!empty($this->params['no-rules'])) {
             return [];
         }
-        $fields = $model->schema()->columns();
+        $schema = $model->schema();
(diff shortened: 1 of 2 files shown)


lorenzo:
Good job!

lorenzo:
I'll see why pho 7 is not happy in travis

jhli:
Thanks!

Out of interest: When 3.2 or 4.x hits multi column unique key constraint should be added?

markstory:
@jhli We'd have to break compatibility with previous versions of bake. That doesn't have to be timed with a change in a CakePHP version necessarily.

jhli:
Okay, I did open that ticket in case I get to bake improvements early next year it's not forgotten.

PR #198, opened 2015-12-12
Add option to selectively bake actions.

markstory:
This allows people to add/remove actions to controller through a CLI
flag. This helps people extend bake or make readonly controllers.

Refs #188
Refs #164


josegonzalez:
Semi-related feature: it would be great to remap old actions to new ones. For instance, `users/register` is basically `users/add`. Don't think it belongs in this PR specifically, but something I wanted to see potentially happen.

markstory:
Remapping could be done with templates couldn't it?

lorenzo:
Great job!

PR #199, opened 2015-12-14
Use non-deprecated method

josegonzalez:


josegonzalez:
Actually this might be wrong. Looks like that method still exists on the `Validator`, but is deprecated on the `Validation` class. That's a bit confusing, is there a reason its deprecated in one but not the other?

lorenzo:
It is 100% confusing, but the current code is correct. The method that was deprecated lives in the `Validation` class, not on the `Validator` class

dereuromark:
Can we maybe make that less confusing in the future?

Issue #202, opened 2015-12-21
1.1.2 bake all with --prefix tries to create table class in prefix path - fine in 1.1.1


ionas:
For details see: https://gist.github.com/ionas/90324d5e48a9c935853f
Not sure but this could be the issue: https://github.com/cakephp/bake/commit/214b2f13309f760f5cd6b9d502bd2437f7a9cd87#diff-0c9ca9e63bf919df5c6c12bcbc430ecfR108

Edit: Fixed the gist link showing the bug

markstory:
I can take a look.

markstory:
Pull request up now.

ionas:
I thank you very much.

PR #206, opened 2015-12-29
Fix prefixes being applied to models.

markstory:
Refs #202

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -875,6 +875,16 @@ class ModelTask extends BakeTask
     }
 
     /**
+     * Models never have routing prefixes applied.
+     *
+     * @return string
+     */
+    protected function _getPrefix()
+    {
+        return '';
+    }
+
+    /**
      * Get an Array of all the tables in the supplied connection
      * will halt the script if no tables are found.
      *
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ matrix:
     - php: 7.0
       env: RUN_CS=1 RUN_TESTS=0
 
-    - php: 5.4
+    - php: 5.6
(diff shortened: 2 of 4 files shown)


src/Shell/Task/ModelTask.php
```diff
@@ -875,6 +875,16 @@ public function listUnskipped()
     }
 
     /**
+     * Models never have routing prefixes applied.
+     *
+     * @return string
+     */
+    protected function _getPrefix()
```
ionas: Might be wise to rename this to ```getRoutingPrefix()``` to match the doc block. If table prefixes make it in --routing-prefix and --table-prefix should be the command line options (and the matching protected method names).

markstory: Renaming the method makes sense, changing the CLI flags around less so, as each prefix is contextual.

ionas: I can't follow?!
Example: ```bake all articles --routing-prefix admin --table-prefix wp_``` ?
Which context? Did this possibly become a bug because neither CLI flag nor matching protected method carried its context?

markstory: My mistake, I didn't consider the `bake all` context. I was only considering that `bake model` and `bake controller` were separate commands with non-overlapping argument sets. I'm not sure we want `bake all` to expose every option each subcommand exposes. I feel like that could end up making bake all overwhelming to use and complicated to maintain.

ionas: My vote goes to deprecate ```--prefix in favour of ```--routing-prefix``` throwing a warning to the CLI. The decision on if bake all supports anything beyond ```--theme``` and ```--force``` OR pass through all the flags can then be made later. Thank you for your great work :)!

Issue #204, opened 2015-12-24
HasMany detection vs HABTM detection


ionas:
From ```ModelTask.php```
```php
            // Exclude habtm join tables.
            $pattern = '/_' . preg_quote($tableName, '/') . '|' . preg_quote($tableName, '/') . '_/';
            $possibleJoinTable = preg_match($pattern, $otherTable);
            if ($possibleJoinTable) {
                continue;
            }
```
Now for me associations were missing between ```suppliers``` and ```managed_suppliers``` - somehow (and I don't know why or how) those associations indeed were detected correctly, *before* (e.g. on cake 3.1.4 and some corresponding bake).

I looked through bake 1.0.14 to 1.1.2 and it is always the same above. ```managed_suppliers``` is not being checked as a possible HasMany assocation because the matching removes it, as it considers it a possible join table.

Do I remember correctly that at least for 2.x the convention for HABTM join tables was:

1. both parts plural (e.g. ```groups_users```, not ```grouped_users```)
2. both tables actually have to exist (e.g. ```groups``` and ```users```)
3. alphabetically sorted (e.g. ```groups_users``` not ```users_groups```)

Is this correct? Not anymore for 3.x?

Trying to auto detect hasMany assocations with bake on ```suppliers``` and ```managed_suppliers``` should fail on convention 1 and convention 2 (3 would be correct, theoretically, if it was ```manages_suppliers```).

ionas:
Ref: http://book.cakephp.org/3.0/en/intro/conventions.html#model-and-database-conventions
> Join tables, used in BelongsToMany relationships between models, should be named after the model tables they will join, arranged in alphabetical order (apples_zebras rather than zebras_apples).

ionas:
SQL: https://gist.github.com/ionas/94558dc1fd30ed614b3d

ionas:
This is especially troublesome when combined with prefixes.
prefix_products and products will try to setup a HABTM even if you intend to just extend products (hasMany/hasOne) by a plugin for instance.

ionas:
**Question:** Would a PR be welcome that checks if both tables actually exists? E.g. `managed_sets` would be treated as a HABTM in the HasMany detection part but only under the circumstance that `managed` and `sets` both actually exist as tables?

A better solution would be to also check if both sides are actually plural (but possibly harder).

Last but not least why not keep HasMany ALONGSIDE HABTM as HABTM is always also HM/BT <> BT/HM?

ceeram:
[...]

PR #207, opened 2015-12-30
ModelTask HasMany detection improvement

ionas:
Refs: https://github.com/cakephp/bake/issues/204#issuecomment-167158781

Improves ModelTask.php's HasMany association detection by only ignoring tables as part of possible HasAndBelongsToMany join tables by being part of head or tail of a complete join table string and only ignores a possible HasMany association in case both join table parts exist as tables in the database table list.

Also clarifies some variable names.

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -104,11 +104,50 @@ class ModelTask extends BakeTask
      */
     public function bake($name)
     {
-        extract($this->getTableObjectSetup($name)); // $tableObject, $data
+        $table = $this->getTable($name);
+        $tableObject = $this->getTableObject($name, $table);
+        $data = $this->getTableContext($name);
         $this->bakeTable($tableObject, $data);
         $this->bakeEntity($tableObject, $data);
         $this->bakeFixture($tableObject->alias(), $tableObject->table());
         $this->bakeTest($tableObject->alias());
+    }
+    
+    /**
+     * Get table object context for bake
+     *
+     * @param string $name The model name to generate.
+     * @return array
+     */
+    public function getTableContext($name)
+    {
+        $table = $this->getTable($name);
+        $tableObject = $this->getTableObject($name, $table);
+
+        $associations = $this->getAssociations($tableObject);
+        $this->applyAssociations($tableObject, $associations);
+
+        $primaryKey = $this->getPrimaryKey($tableObject);
+        $displayField = $this->getDisplayField($tableObject);
+        $propertySchema = $this->getEntityPropertySchema($tableObject);
+        $fields = $this->getFields();
+        $validation = $this->getValidation($tableObject, $associations);
+        $rulesChecker = $this->getRules($tableObject, $associations);
+        $behaviors = $this->getBehaviors($tableObject);
+        $connection = $this->connection;
+
+        return compact(
+            'associations',
+            'primaryKey',
+            'displayField',
+            'table',
+            'propertySchema',
+            'fields',
+            'validation',
+            'rulesChecker',
+            'behaviors',
+            'connection'
+        );
     }
 
     /**
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -104,37 +104,51 @@ class ModelTask extends BakeTask
      */
     public function bake($name)
     {
+        extract($this->getTableObjectSetup($name)); // $tableObject, $data
+        $this->bakeTable($tableObject, $data);
+        $this->bakeEntity($tableObject, $data);
+        $this->bakeFixture($tableObject->alias(), $tableObject->table());
+        $this->bakeTest($tableObject->alias());
+    }
+
+    /**
+     * Get table object setup for bake
+     *
+     * @param string $name The model name to generate.
+     * @return array
+     */
+    public function getTableObjectSetup($name)
+    {
         $table = $this->getTable($name);
-        $model = $this->getTableObject($name, $table);
+        $tableObject = $this->getTableObject($name, $table);
 
-        $associations = $this->getAssociations($model);
-        $this->applyAssociations($model, $associations);
+        $associations = $this->getAssociations($tableObject);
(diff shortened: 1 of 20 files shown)


lorenzo:
Do you know how to add a test for this?

ionas:
I gave it a try. Fixtures and Results "should" be okay, but within the tests bakeTable() is generating the wrong code.

ionas:
More Info see: https://gist.github.com/ionas/0b9935537e73cfe4f407

lorenzo:
@ionas looks like you need to update the list of fixtures you wan to use here https://github.com/ionas/bake/blob/master/tests/TestCase/Shell/Task/ModelTaskTest.php#L46-L59

markstory:
Looking good so far, but the there are a few other failing tests.

ionas:
Fixtures were using App not Cake NS but also were simply not loaded.
I could resolve those two things and renamed the fixtures to feel more natural too, but I still get errors:
https://gist.github.com/ionas/6daae6d3da868c761296

src/Shell/Task/ModelTask.php
```diff
@@ -104,37 +104,51 @@ public function main($name = null)
...
+        $this->bakeTest($tableObject->alias());
+    }
+
+    /**
+     * Get table object setup for bake
+     *
+     * @param string $name The model name to generate.
+     * @return array
+     */
+    public function getTableObjectSetup($name)
```
ionas: I needed to change/add this to have an easy time callig bakeTable() setup like bake itself does (instead of faking it like the other tests do).

ionas:
Okay, I did finally figure out the core issues I think. Please review.

src/Shell/Task/ModelTask.php
```diff
@@ -104,37 +104,51 @@ public function main($name = null)
      */
     public function bake($name)
     {
+        extract($this->getTableObjectSetup($name)); // $tableObject, $data
```
markstory: Can we use list() instead of extract(). extract() has the downside of making the current function's scope totally opaque.

src/Shell/Task/ModelTask.php
```diff
@@ -104,37 +104,51 @@ public function main($name = null)
...
+                'displayField',
+                'table',
+                'propertySchema',
+                'fields',
+                'validation',
+                'rulesChecker',
+                'behaviors',
+                'connection'
+            ),
+        ];
```
markstory: `list($tableObject, $data) = $this->prepareContext($name)` would be a simpler to read interface than extract() provides.

tests/TestCase/Shell/Task/ModelTaskAssociationsTest.php
```diff
@@ -0,0 +1,179 @@
...
+ */
+class ModelTaskAssociationsTest extends TestCase
+{
+
+    /**
+     * Fixtures should be dropped after each tests
+     *
+     * @var bool
+     */
+    public $dropTables = true;
```
markstory: You shouldn't need to drop tables unless you are altering schema.

markstory:
Just a [PHPCS error](https://travis-ci.org/cakephp/bake/jobs/99732314) now.

src/Shell/Task/ModelTask.php
ionas: Urgh. Both is ugly. I switched to list() now, but that relies on order of parameters (order in the array).

ionas:
Okay, I hope refactoring was "Im Sinne des Erfinders" - e.g. what should have been done with the test framework. I hope the new context method name is fine for you @markstory. Sorry for the many fixes and updates.

src/Shell/Task/ModelTask.php
markstory: Yes, but 'tuple unpacking' is much simpler to follow than extract().

src/Shell/Task/ModelTask.php
```diff
@@ -105,21 +105,35 @@ public function main($name = null)
...
 
-        $associations = $this->getAssociations($model);
-        $this->applyAssociations($model, $associations);
+    /**
+     * Get table context for baking a given table.
+     *
+     * @param string $name The model name to generate.
+     * @return array
+     */
+    public function getTableContext($tableObject, $table, $name)
```
markstory: This works too :+1:

markstory:
Only [PHPCS issues left](https://travis-ci.org/cakephp/bake/jobs/99829833#L306).

ionas:
@markstory the CS error is within the bake result comparison files.
```
<?php
namespace App\Model\Table;

use App\Model\Entity\OldProduct;
use Cake\ORM\Query;
use Cake\ORM\RulesChecker;
use Cake\ORM\Table;
use Cake\Validation\Validator;

/**
 * OldProducts Model
 *
 */
class OldProductsTable extends Table
{

    /**
     * Initialize method
     *
     * @param array $config The configuration for the Table.
     * @return void
     */
    public function initialize(array $config)
    {
        parent::initialize($config);

        $this->table('old_products');
        $this->displayField('name');
        $this->primaryKey('id');

        $this->addBehavior('Timestamp');

    }

    /**
     * Default validation rules.
     *
     * @param \Cake\Validation\Validator $validator Validator instance.
     * @return \Cake\Validation\Validator
     */
    public function validationDefault(Validator $validator)
    {
        $validator
            ->add('id', 'valid', ['rule' => 'numeric'])
            ->allowEmpty('id', 'create');

        $validator
            ->requirePresence('name', 'create')
            ->notEmpty('name');

        return $validator;
    }
}
```
It is the empty line before the closing brace. Possibly because no associations were detected but the bake template assumes some are detected and always adds a new line there.

If you deem this fix important I will have to fix the bake templates themselves, then fix the comparison result. I did already kill my test setup for the comparison files. Will need to re-setup. The core issue lies with the templates how they are at the current state.

I will take care about that ASAP.

markstory:
We should probably update the templates so they output PSR2 compliant code.

ionas:
That should be fixed.

Anything else?

PR #209, opened 2016-01-03
Bake 1.2

lorenzo:
This is to be merged right before cake 3.2 is marked as stable

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.4",
-        "cakephp/cakephp": ">=3.1.0"
+        "cakephp/cakephp": ">=3.2.0"
     },
     "require-dev": {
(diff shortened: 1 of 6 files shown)


markstory:
Looks good. I updated the travis config.

josegonzalez:
The comparison files still need to be updated.

dereuromark:
Looks like this should be merged now for the 3.2 version tagged.

markstory:
@dereuromark I'll take care of that.

composer.json
```diff
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.4",
```
ADmad: This should be bumped to 5.5 isn't it?

markstory: I can do that.

ADmad: Made a PR fixing this and few other things

PR #210, opened 2016-01-07
Updated docblock returns to pass CakePHP coding standard

davidyell:
Currently the bake doc block doesn't pass the cakephp standard, as redirect returns a response

### DIFF
--- a/src/Template/Bake/Element/Controller/add.ctp
+++ b/src/Template/Bake/Element/Controller/add.ctp
@@ -18,7 +18,7 @@ $compact = ["'" . $singularName . "'"];
     /**
      * Add method
      *
-     * @return void Redirects on successful add, renders view otherwise.
+     * @return \Cake\Network\Response|void Redirects on successful add, renders view otherwise.
      */
     public function add()
     {
--- a/src/Template/Bake/Element/Controller/edit.ctp
+++ b/src/Template/Bake/Element/Controller/edit.ctp
@@ -22,7 +22,7 @@ $compact = ["'" . $singularName . "'"];
      * Edit method
      *
      * @param string|null $id <%= $singularHumanName %> id.
-     * @return void Redirects on successful edit, renders view otherwise.
+     * @return \Cake\Network\Response|void Redirects on successful edit, renders view otherwise.
      * @throws \Cake\Network\Exception\NotFoundException When record not found.
      */
     public function edit($id = null)
--- a/src/Template/Bake/Element/Controller/login.ctp
+++ b/src/Template/Bake/Element/Controller/login.ctp
@@ -17,7 +17,7 @@
     /**
      * Login method
      *
-     * @return void
+     * @return \Cake\Network\Response|void
      */
     public function login()
     {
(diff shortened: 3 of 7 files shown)


src/Template/Bake/Element/Controller/add.ctp
```diff
@@ -18,7 +18,7 @@ $compact = ["'" . $singularName . "'"];
     /**
      * Add method
      *
-     * @return void Redirects on successful add, renders view otherwise.
+     * @return \Cake\Network\Response|void Redirects on successful add, renders view otherwise.
```
dereuromark: void should be`|null` to be completely correct. But I guess this nitpick is not too important.

davidyell:
Ah, nothing like a bit of force push to fix things :tada:

src/Template/Bake/Element/Controller/edit.ctp
```diff
@@ -22,7 +22,7 @@ $compact = ["'" . $singularName . "'"];
      * Edit method
      *
      * @param string|null $id <%= $singularHumanName %> id.
-     * @return void Redirects on successful edit, renders view otherwise.
+     * @return \Cake\Network\Response|void Redirects on successful edit, renders view otherwise.
```
markstory: The expected results will need to be updated as well.

davidyell: Tests! Always forget something. Thanks for the tip, will sort that.

src/Template/Bake/Element/Controller/add.ctp
davidyell: Why would that be? I don't know what might return `null` in an Add action? I also wondered why `delete()` has `\Cake\Network\Response|null` as the baked action always returns a redirect.

dereuromark: Because void means that this method is not to be expected to return anything.
As soon as you sometimes return response that means you MUST check on the response.
Thus there is no logical point in adding |void. It must be |null now as the check reveals either null or not. never void. See also the other core functions doing it that way.

davidyell: Shall I update the PR with these changes then?

dereuromark: Would be nice as it was be more correct.

dereuromark: It's all semantics :) Semantically one would also need to always add return null explicitly to make the IDE happy once you start returning sth. But we at this point don't care about that too much.
I wonder if actions should be more like Symfony ones and always return an array otherwise with the view vars :) instead of setting them. But thats off topic now HEHE

davidyell: Sounds like a different PR to this one. I think I'll keep it as it is, so there isn't any scope creep in this PR.

dereuromark: |null is not a scope screep :) Its consistent doc block type declaration.

davidyell: The scope of this PR, not of the codebase.

dereuromark: "Currently the bake doc block doesn't pass the cakephp standard, as redirect returns a response" sounds to me like this would fit perfectly into the scope.

davidyell:
Thanks @markstory

PR #213, opened 2016-01-08
Allow the hidden option to work when baking entities

davidyell:
Fixed baking entities with non whitelisted hidden fields, and added a test case.

`$ bin/cake bake model -f --no-table --hidden _matchingData,foo --no-test --no-fixture Examples`

Would not create a `$_hidden` field in the entity.


markstory:
Thanks :+1:

PR #216, opened 2016-01-19
Make baked tests more code-completion/inspection friendly.

ndm2:
This PR aims to make baked tests a little more IDE friendly by adding real properties to the baked tests.

Usually every time I bake tests, the first thing I do is adding real properties for the dynamic ones set in the generated constructor, so that I get proper code-complection, and my IDE doesn't light up like a  :christmas_tree: complaining about non-existing properties :)

### DIFF
--- a/src/Shell/Task/TestTask.php
+++ b/src/Shell/Task/TestTask.php
@@ -535,7 +535,7 @@ class TestTask extends BakeTask
      * - `description` (the property description)
      * - `type` (the property docblock type)
      * - `name` (the property name)
-     * - `value` (the properties initial value)
+     * - `value` (optional - the properties initial value)
      *
      * @param string $type The Type of object you are generating tests for eg. controller
      * @param string $subject The name of the test subject.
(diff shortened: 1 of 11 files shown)


tests/comparisons/Test/testBakeBehaviorTest.php
```diff
@@ -11,6 +11,13 @@ class ExampleBehaviorTest extends TestCase
 {
 
     /**
+     * Test subject
+     *
+     * @var \App\Model\Behavior\ExampleBehavior
+     */
+    public $Example = null;
```
dereuromark: Is the = null necessary? This implies that its nullable, and the doc block would need |null attached.
It might be cleaner to omit it.

lorenzo:
This looks good to me :+1:

tests/comparisons/Test/testBakeBehaviorTest.php
ndm2: @dereuromark You have a point there. Initialising "empty" properties with `null` is a personal habit of mine, I can remove it if it's considered "wrong" or confusing.

lorenzo: Good observation, I think it is better to remove it

ndm2: ...removed

dereuromark:
Nice!

src/Shell/Task/TestTask.php
```diff
@@ -246,11 +248,12 @@ public function bake($type, $className)
         $this->BakeTemplate->set(compact(
             'subject',
             'className',
+            'properties',
             'methods',
             'type',
             'fullClassName',
             'mock',
-            'realType',
```
markstory: Was this not defined before? Removing variables could break userland templates.

ndm2: @markstory  `$realType` was defiend quite some time ago, but it has been removed in adb7f6025f4dae735c88f576a1b2806b5911a2f3 Removing the key shouldn't change the current behavior.

lorenzo:
Thanks @ndm2

PR #217, opened 2016-01-24
remove empty file when baking plugin

chris48s:
Very minor issue, but baking a new plugin should remove `empty` file in plugins directory if it exists.


ionas:
If you put each plugin into its own repository, the empty file will still enforce the dir in the parent git repo, wouldn't it?

markstory:
Seems safe to me. The file will only be deleted if it exists.

PR #220, opened 2016-01-31
Save to own variable first

dereuromark:
Makes debugging it easier.
Like debug($....);die(); etc.


josegonzalez:
Can you also update the bake templates?

dereuromark:
Should be all good now.

PR #223, opened 2016-02-01
No need to duplicate License part.

dereuromark:
License is in root file LICENSE.txt already.

Or we can remove this block alltogether as the icon above states it as well as the uppercase filename in the repo as well quite clearly.


josegonzalez:
Personally, I'd remove it.

dereuromark:
Done.

PR #224, opened 2016-02-01
update travis config and badges

ADmad:


composer.json
```diff
@@ -18,7 +18,7 @@
         "source": "https://github.com/cakephp/bake"
     },
     "require": {
-        "php": ">=5.4",
+        "php": ">=5.5.9",
```
markstory: I think I conflicted you with a :cow: boy push to master.

ADmad: No worries, with the experience I gained in conflict resolution doing few 2.x => 3 merges in the past fixing these one liners is a piece of cake :smile:

Issue #219, opened 2016-01-29
Neither view template nor form element (add, edit) are handling hasOne relations


xavier83ar:
It seems that the view template it trying to show them the same way as belongsTo relation, which is ok, but it ignores it when filtering with map function. BelongsTo and hasOne relations are mixed together into $immediateAssociations array, then this array is filtered checking if the [foreingKey exists as a field](https://github.com/cakephp/bake/blob/master/src/Template/Bake/Template/view.ctp#L22), which is true for belongsTo associations but not for hasOne: 

```php
$immediateAssociations = $associations['BelongsTo'] + $associations['HasOne'];
// ...
$associationFields = collection($fields)
    ->map(function($field) use ($immediateAssociations) {
        foreach ($immediateAssociations as $alias => $details) {
            if ($field === $details['foreignKey']) {
                return [$field => $details];
            }
        }
    })
    ->filter()
    ->reduce(function($fields, $value) {
        return $fields + $value;
    }, []);
```

As for the form template, used on add and edit, there is no clue of hasOne relations. Is it on purpose?

In the meantime I'm working to see if I can fix view template.

markstory:
I don't think its on _purpose_. Most likely its an oversight that was introduced during a refactoring. If you're fixing the templates for your application, a pull request would be appreciated :smile:

xavier83ar:
Ok, working on it. It's my first time doing it, so I'm reading contributing guidelines and such things.

xavier83ar:
Well, I think I did understand the bake's tests structure. As far as I see there is no test case for hasOne relations. 

At first I got confused a bit because on the folder for the espected baked templates results there is a testBakeIndex.ctp file, a testBakeEdit.ctp and a testBakeView.ctp, but the last one is never used, also this file is outdated after the new templates introduced by @midorikocak, even I've deleted this file and all tests are still passed.
Instead I've found a testBakeTemplate.ctp file which is used by testBakeTemplate method to test view.ctp results. 

I think that this should be changed, in my opinion using testBakeView as file name and method name is far more representative of what this test case actually does. However the presence of this file (testBakeView) outdated makes me think that this was like this on the past and gets changed to testBakeTemplate, I can't guess why.  What do you think?

In the meantime I keep working on add some test case for hasOne relations and fixing the issue.

xavier83ar:
[...]

PR #225, opened 2016-02-01
HasOne relations

xavier83ar:
#219 
View template now shows hasOne relations as expected. I've modified tests cases to cover this. This also includes some other minor additions and fixes to phpdocs. And finally I've fixed some other tests cases (this was because when I merge master into my branch, some tests start to failing, after the validator refactoring, I've updated, I don't know if that was right, but I'm pretty new to this).


xavier83ar:
I don't know why commits were duplicated. I guess I've to learn more about git, a some point I screw it up. Sorry for that.

tests/Fixture/BakeTemplateRolesFixture.php
```diff
@@ -0,0 +1,43 @@
+<?php
+/**
+ * Created by javier
+ * Date: 31/01/16
+ * Time: 10:59
+ */
```
markstory: These should be the standard file header used across the project. I can take care of that.

xavier83ar: Oh, I missed that one. It's the default header added by the IDE, besides that I've replaced on other new files, I didn't know what to put in @since, so if you can take care of that too, go ahead.

Issue #184, opened 2015-11-23
<%- crashes my custom cake bake templates


ionas:
```<%-``` will crash... unless you add a space behind like ```<%- // space behind``` instead of only a new line directly following.

Why?
see for example: https://gist.github.com/ionas/946150e83780f3fa2127#file-cakephp3-bake-add-ctp-L43

**Comment by ionas** [comment]
```-%>``` always seems to crash, even if there is a space in front of it.
http://book.cakephp.org/3.0/en/bake/development.html#bake-template-syntax

**Comment by josegonzalez** [comment]
Can you gist any details about the crash?

**Comment by ionas** [comment]
Does this help you?
https://gist.github.com/ionas/3c295efb336eaa6e1dee#file-gistfile1-php-L44

**Comment by josegonzalez** [comment]
Do you maybe have alternate short tags enabled?

**Comment by josegonzalez** [comment]
We probably could handle that by adding the following regex [here](https://github.com/cakephp/bake/blob/master/src/View/BakeView.php#L56):

```php
'/\n[ \t]+<%-\n/' => "\n<%\n",
```

ionas:
@josegonzalez I don't think I did back then (standard puphpet vagrant box with recent php 5.6.x latest stable)

ionas:
@josegonzalez I don't think I did back then (standard puphpet vagrant box with recent php 5.6.x latest stable)

ionas:
@josegonzalez I don't think I did back then (standard puphpet vagrant box with recent php 5.6.x latest stable)

dereuromark:
Would that be solved with https://github.com/cakephp/bake/issues/185#issuecomment-176844479 ?

ionas:
It would be *resolved* by utilizing another implementation. While twig really appeals to me as a *end user templating language* because you can drive it by opt-in configuration, adding another layer of complexity to something that is already advanced cakephp development... I am not so sure about that.

dereuromark:
The layer already exists, its called *bake template*. and it uses a custom engine. I simply say use a standard one.

josegonzalez:
@ionas do you have a PR to fix the regular expressions used for transforming the templates?

AD7six:
> It would be resolved by utilizing an entirely different implementation.

[...]

PR #226, opened 2016-02-03
fixes for <%- and -%>

AD7six:
Make -%> match erb template behavior
Allow <%- to work multiline

Fixes https://github.com/cakephp/bake/issues/184


ionas:
Awesome.
Thank you so move!

PR #227, opened 2016-02-18
Added --no-fixtures option to test baking.

makallio85:
I have controller, that allows post or get requests only. I was not able to bypass controller instantiating https://github.com/cakephp/bake/blob/master/src/Shell/Task/TestTask.php#L307. 

Solution was to add --no-fixtures parameter.

### DIFF
--- a/src/Shell/Task/TestTask.php
+++ b/src/Shell/Task/TestTask.php
@@ -54,6 +54,7 @@ class TestTask extends BakeTask
         'Behavior' => 'Model\Behavior',
         'Helper' => 'View\Helper',
         'Shell' => 'Shell',
+        'Task' => 'Shell\Task',
         'Shell_helper' => 'Shell\Helper',
         'Cell' => 'View\Cell',
         'Form' => 'Form',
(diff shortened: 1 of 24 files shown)


makallio85:
Is there something I could do for this pr?

src/Shell/Task/TestTask.php
```diff
@@ -212,15 +212,17 @@ public function bake($type, $className)
...
-        if (!empty($this->params['fixtures'])) {
-            $fixtures = array_map('trim', explode(',', $this->params['fixtures']));
-            $this->_fixtures = array_filter($fixtures);
-        } elseif ($this->typeCanDetectFixtures($type) && class_exists($fullClassName)) {
-            $this->out('Bake is detecting possible fixtures...');
-            $testSubject = $this->buildTestSubject($type, $fullClassName);
-            $this->generateFixtureList($testSubject);
-        }
+		
+		if (empty($this->params['no-fixtures'])) {
```
antograssiot: indentation seems off here, please use spaces.

antograssiot:
@makallio85  The flag to avoid to bake fixture when baking a model is ``no-fixture`` instead of ``no-fixtures``.
We should be using the same name, it would at the same time solve the issue where when using something like ``bin/cake bake model Users --no-fixture`` will not bake fixtures files but will include the $fixture array in the ``UsersTableTest.php``.

Finally, it would be awesome if you could rebase your changes on top of master so the test suite will work properly

makallio85:
Ok, Ill fix this next week.

PR #229, opened 2016-03-01
Add a basic OptionParser into the baked shell, for ease of customisation

davidyell:
I always have to look this up, so figured it'd sure be handy to have it right there in the baked shell with a link to the docs.

I have run the test suite and everything passed. Let me know if you'd like me to make an attempt at including a new test case.

### DIFF
--- a/src/Template/Bake/Shell/shell.ctp
+++ b/src/Template/Bake/Shell/shell.ctp
@@ -25,11 +25,26 @@ class <%= $name %>Shell extends Shell
 {
 
     /**
+     * Manage the available sub-commands along with their arguments and help
+     *
+     * @see http://book.cakephp.org/3.0/en/console-and-shells.html#configuring-options-and-generating-help
+     *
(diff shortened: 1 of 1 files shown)


markstory:
I like the change, but the tests will need to be updated.

davidyell:
The tests are not failing because of my pull request. They are failing for other reasons. @markstory please can you clarify if you want me to add a new test case or not. Fixing the failing Travis is outside the scope of this PR.

The test suite passes locally
Terminal

chinpei215:
I could reproduce the test failing on **master**. So it is not related with this PR.
bake-error

markstory:
@chinpei215 Interesting, the tests passed for me locally using sqlite. I'll take a look tonight at home with a postgres db.

markstory:
Postgres tests should be fixed with #230

davidyell:
Thanks @markstory :+1: :tada:

PR #230, opened 2016-03-03
Fix tests failing on postgres.

markstory:
Treat timestamp columns like datetime columns for postgres.


chinpei215:
I've fixed the error on AppVeyor in #231. I forgot to create the PR yesterday :(

PR #245, opened 2016-05-02
Adds @method and @mixin annotations to Table plus other fixes

### LINKED ISSUE #237: Add @method stubs for tables to declare return type.
When you generate a `Table` with bake. You get a class phpDoc like this.

For example, `PostsTable` with some associations.

```
/**
 * Posts Model
 *
 * @property \Cake\ORM\Association\HasMany $PostLikes
 * @property \Cake\ORM\Association\HasMany $PostViews
 * @property \Cake\ORM\Association\BelongsToMany $Boards
 */
```

Afterwards when you want to use `get` or `newEntity` you have to do this to get proper type auto-complete.

```
    /**
     * @type \App\Model\Entity\Post $post
     */
    $post = $this->Posts->get($id);
```

It peppers the source code with type declares and is a pain to always write. So what I do is declare the return types using `@method`. Like this

```
/**
 * Posts Model
 *
 * @property \Cake\ORM\Association\HasMany $PostLikes
 * @property \Cake\ORM\Association\HasMany $PostViews
 * @property \Cake\ORM\Association\BelongsToMany $Boards
 *
 * @method get($primaryKey, $options = []) : Post
 * @method newEntity($data = null, array $options = []) : Post
 * @method patchEntity(EntityInterface $entity, array $data, array $options = []) : Post
 */
```

Now auto-complete works in the IDE

```
     $this->Posts->get($id)->title;   <-- title marked as property from Post entity.
```

There are other methods from the table that could use this trick as well. Like "save", "newEntities", etc. Any method where the return type is the `EntityInterface`

Would this be something worth adding to the bake plugin?

### LINKED ISSUE #238: Add @mixin for tables that use behaviors.
To expand on #237 that adds return stubs. This improvement would add mixins for behaviors.

If bake adds the `TimestampBehavior` it should also add the `@mixin` for that behavior to the class.

```
/**
 * Posts Model
 *
 * @property \Cake\ORM\Association\HasMany $PostLikes
 * @property \Cake\ORM\Association\HasMany $PostViews
 * @property \Cake\ORM\Association\BelongsToMany $Boards
 *
 * @mixin \Cake\ORM\Behavior\TimestampBehavior
 */
```

This would give you auto-complete on all the public functions from that behavior on the table.

```
    $this->Posts->touch($post); // <- IDE sees touch as valid method of Posts
```

Without the `@mixin` the IDE will mark `touch` as invalid method for that class.

thinkingmedia:
This PR closes #237 and closes #238 

Updates the baking for Table

- @method stubs that declare return type are added to Table classes, ~~but only for PHP7~~
- @mixin are added to Table classes for behaviors, ~~but only for PHP7~~

Updates the baking for Entity

- @property for table columns are listed first
- @property for associations are listed last
- trailing comma on $_accessible fixed.

General updates

- Both Table and Entity now use DocBlock helper to write class description.
- ~~for tests that differ on PHP7 you can just add a "7" to the end of the comparison file.~~
- added declaration of properties for unit tests

~~This is the first time that Bake will render different results based upon the PHP version. If you are using PHP7 the @method stubs for Table will be added, otherwise everything works as it did before.~~

~~There isn't a way to configure Bake per project. I would like to open a issue to add config support and then make PHP7 behavior configurable.~~


thinkingmedia:
Forgot to mention that PHP7 and PHP5 differences are part of the travis tests.

src/Shell/Task/ModelTask.php
```diff
@@ -173,7 +173,7 @@ public function all()
     public function getTableObject($className, $table)
     {
         if (TableRegistry::exists($className)) {
-            return TableRegistry::get($className);
+                return TableRegistry::get($className);
```
markstory: Indentation looks wrong.

src/Template/Bake/Model/entity.ctp
```diff
@@ -63,11 +55,7 @@ class <%= $name %> extends Entity
      *
      * @var array
      */
-    protected $_accessible = [
-<% foreach ($accessible as $field => $value): %>
-        '<%= $field %>' => <%= $value %>,
-<% endforeach; %>
-    ];
+    protected $_accessible = [<%= $this->Bake->stringifyList($accessible,['quotes'=>false]) %>];
```
markstory: You need spaces around the `,` and `=>`.

src/Template/Bake/Model/table.ctp
```diff
@@ -13,6 +13,29 @@
...
+if(PHP_MAJOR_VERSION >= 7)
+{
+    $annotations[] = "@method get(\$primaryKey, \$options = []) : {$entity}";
+    $annotations[] = "@method newEntity(\$data = null, array \$options = []) : {$entity}";
+    $annotations[] = "@method newEntities(array \$data, array \$options = []) : {$entity}[]";
+    $annotations[] = "@method save(EntityInterface \$entity, \$options = []) : {$entity}";
+    $annotations[] = "@method patchEntity(EntityInterface \$entity, array \$data, array \$options = []) : {$entity}";
+    $annotations[] = "@method patchEntities(\$entities, array \$data, array \$options = []) : {$entity}[]";
+    foreach ($behaviors as $behavior => $behaviorData)
+    {
```
markstory: Should be `foreach (...) {`.

src/Template/Bake/Model/table.ctp
```diff
@@ -25,22 +48,16 @@ $uses = [
     'use Cake\ORM\Table;',
     'use Cake\Validation\Validator;'
 ];
+if(PHP_MAJOR_VERSION >= 7)
+{
```
markstory: Should be `if (...) {`

src/View/Helper/DocBlockHelper.php
```diff
@@ -1,14 +1,52 @@
...
+                }
+                $previous = $type;
+            }
+            $lines[] = $ann;
+        }
+
+        $lines = array_merge(["/**"], (new Collection($lines))->map(function ($line) {
+            return rtrim(" * {$line}");
+        })->toArray(), [" */"]);
+        return implode(PHP_EOL, $lines);
```
markstory: This will use `\r\n` on windows. Is that what windows users expect in their source code?

tests/TestCase/Php7CompareTrait.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    use StringCompareTrait {
+        assertSameAsFile as assertSameAsFileBase;
+    }
+
+    /**
+     * Compare the result to the contents of the file based upon which
+     * version PHP is being used.
+     *
+     * The env variable UPDATE_TEST_COMPARISON_FILES will not generate "php7"
```
markstory: Why are php7 files different?

tests/TestCase/Php7CompareTrait.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public function assertSameAsFile($path, $result)
+    {
+        if (!file_exists($path)) {
+            $path = $this->_compareBasePath . $path;
+        }
+
+        if (PHP_MAJOR_VERSION >= 7 && file_exists($path . '7')) {
+            $path = $path . '7';
+        }
```
markstory: Why?

tests/TestCase/Php7CompareTrait.php
markstory: Are `@method` documentation tags somehow invalid in PHP5?

src/Template/Bake/Model/table.ctp
josegonzalez: Why is this scoped to PHP7? Do they not work under PHP 5.x?

thinkingmedia: @josegonzalez type hinting the return value is PHP7 only.

tests/TestCase/Php7CompareTrait.php
thinkingmedia: @markstory The @method tags use type hinting on the return value which is PHP7 only.

thinkingmedia:
I might be wrong about how I did the return types.

Looks like you can write it the PHP7 way, but you can also write it the PHP5 way.

```
@method Entity get(\$primaryKey, \$options = [])";
```

So I will be updating this PR. If it is not just a PHP7 feature, than I'll remove the PHP7 specific changes.

src/Template/Bake/Model/table.ctp
josegonzalez: I do not understand what that has to do with annotations though, which should work fine on older versions.

thinkingmedia: @josegonzalez On PHP7 its written like this `@method foo() : array` and it doesn't work in PhpStorm is the version is set to PHP 5. But I think if I write like this `@method array foo()` it works in all versions. I'll fix it tomorrow.

src/View/Helper/DocBlockHelper.php
thinkingmedia: @markstory I thought using `PHP_EOL` was standard practice? I can change it to `"\n"` if you want.

src/Template/Bake/Model/table.ctp
josegonzalez: I mean, if by "doesn't" work, you mean "doesn't autocomplete", then that's fine, no?

src/View/Helper/DocBlockHelper.php
markstory: I have no idea. I've not worked in windows environments in a long time.

thinkingmedia:
I have some improvements that can be made to this PR. I'll work on it today.

src/View/Helper/DocBlockHelper.php
```diff
@@ -51,23 +89,49 @@ public function buildEntityPropertyHintTypeMap(array $propertySchema)
...
+     * @return array The property DocType map.
+     */
+    public function buildEntityAssociationHintTypeMap(array $propertySchema)
+    {
+        $properties = [];
+        foreach ($propertySchema as $property => $info) {
+            if ($info['kind'] == 'association') {
+                $type = $this->associatedEntityTypeToHintType($info['type'], $info['association']);
+                if ($info['association']->type() === Association::MANY_TO_ONE) {
+                    $properties = $this->_insertAfter(
```
ndm2: If you split the properties, then doing this isn't necessary anymore, as it is used to position the association properties after the corresponding foreign key properties (`kind = column`).

src/Shell/Task/ModelTask.php
thinkingmedia: it wasn't picked up by phpcs

src/Template/Bake/Model/table.ctp
```diff
@@ -13,34 +13,42 @@
...
 use Cake\Utility\Inflector;
+
+$annotations = [];
+foreach ($associations as $type => $assocs) {
+    foreach ($assocs as $assoc) {
+        $typeStr = Inflector::camelize($type);
+        $annotations[] = "@property \Cake\ORM\Association\\{$typeStr} \${$assoc['alias']}";
+    }
+}
+$annotations[] = "@method {$entity} get(\$primaryKey, \$options = [])";
```
ndm2: Should the return types maybe be FQNs? Same for `EntityInterface` in the arguments.

src/Template/Bake/Model/table.ctp
```diff
@@ -13,34 +13,42 @@
...
+$annotations[] = "@method {$entity} newEntity(\$data = null, array \$options = [])";
+$annotations[] = "@method {$entity}[] newEntities(array \$data, array \$options = [])";
+$annotations[] = "@method {$entity} save(EntityInterface \$entity, \$options = [])";
```
ndm2: I think this should include `bool`, ie {$entity}`|bool`.

thinkingmedia: I'll double check the return types. thanks.

src/Template/Bake/Model/table.ctp
thinkingmedia: I have no idea on that. I was adding a `use` for those, but if FQN is the preferred way then I'm fine with that.

ndm2: The general convention for docblock typehints is to use FQNs, I think it should apply to `@method` too.

thinkingmedia:
I get over 300 warnings from PHPUnit that `getMock` is deprecated. Should I fix this in this PR or ignore it?

ADmad:
> Should I fix this in this PR or ignore it?

A separate PR to fix those would be nice.

tests/comparisons/Model/testBakeAssociationDetectionOldProductsTable.php
```diff
@@ -14,7 +14,7 @@
  * @method OldProduct get($primaryKey, $options = [])
  * @method OldProduct newEntity($data = null, array $options = [])
  * @method OldProduct[] newEntities(array $data, array $options = [])
- * @method OldProduct save(EntityInterface $entity, $options = [])
+ * @method OldProduct| save(EntityInterface $entity, $options = [])
```
ADmad: This doesn't look right.

src/Template/Bake/Model/table.ctp
```diff
@@ -13,13 +13,30 @@
...
+foreach ($associations as $type => $assocs) {
+    foreach ($assocs as $assoc) {
+        $typeStr = Inflector::camelize($type);
+        $annotations[] = "@property \Cake\ORM\Association\\{$typeStr} \${$assoc['alias']}";
+    }
+}
+$annotations[] = "@method {$namespace}\\Model\\Entity\\{$entity} get(\$primaryKey, \$options = [])";
+$annotations[] = "@method {$namespace}\\Model\\Entity\\{$entity} newEntity(\$data = null, array \$options = [])";
+$annotations[] = "@method {$namespace}\\Model\\Entity\\{$entity}[] newEntities(array \$data, array \$options = [])";
+$annotations[] = "@method {$namespace}\\Model\\Entity\\{$entity}|bool save(Cake\\Datasource\\EntityInterface \$entity, \$options = [])";
```
ndm2: FQNs begin with a leading namespace separator ;)

thinkingmedia: Thanks, I'm going to run the plugin on one of my projects and load up the files in the IDE and make sure it all works as expected.

Issue #239, opened 2016-04-08
Be able to rewrite the validationDefault method on a table class.


thinkingmedia:
I don't think there are any features in bake that allow you to modify an existing class, but if it could. I would like to be able to regenerate the `validationDefault` method on a table.

It shouldn't be technically difficult to rewrite just one method.

josegonzalez:
It actually would because you'd need to parse the entire class into tokens and then replace *just* your tokens without modifying the rest of the code.

Related: https://github.com/cakephp/bake/issues/205

markstory:
Bake has never done inline edits. Its a really hard problem that I'd prefer to not try and solve.

thinkingmedia:
@markstory I accept your challenge and will raise you one pull request.

markstory:
Closing as editing existing files is well outside of bake's scope for the present. Programmatically editing user modified code requires more complex methods like AST transforms, and I'm not sure we want to take that scope on in bake.

PR #246, opened 2016-05-02
Adds the ability to write bake output to clipboard.

thinkingmedia:
This PR adds a new feature to bake. 

I wanted a way to update existing files with changes from the database, but didn't want to overwrite my files, and it had to work on the different OS platforms that I use. So this PR allows you to write all bake changes to the clipboard instead of files.

- Adds a new `--clipboard` or `-c` option for all bake commands.
- Redirects file writes to a temp file.
- After baking is finished the temp file is sent to the clipboard.
- Works on Windows, Linux and Mac.

I have not tested this with all bake commands. It works by overriding the `createFile` method in `Shell` from `BakeShell`. This should cover all commands.

### DIFF
--- a/src/Shell/RedirectClipboardTrait.php
+++ b/src/Shell/RedirectClipboardTrait.php
@@ -1,0 +1,162 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         1.1.1
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Bake\Shell;
+
+use Cake\Filesystem\File;
+
+/**
+ * Redirects the file output of a Shell to the clipboard. All files are copied to the clipboard
+ * in the order they are written.
+ *
+ * @see \Cake\Console\Shell This traits works only on Shell classes.
+ */
+trait RedirectClipboardTrait
+{
+
+    /**
+     * @var null|string Name of the temp file used for clipboard output.
+     */
+    private static $_clipboardTemp = null;
+
+    /**
+     * @var null|File File handle used for clipboard output.
+     */
+    private static $_clipboardFile = null;
+
+    /**
+     * Opens the temporary file that receives clipboard contents.
+     *
+     * @return bool
+     */
+    private static function openClipboardFile()
+    {
+        if (self::$_clipboardFile !== null) {
+            return true;
+        }
+        self::$_clipboardTemp = tempnam(sys_get_temp_dir(), "tmp");
+        self::$_clipboardFile = new File(self::$_clipboardTemp, true);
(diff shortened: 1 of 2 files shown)


thinkingmedia:
Closes #239

markstory:
Dumb question, but if we supported an `--output` argument, couldn't people do things like `bin/cake bake model --output | <clipboard>` ? This would also allow people to redirect the output to other CLI based tools.

src/Shell/Task/BakeTask.php
```diff
@@ -237,6 +239,10 @@ public function getOptionParser()
             'short' => 't',
             'help' => 'The theme to use when baking code.',
             'choices' => $bakeThemes
+        ])->addOption('clipboard', [
+            'short' => 'c',
```
markstory: You can't use `-c`. Its used for `connection`.

jrbasso:
I agree with @markstory . It is usually called as dry run.

jhli:
Generally a good idea bit -- output is what we want. Then one can pass things on the cli.

Wonder if diffing on *nix could be added to make things convenient, e.g bin/cake model foo --output x --output-diff | clipboard

thinkingmedia:
How many people knew they could send output to the clipboard on the CLI or even know how?

I think an output to stdio is a great feature but it's not going to be easy to add. All logging messages have to be silenced, and I don't think people will want a stream of multiple files.

This PR works as is. Those that need to regenerate files that already exist but don't want to over write would benefit from this.

src/Shell/Task/BakeTask.php
thinkingmedia: Can you recommend an alternate letter?

dereuromark: Maybe it doesn't need one?

saeideng:
you can use git , and review history of changes
OR
we need move exist original file to other directory  like `tempbake`
for example of a Entity
`src/Model/Entity/Tag.php` move to `tempbake/{version_code}/Model/Entity/Tag.php`

OR, move new files to `tempbake`

thinkingmedia:
What's the plan with this PR? Is it going to get merged or you want something more generic?

I think it's useful in it's current form, because it doesn't require the developer to do anything extra.

src/Shell/RedirectClipboardTrait.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Optionally redirects file output to the clipboard.
+     *
+     * @param string $path Where to put the file.
+     * @param string $contents Content to put in the file.
+     * @return bool Success
+     */
+    public function createFile($path, $contents)
```
markstory: Why would you write multiple files to the clipboard?

thinkingmedia: There's no option right now in Bake to pick which of the set of files it bakes. So if the bake operation generates 3 files. I have to send them all to the clipboard.

thinkingmedia:
We don't have to accept this PR. I just need to know what I should do on it, or if we should close it.

inoas:
@thinkingmedia what about making bake return to the stdout. That way you can catch it with `pbcopy` or any other unixy tool, easily? Redirect it basically where you want it to go to?

thinkingmedia:
@inoas I think that's a good idea, but silencing the bake/console output so that doesn't get mixed in would be difficult.

saeideng:
What you thing about write in tmp folder?

thinkingmedia:
@saeideng that's a good idea. Or define a new target directory in the params.

inoas:
@thinkingmedia `--silence` and `--stdout` ? both together would return the contents to the stdout?

thinkingmedia:
@inoas 🤓 I stand corrected.

Issue #243, opened 2016-04-30
Bake template don't use the correct --connection parameter


alysson-azevedo:
I have two datasources, the default is used on my virtual machine, it connects to postgres using unix soquet and the 'cli' connection i use to run commands from my local terminal. Both datasources connects to the same database.
```php
        'default' => [
            'className' => 'Cake\Database\Connection',
            'driver' => 'Cake\Database\Driver\Postgres',
            'persistent' => false,
            'unix_socket' => true,
            'host' => '',
            'username' => 'vagrant',
            'database' => 'my_db',
            'quoteIdentifiers' => false
        ],
        'cli' => [
            'className' => 'Cake\Database\Connection',
            'driver' => 'Cake\Database\Driver\Postgres',
            'persistent' => false,
            'unix_socket' => false,
            'host' => 'dev1.local',
            'username' => 'vagrant',
            'database' => 'my_db',
            'quoteIdentifiers' => false
        ],
```

This works. If i run from my local terminal:
``bin/cake bake template -c cli -p Fluid --prefix App -v``  
Bake returns:  

    Possible tables to bake view templates for based on your current database:
    - Pages
    - Sessions
    - Users

Now, i run:
``bin/cake bake template --connection cli -p Fluid --prefix App -v Users``  
And the bake returns:

    Exception: SQLSTATE[08006] [7] could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"? in [/mnt/data/appweb/vendor/cakephp/cakephp/src/Database/Driver/PDODriverTrait.php, line 48]

When i changed the 'default' connect the database using the same values as 'cli' and the same command as before, it works.

This issue is similar to #63.

markstory:
If I'm understanding you correctly, the `-c` option works while `--connection` does not?

LACabeza:
no, both don't works.  
what i meant is `bin/cake bake template -c cli -p Fluid --prefix App` works, it show the database's tables while `bin/cake bake template --connection cli -p Fluid --prefix App Users` don't work, it raises the error.

LACabeza:
An easy way to reproduce this problem is set incorrect parameters on 'default' datasource and correct params on another datasource.

Then run the command `bin/cake bake template -c corret_connection`, it will works, but it will fail when you give the table name to generate the template `bin/cake bake template -c corret_connection TableName`

markstory:
[...]

PR #247, opened 2016-05-04
Uses the --connection parameter to create Table object.

alysson-azevedo:
See #243.


markstory:
Many [tests are failing](https://travis-ci.org/cakephp/bake/jobs/127674115). I think you'll either need to clear the TableRegistry, or use `exists()` like FixtureTask does.

alysson-azevedo:
Yeah, i also [notice the problem](https://github.com/cakephp/bake/issues/243#issuecomment-216717987) before.

And sorry for not testing, i just installed phpunit (and i'll learn about it soon) and now all the tests pass.

PR #257, opened 2016-06-27
Checking a field's type against interface instead of implementation.

mattmemmesheimer:
#98 is still an issue if immutable time objects are being used (specified in `boostrap.php`).  `$value` will be an instance of `Cake\I18n\FrozenTime` instead of `DateTime`.  This change compares fields to the `DateTimeInterface` instead of the implementation `DateTime`.


markstory:
Thanks :+1:

PR #260, opened 2016-07-14
Removed tests (except for fixtures) from `git archive`

### LINKED ISSUE #259: Bake makes IDE unhappy
I use PhpStorm and recently I started seeing this warning: 
screenshot_7

When I stated digging around I found that the class is defined in following files:

* src/Model/Table/UsersTable.php
* vendor/cakephp/bake/tests/comparisons/Model/testBakeWithRules.php

So the problem is that I **cannot** exclude the files from project as they are automatically added as external libraries due to use of composer. Disabling that auto-behavior will create a lot of problems so that's out of question.

After further investigation it turns out that come classes are prefixed (eg. `BakeArticles` and all entities) but some aren't:

* CategoriesProductsTable
* CategoriesTable
* OldProductsTable
* ProductsTable
* ProductVersionsTable
* UsersTable

So all of them might cause warnings in case there is an overlap with the ones in the actual project.

What can be done (my idea/opinion):

* Prefix all of the Tables ***OR***
* Like the core CakePHP package, remove the tests from composer installations (probably safe bet)

Xymanek:
Fixes #259

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -38,3 +38,6 @@ CONTRIBUTING.md export-ignore
 Makefile export-ignore
 phpunit.xml.dist export-ignore
 .travis.yml export-ignore
+tests/test_app export-ignore
+tests/TestCase export-ignore
+tests/comparisons export-ignore


Xymanek:
Not sure why code style check failed... I didn't even change any files that it checks

dereuromark:
You can ignore that, that's from someone elses work :)

dereuromark:
What about Fixture? Are they needed?

Xymanek:
They are kept in core so I thought it should be the same here. Want me to remove them as well?
Also if we remove them, we might break someone's tests ([docs](http://book.cakephp.org/3.0/en/development/testing.html#loading-fixtures-in-your-test-cases))

dereuromark:
Good point. I think this looks good to go.

dereuromark:
Thank you!
It will be part of the next (bugfix) release.

Xymanek:
I'm glad I could help. Also thank you for teaching me how git works :)

PR #261, opened 2016-07-14
Remove deprecated getMock in tests and generated code + new cs rule

antograssiot:


antograssiot:
I can tag a new version once it will be merged to avoid to continue to generate deprecated code in tests.

dereuromark:
Why hiding and not fixing the doc blocks?
Those are not void, but returning null etc as it looks like.

src/Shell/Task/ControllerTask.php
```diff
@@ -46,7 +46,7 @@ class ControllerTask extends BakeTask
      * Execution method always used for tasks
      *
      * @param string|null $name The name of the controller to bake.
-     * @return void
+     * @return void|bool
```
dereuromark: null|bool
void by definition cannot be mixed with others
either you are supposed to check on the return value (non-void) or you are not (void).
mixing void with non-void means this behavior is undefined ;)

src/Shell/Task/SimpleBakeTask.php
```diff
@@ -111,7 +111,7 @@ public function bake($name)
      * Generate a test case.
      *
      * @param string $className The class to bake a test for.
-     * @return string|bool
+     * @return void|string|bool
```
dereuromark: string|bool|null
I would always put the least expected to the end.

antograssiot: changing void to null will unplease phpcs

composer.json
```diff
@@ -22,7 +22,7 @@
         "cakephp/cakephp": ">=3.2.0"
     },
     "require-dev": {
-        "cakephp/cakephp-codesniffer": "dev-master",
+        "cakephp/cakephp-codesniffer": "^2.1",
```
markstory: Thanks :+1:

markstory:
Looks good to me.

PR #262, opened 2016-07-16
Accessibility

mik-laj:
Hello

Web accessibility is important, so today i add a scope attribute to genrated table.
http://webaim.org/techniques/tables/data

Thanks in advance.

### DIFF
--- a/src/Template/Bake/Template/index.ctp
+++ b/src/Template/Bake/Template/index.ctp
@@ -56,9 +56,9 @@ if (!empty($indexColumns)) {
         <thead>
             <tr>
 <% foreach ($fields as $field): %>
-                <th><?= $this->Paginator->sort('<%= $field %>') ?></th>
+                <th scope="col"><?= $this->Paginator->sort('<%= $field %>') ?></th>
 <% endforeach; %>
-                <th class="actions"><?= __('Actions') ?></th>
+                <th scope="col" class="actions"><?= __('Actions') ?></th>
             </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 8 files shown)


antograssiot:
@mik-laj tests will also need to be updated

thinkingmedia:
`scope` is a deprecated attribute in **HTML5**

See the HTML5 specs of obsolete features.

https://www.w3.org/TR/html5/obsolete.html

mik-laj:
`scope` is deprecated only for `td`.  For `th` is still valid.

dereuromark:
That seems to be true.

thinkingmedia:
I'm googling on this now. I've never heard of scope before, but I'm thinking that we should use `area` attributes if we want to support accessibility. So far, Mozilla and W3 have conflicting specs on the topic. One of the reasons I find these hidden attributes so difficult to use.

thinkingmedia:
Okay, I think I'm figuring this out.

Here's a good spec on the role attribute in HTML5.

https://www.w3.org/wiki/PF/XTech/HTML5/RoleAttribute

Here is a spec that explains all the different role attributes, and how they are related to widgets.

https://www.w3.org/TR/2008/WD-wai-aria-20080806/#roles

For tables, there are three role attributes.

"gridcell", "rowheader" and "columnheader"

So I think `scope="col"` should be changed to `role="columnheader"`, and `scope="row"` should be changed to `role="rowheader"`.

Actually, I think rowheader is just for when a `<tr>` has an inner `<th>`.

thinkingmedia:
Here's the start of role attributes for grids.

https://www.w3.org/TR/2008/WD-wai-aria-20080806/#grid

I think that makes for sense now that I read it.

thinkingmedia:
That wasn't the latest document in the links.

https://www.w3.org/TR/wai-aria/

thinkingmedia:
Here is the latest specs for `role="grid"` and the others.

https://www.w3.org/TR/wai-aria/roles#grid

Sorry for the spam comments.

dereuromark:
So all good to be merged then?

Issue #263, opened 2016-07-20
No bake commands for plugin with namespace


bobmulder:
Hi all,

I've found an issue on creating bake commands for a plugin with a namespace. Lets say we have a plugin named `Namespace/MyPlugin`.

Creating a bake command like http://book.cakephp.org/3.0/en/bake/development.html#creating-new-bake-command-options won't show the bake type in the list.

This is because of the following: at line 132 of the file https://github.com/cakephp/bake/blob/master/src/Shell/BakeShell.php#L132 you see that the namespace should be parsed, but the plugin name is parsed. That means `Namespace/MyPlugin` is parsed while it should be `Namespace\MyPlugin`.

For now I've fixed it by replacing line 132 with `str_replace('/', '\\', $plugin),` but I've no idea how to fix this the Cake way (how to get the plugin's namespace?).

Any ideas on this?

lorenzo:
Sorry for coming to this ticket with a silly question... but do you actually need that type of namespace? Coming with a unique name is not that problematic compared to all the pain using the`Vendor/Name` syntax makes

bobmulder:
What kind of pains do you mean Lorenzo? I think choosing for the vendor in the namespace has good and bad points. Would like to know yours...

josegonzalez:
I normally use a vendor in my namespace.

HavokInspiration:
I'm currently contributing to an already existing plugin where the namespace is of the form `Vendor/PluginName` and it is currently impossible to add a bake task to such a plugin.

I think the case should be taken into account.

markstory:
I think its reasonable to to auto-find shells in vendored plugins.

HavokInspiration:
PR #264 opened to address this.

bobmulder:
Thank you @HavokInspiration!

PR #264, opened 2016-07-21
Allows vendored plugins to have bake tasks available

HavokInspiration:
Refs #263


composer.json
```diff
@@ -33,6 +33,7 @@
     "autoload-dev": {
         "psr-4": {
             "BakeTest\\": "tests/test_app/Plugin/BakeTest/src",
+            "Pastry\\PastryTest\\": "tests/test_app/Plugin/PastryTest/src",
```
markstory: :clap:

PR #266, opened 2016-08-22
Require recent PHP and CakePHP versions.

inoas:


inoas:
Ah thanks, yes for 3.4.x it should be 5.6. I am sorry, ahead of my time ;)

markstory:
Not a problem :smile:

PR #271, opened 2016-09-02
DashedRoute::class

inoas:
https://github.com/cakephp/docs/pull/4290

### DIFF
--- a/src/Template/Bake/Plugin/config/routes.php.ctp
+++ b/src/Template/Bake/Plugin/config/routes.php.ctp
@@ -16,6 +16,7 @@
 <?php
 use Cake\Routing\RouteBuilder;
 use Cake\Routing\Router;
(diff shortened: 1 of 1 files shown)


lorenzo:
Thanks

src/Template/Bake/Plugin/config/routes.php.ctp
```diff
@@ -16,11 +16,12 @@
 <?php
 use Cake\Routing\RouteBuilder;
 use Cake\Routing\Router;
+use Cake\Routing\Route\DashedRoute;
```
dereuromark: Alphabetically ordered would be Route before Router

inoas: Done in https://github.com/cakephp/bake/pull/272

PR #273, opened 2016-09-03
Fix tests & Fix version requirements.

markstory:
Pinning to 3.3.* in a plugin makes upgrading a pain, as we'll need plugin authors to continuously do releases as CakePHP does new versions. Instead it is better if we set sane minimum versions and exclude backwards incompatible milestones.

### DIFF
--- a/src/Template/Bake/Plugin/composer.json.ctp
+++ b/src/Template/Bake/Plugin/composer.json.ctp
@@ -20,7 +20,7 @@ $namespace = str_replace('\\', '\\\\', $namespace);
     "type": "cakephp-plugin",
     "require": {
         "php": ">=5.5.9",
-        "cakephp/cakephp": "3.3.*"
+        "cakephp/cakephp": ">=3.3.0 <4.0.0"
     },
     "require-dev": {
         "phpunit/phpunit": "*"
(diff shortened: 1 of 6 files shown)


src/Template/Bake/Plugin/composer.json.ctp
```diff
@@ -20,7 +20,7 @@ $namespace = str_replace('\\', '\\\\', $namespace);
     "type": "cakephp-plugin",
     "require": {
         "php": ">=5.5.9",
-        "cakephp/cakephp": "3.3.*"
+        "cakephp/cakephp": ">=3.3.0 <4.0.0"
```
antograssiot: that where the carret make sense. 
`^3.3.0` but your version work also.

antograssiot: Also given how was the 3.3.0 release we might better recommend 3.3.2 what do you think

markstory: Makes sense

inoas:
Great stuff. Could the minimum PHP and CakePHP version be derived from some central source so that book/docs, composer files in all the plugins like bake etc. use it when making a new release?

markstory:
@inoas Maybe, but I'm not sure its worth the effort and maintenance costs.

Issue #267, opened 2016-08-23
bake tree Table


saeideng:
tree bake tools returned
```
        $validator
            ->integer('lft')
            ->requirePresence('lft', 'create')
            ->notEmpty('lft');

        $validator
            ->integer('rght')
            ->requirePresence('rght', 'create')
            ->notEmpty('rght');

```
we need to comment requirePresence(s) after bake every time

dereuromark:
Those should probably be blacklisted from being bake outputted if the behavior is attached IMO.

markstory:
Pull request open now.

Issue #254, opened 2016-05-31
Bake test for controller with prefix returns error


markstory:
_From @tzvio on May 30, 2016 6:35_

This is a (multiple allowed):
- [x] bug
- [ ] enhancement
- [ ] feature-discussion (RFC)
- CakePHP Version: 3.2.9
- Platform and Target: Ubuntu 16.04  Apache 2.4.18 Postgresql 9.5.2 
  using cake bake from console.
### What you did

I'm having an admin section in my app , 
and some of the controllers use admin prefix.
run this command from console:

`bin/cake bake test Controller --prefix admin CustomersController`
### Expected Behavior

Generate test for CustomersControllerTest in TestCase/Controller/Admin
### Actual Behavior

Received error:

_Error: Unknown option `prefix`_

_Copied from original issue: cakephp/cakephp#8907_

markstory:
It looks like test baking for prefixed controller works fine when the tests are generated at the same time as the controller. This is due to `ControllerTask` customizing the classname when invoking the test task. You might have better luck using `bin/cake bake test Controller 'Admin\CustomsController'` as a workaround/solution.

tzvio:
Thanks, I'll try that

markstory:
Pull request open now.

hisame64:
@markstory 
Please reopen this issue, this problem occurs now yet.
I did `bake controller` with `--prefix` , occurred same problem.  

```
[root@localhost public]# /bin/bash bin/cake bake controller --prefix Cart View

Welcome to CakePHP v3.3.10 Console
---------------------------------------------------------------
App : src
Path: /home/testproj/public/src/
PHP : 7.0.12
---------------------------------------------------------------

Baking controller class for View...

Creating file /home/testproj/public/src/Controller/Cart/ViewController.php
Wrote `/home/testproj/public/src/Controller/Cart/ViewController.php`
Bake is detecting possible fixtures...
Exception: Call to a member function alias() on null in [/home/testproj/public/vendor/cakephp/bake/src/Shell/Task/TestTask.php, line 436]
2016-12-23 17:08:45 Error: [Error] Call to a member function alias() on null
Stack Trace:
#0 /home/testproj/public/vendor/cakephp/bake/src/Shell/Task/TestTask.php(471): Bake\Shell\Task\TestTask->_processModel(NULL)
[...]

Issue #236, opened 2016-04-05
Bake confused by "_id" field name.


ADmad:
_From @thinkingmedia on April 5, 2016 13:45_

This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.2.6
* Platform and Target: Apache2, MySQL, Php 7

### What you did

Added a field named `_id` as an integer to the MySQL table. It was not the primary key of the table as it already had an `id` field that is uuid. 

The `_id` is just a field to store an reference value from a third-party source.

### Expected Behavior

Bake to treat this field as regular value field.

I understand that the field name "_id" does not follow Cake naming conventions. I needed the prefix since all field names mapped to  third-party JSON fields. This just made my job of processing data easier.

### Actual Behavior

Bake generated an empty `belongsTo` association in the table.

```
        $this->belongsTo('', [
            'foreignKey' => '_id',
            'joinType' => 'INNER',
            'className' => 'TheMovieDB.'
        ]);
```

Bake generated an empty `@property` in the table.

```
/**
 * Genres Model
 *
 * @property \Cake\ORM\Association\BelongsTo $
 * @property \Cake\ORM\Association\BelongsToMany $Movies
 * @property \Cake\ORM\Association\BelongsToMany $TvShows
 */
```

Bake generated an empty `@property` in the entity.

```
/**
 * Genre Entity.
 *
 * @property string $id
 * @property int $_id
 * @property \TheMovieDB\Model\Entity\ $
 * @property string $name
 * 
 * @property \Cake\I18n\Time $created
 * @property \Cake\I18n\Time $modified
 */
```

> Note: TheMovieDB is the name of my plugin.

I assume that this can be an easy fix. Bake just needs to validate the format of a field name before treating it as an associated. I think bake is just checking if it ends in `_id`. It needs to end with `_id` and be more then 3 characters.

_Copied from original issue: cakephp/cakephp#8595_

ADmad:
@thinkingmedia Would you like to submit a fix for this?

thinkingmedia:
@ADmad Yes, I will do that.

ionas:
Solution name your non primary/foreign keys or custom primary keys `key`and `_key`

thinkingmedia:
[...]

PR #274, opened 2016-09-03
Fix a few defects

markstory:
Fix a few problems that have been in the issue queue for a while.

* #236 
* #254
* #267 

Each issue has been fixed with a single commit and the messages have my rationale in them :smile:


lorenzo:
Nice :)

dereuromark:
:+1:

PR #275, opened 2016-09-14
Clear TableRegistry on tearDown

davidyell:
During the teardown the `TableRegistry` should be cleared, so that changes made to the table instance during the test-case are not persisted between tests, despite the class instance being unset.

### DIFF
--- a/src/Template/Bake/tests/test_case.ctp
+++ b/src/Template/Bake/tests/test_case.ctp
@@ -89,6 +89,7 @@ class <%= $className %>Test extends TestCase
     public function tearDown()
     {
         unset($this-><%= $subject %>);
+        TableRegistry::clear();
 
         parent::tearDown();
     }
--- a/src/Template/Bake/tests/test_case.ctp
+++ b/src/Template/Bake/tests/test_case.ctp
@@ -89,7 +89,6 @@ class <%= $className %>Test extends TestCase
     public function tearDown()
     {
         unset($this-><%= $subject %>);
-        TableRegistry::clear();
 
         parent::tearDown();
     }


src/Template/Bake/tests/test_case.ctp
```diff
@@ -89,6 +89,7 @@ class <%= $className %>Test extends TestCase
     public function tearDown()
     {
         unset($this-><%= $subject %>);
+        TableRegistry::clear();
```
dereuromark: Did you include the use statement for this as well?

dereuromark:
I usually did that on a per test file basis, as only a minority of tests usually need this.

src/Template/Bake/tests/test_case.ctp
ndm2: Corresponding `use` statement is required.

davidyell: It's included already because the table which the test is baked against is loaded using the TableRegistry, which is why this 'bug' exists.

davidyell:
Yes, I thought so too, but it seems a sensible default to me. I'm not too fussed if it's not merged, but to me it makes sense for bake to produce something which provides sensible defaults.

In my test-case I found it confusing the that the class variable was unset, yet changes to the table class were being persisted.

src/Template/Bake/tests/test_case.ctp
ndm2: That's only the case for table tests though.

ndm2:
I've just had a look at a few projects of mine, and all table tests, and pretty much all integration tests do clear the table registry on tear down. I very well remember running into a lot of trouble with integration test cases that didn't clean up the registry, had a lot of fun with it in the last few weeks.

So I'd say that for these two types of tests (and maybe also for entity tests, even though entities accessing tables is probably very rare), this might indeed be a sane default.

src/Template/Bake/tests/test_case.ctp
markstory: This won't work for Controllers, & Helpers nor should it be required.

davidyell: Shall I add a conditional check to restrict it to table tests?

ndm2: @markstory I think it's fair to assume that controllers are very likely to use tables in most of the cases, and from personal experience I know that it's very easy to trip over this in integration tests.

Take this example, on the second test things will break, because the associations do not exist anymore. So my vote would be, table, controller, and maybe even entity tests.

```php
<?php
namespace App\Controller;

use App\Controller\AppController;

class ArticlesController extends AppController
{
    public function first()
    {
        $this->Articles->associations()->removeAll();
    }

    public function second()
    {
        $this->Articles->find()->contain('Tags')->all();
    }
}
```

```php
<?php
namespace App\Test\TestCase\Controller;

use Cake\TestSuite\IntegrationTestCase;

class ArticlesControllerTest extends IntegrationTestCase
{
    public $fixtures = [
        'app.articles'
    ];

    public function testFirst()
    {
        $this->get('/articles/first');
    }

    public function testSecond()
    {
        $this->get('/articles/second');
    }
}
```

markstory:
Could we push the TableRegistry::clear call into the vase TestCase class?

davidyell:
Sorry, what do you mean @markstory, is that a typo?

markstory:
@davidyell Yes. I meant 'base' instead of 'vase'. I was trying to suggest that clearing the TableRegistry could be done in `Cake\TestSuite\TestCase`.

davidyell:
Would that be hiding it away from the developer too much? Having it in the actual baked test case makes it obvious what is happening in the context of the test as the code is present immediately.

Happy to update the pull request as needed though, if the descision is for the base test case 😄

dereuromark:
I think it is fine in Base class. Better than possibly broken by mssing use Statement here.

PR #276, opened 2016-09-19
Add prefix to (classes) table/entity names

blieb:
We are working with multiple databases, to prevent name conflicts we have to prefix the names of the table/Entity classes.  

I've added --model-prefix to add a prefix to the name of de classes.

example:
bin/cake bake model AccessTokens -c db_oauth --model-prefix oauth

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -175,6 +175,8 @@ class ModelTask extends BakeTask
      */
     public function getTableObject($className, $table)
     {
+        $className = $this->getModelPrefix() . $className;
+
         if (TableRegistry::exists($className)) {
             return TableRegistry::get($className);
         }
@@ -259,6 +261,7 @@ class ModelTask extends BakeTask
     public function findBelongsTo($model, array $associations)
     {
         $schema = $model->schema();
+
         foreach ($schema->columns() as $fieldName) {
             if (!preg_match('/^.+_id$/', $fieldName)) {
                 continue;
@@ -273,11 +276,14 @@ class ModelTask extends BakeTask
                 ];
             } else {
                 $tmpModelName = $this->_modelNameFromKey($fieldName);
+
                 if (!in_array(Inflector::tableize($tmpModelName), $this->_tables)) {
                     $found = $this->findTableReferencedBy($schema, $fieldName);
                     if ($found) {
                         $tmpModelName = Inflector::camelize($found);
                     }
+                } else {
+                    $tmpModelName = $this->getModelPrefix() . $tmpModelName;
                 }
                 $assoc = [
                     'alias' => $tmpModelName,
@@ -795,7 +801,7 @@ class ModelTask extends BakeTask
         if (!empty($this->params['no-entity'])) {
             return null;
         }
-        $name = $this->_entityName((!empty($this->params['model-prefix'])?$this->_camelize($this->params['model-prefix']):'') . $model->alias());
+        $name = $this->_entityName($model->alias());
 
         $namespace = Configure::read('App.namespace');
         $pluginPath = '';
@@ -844,8 +850,9 @@ class ModelTask extends BakeTask
             $namespace = $this->_pluginNamespace($this->plugin);
         }
 
-        $name = (!empty($this->params['model-prefix'])?$this->_camelize($this->params['model-prefix']):'') . $model->alias();
-        $entity = $this->_entityName((!empty($this->params['model-prefix'])?$this->_camelize($this->params['model-prefix']):'') . $model->alias());
+        $name = $model->alias();
+
+        $entity = $this->_entityName($model->alias());
         $data += [
             'plugin' => $this->plugin,
             'pluginPath' => $pluginPath,
@@ -973,6 +980,16 @@ class ModelTask extends BakeTask
     }
 
     /**
+     * Get the prefix for model if exist.
+     *
+     * @return string
+     */
+    public function getModelPrefix()
+    {
+        return !empty($this->params['model-prefix']) ? $this->_camelize($this->params['model-prefix']) : '';
+    }
+
+    /**
      * Gets the option parser instance and configures it.
      *
      * @return \Cake\Console\ConsoleOptionParser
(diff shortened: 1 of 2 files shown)


dereuromark:
Wouldn't `--table-prefix` be the more appropriate them here? After all this should only be relevant for finding the prefixed table in the DB schema.

blieb:
@dereuromark This only adds a prefix to the files/classes not the database table itself.

it adds the prefix to both of the model parts. Table and Entity. 

example: Table in database is 'accessTokens'

Model/Table/OauthAccessTokensTableTest.php
Model/Entity/OauthAccessToken.php

Not sure if --table-prefix is a correct name. Sounds more like a prefix to the table in the mysql database.

dereuromark:
IMO it should be the other way around.
If a DB (for whatever legacy reason) has prefixed tables like `foo_users`, bake would not find it properly.
A `--table-prefix foo_` would then find it and bake it as `Users` table and `User` entity and just set the manual table('foo_users').

Why would you want it the other way around? That is where plugins and namespaces come into play.

blieb:
I see there is an other pull request open for other way around.

I think making plugins for every database will be a little overkill and models are in a lot of different folders. 

We have 2 databases with table 'products' in it. Thats why we have to prefix the class names.

thinkingmedia:
Just to clarify. You have two databases with tables of the same name. So you want to differentiate between them?

I think using two plugins would be a better approach. A plugin for each database.

thinkingmedia:
@blieb sorry, just read your last comment.

I think prefixing table classes conflicts with cakephp's naming conventions.

dereuromark:
Most importantly you will need to different DB connections here.

blieb:
We have a big project. Putting everything in a single database will be a mess. 

We sell (our) products to our customers. Most of them are stores. One of the products is a webshop.
So the second database will have products of the stores. 

(Today I shocked a little, we have 20+ databases that have to work with this project)

I believe cakephp does not have a good solution with working with multiple databases. It is possible. But I have to prefix the classes to prevent name conflicts.

inoas:
you can set the prefix through $this->table('') within table classes I think.

dereuromark:
I already told him that solution above.

blieb:
@inoas i Dont have to prefix the tables in mysql but in cakephp to prevent classname conflicts.

inoas:
You want to use Namespaces for that. But honestly. Plugins will give you namespaces and are exactly what you are looking for IMHO.

inoas:
I am not saying this PR feature is bad btw...!

src/Shell/Task/ModelTask.php
```diff
@@ -795,7 +795,7 @@ public function bakeEntity($model, array $data = [])
         if (!empty($this->params['no-entity'])) {
             return null;
         }
-        $name = $this->_entityName($model->alias());
+        $name = $this->_entityName((!empty($this->params['model-prefix'])?$this->_camelize($this->params['model-prefix']):'') . $model->alias());
```
markstory: You'll need to add spaces around the operators here.

src/Shell/Task/ModelTask.php
```diff
@@ -844,8 +844,8 @@ public function bakeTable($model, array $data = [])
             $namespace = $this->_pluginNamespace($this->plugin);
         }
 
-        $name = $model->alias();
-        $entity = $this->_entityName($model->alias());
+        $name = (!empty($this->params['model-prefix'])?$this->_camelize($this->params['model-prefix']):'') . $model->alias();
+        $entity = $this->_entityName((!empty($this->params['model-prefix'])?$this->_camelize($this->params['model-prefix']):'') . $model->alias());
```
markstory: Can this repeated logic be extracted into a method?

inoas: Spaces also around operators, also.

blieb:
the logic is now in apart function, and also added the spaces around operators. 

Also fixed the relationships. 
I have to jump in to the tests because I never created one yet :)

src/Shell/Task/ModelTask.php
dereuromark: He could just reuse $name from above which does exactly the same, right?
And just say

    $entity = $this->_entityName($name);

src/Shell/Task/ModelTask.php
```diff
@@ -973,6 +980,16 @@ public function getTable($name)
     }
 
     /**
+     * Get the prefix for model if exist.
+     *
+     * @return string
+     */
+    public function getModelPrefix()
```
dereuromark: But does it need to be part of the public API here i wonder? It is only used internally here, right?

blieb: Oops need to be private? That's true, only used internally

dereuromark: protected

blieb:
@markstory tried to create a test. But for some reason it lost my params during 'bakeTable'. Do I miss something?

`     /**
     * test bake() with a -model-prefix param
     *
     * @return void
     */
    public function testBakeTableWithModelPrefix()
    {

        $this->Task->params['model-prefix'] = 'test';

        $filename = $this->_normalizePath(APP . 'Model/Table/TestBakeArticlesTable.php');
        $this->Task->expects($this->once())->method('createFile')
            ->with($filename, $this->stringContains('class TestBakeArticlesTable extends'));


        $model = TableRegistry::get('BakeArticles');
        $result = $this->Task->bakeTable($model);
        $this->assertSameAsFile(__FUNCTION__ . '.php', $result);
    }
`

markstory:
@blieb Odd that is how the other tests work too.

blieb:
now I merged the master branch to my branch, now I have more unittests errors, also the master branch is failing the tests :(

dereuromark:
Actually, looks good.
You just have to fix your CS errors: https://travis-ci.org/cakephp/bake/jobs/170403000

blieb:
@dereuromark I've fixed the CS, missed a space.

But https://ci.appveyor.com/project/cakephp/bake/build/1.0.421 also still fails? I don't see why. (and 2 errors I merged from the master branch)

dereuromark:
Appveyor always fails usually. Travis must be fine.

lorenzo:
Is this waiting for review?

blieb:
It is still waiting, but I think there are some problems with the tests. 

I've decided in my project to move the database tables to different plugins instead of prefixing the model names. so if you think this is not needed you can also close this pull request.

dereuromark:
I see more benefit for allowing prefixed tables to be used without the prefix showing up (the opposite of this PR).
:-1: for continuing this one.

Issue #295, opened 2017-01-17
Why don't shell methods return?


davidyell:
When baking a shell, the `main()` method is created, and it looks like the following.

```php
    /**
     * main() method.
     *
     * @return bool|int Success or error code.
     */
    public function main()
    {
        $this->out($this->OptionParser->help());
    }
```

However, when I run PHPCS with the CakePHP standard I get a warning, saying, that the function has no return so should be `@return void`. 

![PHPStormPHPCS](http://i.imgur.com/5G3qokv.png)

I'm not sure what the correct course of action is here, as the `out()` method does actually have the same docblock return.

dereuromark:
Technically, returning null means here true from the implementation.
You could make a PR that also adds `|null`.

But I agree with you that it would be cleaner to expect either bool or int, and at the end such a method would return true :)
Same issue exists with actions.

For now, the PR with the doc block change might be the easiest solution in 3.x.

PR #296, opened 2017-01-20
Fix doc blocks and remove not necessary else.

dereuromark:
Resolves https://github.com/cakephp/bake/issues/295 documentation part.
Also cleanup up some warnings about not properly working with void vs null.

### DIFF
--- a/src/Shell/Task/TemplateTask.php
+++ b/src/Shell/Task/TemplateTask.php
@@ -340,7 +340,7 @@ class TemplateTask extends BakeTask
     /**
      * handle creation of baking a custom action view file
      *
-     * @return void|null
+     * @return void
      */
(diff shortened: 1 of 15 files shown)


src/Shell/Task/TemplateTask.php
```diff
@@ -366,7 +366,8 @@ public function customAction()
         if (strtolower($looksGood) === 'y') {
             $this->bake($action, ' ');
 
-            return $this->_stop();
+            $this->_stop();
+            return;
```
dereuromark: Boah, this is really annoying ;)

markstory:
Looks good to me.

davidyell:
Thanks @dereuromark 👍

PR #297, opened 2017-01-23
Add typehinting to baked templates

hmic:


src/Template/Bake/Template/login.ctp
```diff
@@ -13,6 +13,12 @@
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 %>
+<?php
+
+/**
+  * @var <%= $namespace %>\View\AppView $this
```
dereuromark: Is there a prepending \ ? Should be

hmic: You be right :-/

dereuromark:
Thats a valuable addition to templates.

src/Template/Bake/Layout/default.ctp
```diff
@@ -13,4 +13,10 @@
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 %>
+<?php
+
+/**
+  * @var \<%= $namespace %>\View\AppView $this
```
markstory: In the context of a plugin this won't be `PluginName\View\AppView` it will still be `App\View\AppView`. It think you might need to always have the app's namespace available.

markstory:
The tests also need to be updated. The travis build configuration has a syntax error which is why travis passed when it shouldn't have.

dereuromark:
Rebasing on top of master should do the trick.

src/Template/Bake/Layout/default.ctp
hmic: How would that possibly work?

hmic: Oh, easy :1st_place_medal:

hmic:
Yes, just noticed...

dereuromark:
We can now also add annotations to already baked code via https://github.com/dereuromark/cakephp-ide-helper/commit/27ba7cb7260402d0e78c69fa4034b84b8cbfb190 and the Annotation shell, it can modify existing code.

PR #303, opened 2017-01-29
Fixing a bug that appears when trying to bake plugin models

burzum:
This fixes a problem that happens when you already have an existing model of the same name in your app but want to bake one of the same name in your plugin. I've used as well the --table option. My main app is using an users table, now for a plugin I wanted to bake a Users model that is using the wp_users table from another DB. This was not possible without this change.

Tried this:

```sh
.\bin\cake bake model Users --table wp_fk_users -c wordpress -p Burzum/WpFrontend
```

Resulting in:

```sh
Error: [Cake\Database\Exception] SQLSTATE[42S02]: Base table or view not found: 1146 Table 'wordpressfk.users' doesn't exist
```

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -175,6 +175,11 @@ class ModelTask extends BakeTask
      */
     public function getTableObject($className, $table)
     {
+        $plugin = $this->param('plugin');
+        if (!empty($plugin)) {
+            $className = $plugin . '.' . $className;
+        }
+
         if (TableRegistry::exists($className)) {
             return TableRegistry::get($className);
         }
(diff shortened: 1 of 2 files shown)


src/Shell/Task/ModelTask.php
```diff
@@ -175,6 +175,11 @@ public function all()
      */
     public function getTableObject($className, $table)
     {
+        $plugin = $this->param('plugin');
+        if (!empty($plugin)) {
+            $className = $plugin . '.' . $className;
```
markstory: Any way we could get a test for this?

burzum: If you tell me how? All I know is that testing console stuff is pain in the ass. :(

dereuromark: We could unit test the getTableObject() here.

Issue #299, opened 2017-01-24
Bake Model with Unique date field


markstory:
_From @jonathanlaf on January 23, 2017 20:29_

This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.3.8
* Platform and Target: Bake in windows terminal, querying a InnoDB under a distant cPanel

### What you did
bake.bat bake all releases

### What happened
In ReleasesTable.php in the validator it add 
`->add('release_date', 'unique', ['rule' => 'validateUnique', 'provider' => 'table']);`
Then when I try to add a date even if there's no presence of this date anywhere I receive this error while saving : "The provided value is invalid". If I remove this line from my model, everything work has expected, I can add new date, and if it's not unique, I receive an error stating that value is already in use.

### What you expected to happen
I don't exactly know what does `->add('release_date', 'unique', ['rule' => 'validateUnique', 'provider' => 'table']);` really do, since `$rules->add($rules->isUnique(['release_date']));` is already present and do exactly has expected

Before you open an issue, please check if a similar issue already exists or has been closed before.

_Copied from original issue: cakephp/cakephp#10094_

markstory:
_From @saeideng on January 23, 2017 20:37_

what you not using `$rules->add($rules->isUnique(['release_date']));` in application rules?

markstory:
Closing as a pull request is open that fixes the issue in theory.

markstory:
_From @cleptric on January 24, 2017 1:21_

The problem is inside `Table::validateUnique`.

```
        foreach ($values as $field) {
            if ($field !== null && !is_scalar($field)) {
                return false;
            }
        }
```

`$filed` is an array when you're saving a date and so the `!is_scalar($field)` check is true.

markstory:
Sounds like we shouldn't generate unique validation rules for date/time/datetime fields. Moving this issue to cakephp/bake as that is where the code generation should be updated.

PR #309, opened 2017-02-11
Don't generate unique validators for datetime fields.

markstory:
Date & Time fields should how have unique validators generated, as the validator method doesn't work well on non-scalar fields.

Refs #299


lorenzo:
Good job!

Issue #310, opened 2017-02-20
CakePHP 3.4.1 new installation model bake showing derecated method


markstory:
_From @mdmamunhasan on February 20, 2017 7:4_

$this->table('users');
$this->displayField('name');
$this->primaryKey('id');

_Copied from original issue: cakephp/cakephp#10258_

markstory:
_From @dereuromark on February 20, 2017 9:42_

These kind of tickets should be opened in https://github.com/cakephp/bake instead.

markstory:
_From @inoas on February 20, 2017 14:56_

Wrong issue

~~I think there is a lot that can be done for the default bake templates and layout such as new setters/getters, schema constraint detection, schema foreign key/assoc detection, singular/plural convention switching for tablenames and foreign keys, frontend content and presentation decoupling, div/class soup clearing, mobile first, async js belongsTo lookup, unobstrusive xhr delete confirms, etc... to improve it as a RAD/XP tool before adding search to it, but thats just my 2 cents~~

Issue #294, opened 2017-01-15
@CakePHP3.4 Bake Output: Replace FormHelper::*input*() with FormHelper::*control*()


inoas:
CakePHP 3.4 will bring a deprecation of *FormHelper::input()* and similar methods in favour of *FormHelper::control()*. Bake template output should reflect that.

### Issue

From there-on bake will not be backwards compatible with < 3.4, so it would require a new major release.

antograssiot:
We can keep it as is for now and keep it working with as many version as it can right now.
I'very been thinking about this and a lot of deprecated function are added by bake but I think that itranslation better benefits to people if it keep working and be supported for most 3.x version

inoas:
We can also switch around the template output by creating a helper method and detect Configure::version()?

antograssiot:
That's a possibility that I didn't considered.
Testing the output will probably not be fun...

inoas:
I am afraid adoption will lack if bake doesn't do it.
There could be a switch around the test case to use based on Configure::version() again, or the version could be statically set to 3.4 for the tests.

markstory:
We can also use composer requirements :)

PR #312, opened 2017-02-21
Update bake for 3.4 deprecations

markstory:
Update bake to use un-deprecated methods that were added in 3.4.0

Refs #294 #310

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: php
 
 php:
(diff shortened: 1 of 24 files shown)


antograssiot:
Glad you removed support for <3.4 :+1:
It was stopping me removing deprecations as testing was too complicated.
I can give you a hand in removing the other method calls

PR #313, opened 2017-02-22
3.4 deprecations - Remove Form::input generation

antograssiot:
Also remove description() and epilog() usage in shell tasks


markstory:
Nice work!

PR #302, opened 2017-01-29
Adding table prefix handling

burzum:
Replaces https://github.com/cakephp/bake/pull/191

PR #315, opened 2017-02-23
Adding table prefix handling

dereuromark:
This finishes https://github.com/cakephp/bake/pull/302 by adding a test case.


Oxicode:
+1

PR #318, opened 2017-03-01
Fix phpstan errors

cleptric:
Fixed some errors which came up with phpstan.

Wrong FQDN in a docblock.
`\Cake\Console\Shell::_stop();` throws an exception, so no need to return here.
`\Bake\Shell\Task\TestTask::hasMockClass ()` only accepts one parameter.

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -953,13 +953,13 @@ class ModelTask extends BakeTask
             $this->err(
                 'Connections need to implement schemaCollection() to be used with bake.'
             );
-            $this->_stop();
+            return $this->_stop();
         }
         $schema = $db->schemaCollection();
         $tables = $schema->listTables();
         if (empty($tables)) {
             $this->err('Your database does not have any tables.');
-            $this->_stop();
+            return $this->_stop();
         }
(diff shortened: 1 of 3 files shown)


src/Shell/Task/ModelTask.php
```diff
@@ -953,15 +953,13 @@ protected function _getAllTables()
             $this->err(
                 'Connections need to implement schemaCollection() to be used with bake.'
             );
-
-            return $this->_stop();
+            $this->_stop();
```
markstory: We were returning here to make the behavior consistent when `_stop()` is mocked.

dereuromark: But ist this still necessary with throwing exceptions?

markstory: Do the mocks throw exceptions?

markstory:
Now there are phpcs errors 😅

src/Shell/Task/ModelTask.php
dereuromark: Maybe we shouldnt mock them?

cleptric:
Hmm, codecov complains about a missing uploaded a coverage report.

Issue #323, opened 2017-03-12
Associations annotations are incomplete


dereuromark:
The associations are added as
```
@property \Cake\ORM\Association\BelongsTo $Events
```
for examle

see https://github.com/cakephp/bake/blob/master/src/Template/Bake/Model/table.ctp#L21

But that is not enough, it also needs the concrete table object, otherwise chaining fails.

So it should be:
```
@property \App\Model\Table\EventsTable|\Cake\ORM\Association\BelongsTo $Events
```

dereuromark:
Resolved with https://github.com/cakephp/bake/pull/340

PR #324, opened 2017-03-12
Fix up relations in annotations.

dereuromark:
This is a first shot at https://github.com/cakephp/bake/issues/323

Also needed for IdeHelper to generate more correct output (refs https://github.com/dereuromark/cakephp-ide-helper/issues/5).

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -15,6 +15,7 @@
 namespace Bake\Shell\Task;
 
 use Cake\Console\Shell;
+use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Database\Schema\Table as SchemaTable;
 use Cake\Datasource\ConnectionManager;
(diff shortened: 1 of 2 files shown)


dereuromark:
Anyone able to lend me a hand here to get things merged?
Not sure why it worked for me, but the tests fail.

ndm2:
If you wanted to avoid pollution, then maybe a helper method that builds association info might be an option too. This could easily be extended with various other details in the future with interfering with other code that uses the bare association data.

```php
public function getAssociationInfo(array $associations)
{
	$info = [];

	$namespace = Configure::read('App.namespace');
	foreach ($associations as $type => $assocs) {
		foreach ($assocs as $assoc) {
			$name = isset($assoc['className']) ? $assoc['className'] : $assoc['alias'];
			$targetClass = $namespace . '\Model\Table\\' . $name . 'Table';

			$info[$assoc['alias']] = [
				'targetFqn' => '\\' . $targetClass
			];
		}
	}

	return $info;
}
```

```php
$targetFqn = $associationInfo[$assoc['alias']]['targetFqn'];
$annotations[] = "@property {$targetFqn}|\Cake\ORM\Association\\{$typeStr} \${$assoc['alias']}";
```

Or maybe go the kinda breaking change way, and change the `$associations` structure so that the association details and additional info are separated, like

```php
[
	'belongsToMany' => [
		'Products' => [
			'info' => [
				targetFqn' => '\App\Model\Table\ProductsTable'
			],
			'association' => [
				'foreignKey' => 'category_id',
				'targetForeignKey' => 'product_id',
				'joinTable' => 'categories_products'
			]
		]
	]
]
```

dereuromark:
@ndm2 Do you want to make an alternative PR with your improved version?
You can also put commits on top of this one, if you want to.

ndm2:
It's not that I really really really want to, but I can do it if I find a little more free time the next days :)

Issue #325, opened 2017-03-19
Incorrect annotation for date/time properties of entities.


ADmad:
Currently for baked entities, when annotating time/date/datetime/timestamp properties hardcoded class `\Cake\I18n\Time` is used. Instead bake should introspect the class configured for particular field type and use proper class like `Cake\I18n\FrozenTime`,  `Cake\I18n\FrozenDate` etc.

ADmad:
Related PR has been merged.

PR #326, opened 2017-03-19
Use correct class when annotating entities' datetime properties.

ADmad:
Refs #325


dereuromark:
Nice, looks like as soon as this one is tagged and released (and with the right CakePHP version) this will also automatically work inside the IdeHelper annotations then.

markstory:
Tests [are failing](https://travis-ci.org/cakephp/bake/jobs/212800786)

Issue #61, opened 2015-03-24
Bake should not generate validation for foreign key columns


lorenzo:
This kind of validation is done with the rules system and most users have problems because of the default validation

markstory:
I can take care of that.

markstory:
Pull request up now.

PR #327, opened 2017-04-01
Use foreign keys in validation rules

saeideng:
Refs https://github.com/cakephp/cakephp/issues/10355 cc @markstory 

Revert https://github.com/cakephp/bake/pull/62 cc @markstory 
old issue https://github.com/cakephp/bake/issues/61 cc @lorenzo

### DIFF
--- a/tests/comparisons/Model/testBakeAssociationDetectionCategoriesProductsTable.php
+++ b/tests/comparisons/Model/testBakeAssociationDetectionCategoriesProductsTable.php
@@ -48,6 +48,25 @@ class CategoriesProductsTable extends Table
     }
 
     /**
+     * Default validation rules.
+     *
+     * @param \Cake\Validation\Validator $validator Validator instance.
+     * @return \Cake\Validation\Validator
+     */
+    public function validationDefault(Validator $validator)
+    {
+        $validator
+            ->integer('category_id')
+            ->allowEmpty('category_id', 'create');
+
+        $validator
+            ->integer('product_id')
+            ->allowEmpty('product_id', 'create');
+
+        return $validator;
+    }
+
+    /**
      * Returns a rules checker object that will be used for validating
      * application integrity.
      *
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -131,7 +131,7 @@ class ModelTask extends BakeTask
         $displayField = $this->getDisplayField($tableObject);
         $propertySchema = $this->getEntityPropertySchema($tableObject);
         $fields = $this->getFields();
-        $validation = $this->getValidation($tableObject, $associations);
+        $validation = $this->getValidation($tableObject);
         $rulesChecker = $this->getRules($tableObject, $associations);
         $behaviors = $this->getBehaviors($tableObject);
         $connection = $this->connection;
@@ -596,16 +596,8 @@ class ModelTask extends BakeTask
 
         $validate = [];
         $primaryKey = (array)$schema->primaryKey();
-        $foreignKeys = [];
-        if (isset($associations['belongsTo'])) {
-            foreach ($associations['belongsTo'] as $assoc) {
-                $foreignKeys[] = $assoc['foreignKey'];
-            }
-        }
+
         foreach ($fields as $fieldName) {
-            if (in_array($fieldName, $foreignKeys)) {
-                continue;
-            }
             $field = $schema->column($fieldName);
             $validation = $this->fieldValidation($schema, $fieldName, $field, $primaryKey);
             if (!empty($validation)) {
--- a/tests/TestCase/Shell/Task/ModelTaskTest.php
+++ b/tests/TestCase/Shell/Task/ModelTaskTest.php
@@ -831,29 +831,6 @@ class ModelTaskTest extends TestCase
     }
 
     /**
-     * test getting validation rules and exempting foreign keys
-     *
-     * @return void
-     */
-    public function testGetValidationExcludeForeignKeys()
-    {
-        $model = TableRegistry::get('BakeArticles');
-        $associations = [
-            'belongsTo' => [
-                'BakeUsers' => ['foreignKey' => 'bake_user_id'],
-            ]
-        ];
-        $result = $this->Task->getValidation($model, $associations);
-        $expected = [
-            'title' => ['valid' => ['rule' => false, 'allowEmpty' => false]],
-            'body' => ['valid' => ['rule' => false, 'allowEmpty' => true]],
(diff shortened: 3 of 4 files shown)


markstory:
We also have the `allowNullableNulls` option now. Instead of reverting the earlier change, could bake enable this option when a nullable foreign key is found?

saeideng:
generated for null schema
```
        $validator
            ->integer('product_id')
            ->allowEmpty('product_id', 'create');
```
for not null schema 
```
        $validator
            ->integer('product_id')
            ->requirePresence('product_id', 'create')
            ->notEmpty('product_id');
```

saeideng:
i think this change not related to `allowNullableNulls` option

saeideng:
probably you say bake should bake like : 
```
$rules->add($rules->existsIn(['parent_id', 'site_id'],'ParentNodes',['allowNullableNulls' => false]));
```

markstory:
Right now we don't generate any validators for foreign keys, and instead rely on the `ExistsIn` rules. Enabling `allowNullableNulls` should allow null to be accepted as a valid value. From what I've understood saving nulls, which currently causes rules failures is the problem. I may be misunderstanding the problem though.

saeideng:
problem is not `allowEmpty` or `notEmpty` rule
problem is `requirePresence` rule

saeideng:
```
CREATE TABLE IF NOT EXISTS `comments` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `user_id` int(11) NOT NULL,
  `title` varchar(1000) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB  DEFAULT CHARSET=utf8 ;
```

```
    public function validationDefault(Validator $validator)
    {
        $validator
            ->integer('id')
            ->allowEmpty('id', 'create');


        $validator
            ->requirePresence('title', 'create')
            ->notEmpty('title');

        return $validator;
    }

    public function buildRules(RulesChecker $rules)
    {
        $rules->add($rules->existsIn(['user_id'], 'Users'));

        return $rules;
    }
}
```

```
//controller
		$data=[
			'title'=>'new title',
			//'user_id'=>1 // user_id not present 
		];
		$entity=$this->Comments->newEntity($data);
		$this->Comments->save($entity);
```

```
BEGIN
INSERT INTO comments (title) 
VALUES 
  ('new title')
COMMIT
```

```
//database
id=an Id
user_id=0
title="new title"
```

result of 
`$rules->add($rules->existsIn(['user_id'], 'Users'));`
or
`$rules->add($rules->existsIn(['user_id'], 'Users',['allowNullableNulls' => false]));`
or
`$rules->add($rules->existsIn(['user_id'], 'Users',['allowNullableNulls' => true]));`
is some

we should validate it in `validationDefault()`

`allowNullableNulls` is not related to this problem

saeideng:
> and instead rely on the ExistsIn rules. 

i think no, is not true

saeideng:
> and instead rely on the ExistsIn rules. 

i think no, is not true
this lines executed in my exam
https://github.com/cakephp/cakephp/blob/master/src/ORM/Rule/ExistsIn.php#L109-L111

saeideng:
Mentioned lines in core is invalid
That lines applyed in both save ,update
We can fix it for all programs (old,new) by fixing existsIn()
Or fixing bake for new programs
i think validator is good place to validation ,what you think?

saeideng:
I think i found another problem in core that make conflict with PR
We can not use inline setter like bellow code after merging thisPR !?
```
$entity=patchEntity(....);
$entity->user_id=1;
```
I want to resolve not present field in core ,closing until that time

saeideng:
Error will remove   in setDirty after inline set
https://github.com/cakephp/cakephp/blob/master/src/Datasource/EntityTrait.php#L761
Reopen until found a way

markstory:
Ok, I was under the impression you were having problems with nullable foreign keys. Not required foreign key columns.  The code you linked to in ExistsIn cannot be removed as it will cause breakage in existing applications. It sounds like we should at least be adding the `requirePresence('create')` validation rules, which is part of what this pull request is doing. 

I'm not sure we need to also generate the unique validators when the foreign key columns also have SQL constraints though. The `existsIn` rule should be enough, and I think was the problematic part we wanted to remove in #61

saeideng:
> The code you linked to in ExistsIn cannot be removed as it will cause breakage in existing application.

I want to fix it instead of removing for old programs or programs that not using bake
I have an idea that i will open PR as soon

saeideng:
I think you can merge this
Or you need just adding requirePresence() rule ?

markstory:
@saeideng Right now bake will start re-adding `validateUnique` validators for foreign keys that have SQL constraints. I don't think that is desirable based on the confusion it caused earlier.

saeideng:
ping


some of people not using foreign key constraints ,because cakephp handle it

dereuromark:
@markstory What shall we do here?

tests/comparisons/Model/testBakeAssociationDetectionCategoriesProductsTable.php
```diff
@@ -47,6 +47,25 @@ public function initialize(array $config)
...
+     * Default validation rules.
+     *
+     * @param \Cake\Validation\Validator $validator Validator instance.
+     * @return \Cake\Validation\Validator
+     */
+    public function validationDefault(Validator $validator)
+    {
+        $validator
+            ->integer('category_id')
+            ->allowEmpty('category_id', 'create');
```
markstory: Is the difference with this implementation vs the old one that keys can be blank on create?

saeideng:
we can remove  validation when foreign key was `null` able 
but for `not null` column we should add bellow validation IMO
```php
        $validator
            ->requirePresence('category_id', 'create')
            ->notEmpty('category_id');
            //or
            ->notEmpty('category_id',  'create');
```

tests/comparisons/Model/testBakeAssociationDetectionCategoriesProductsTable.php
saeideng: looks for `null` able column validation is not necessary , but for `not null` it is important

dereuromark:
needs rebase.

dereuromark:
@saeideng Shall we finish this PR?

dereuromark:
Also refs https://github.com/cakephp/migrations/issues/385

dereuromark:
@saeideng Are u still interested in finishing this up?

saeideng:
let me to check again whole changes in next few days

dereuromark:
@saeideng ping
We should close otherwise.

saeideng:
I think it is valid yet , but whole PR needs to re create with change on command class
I want to close now maybe until next concrete reason

PR #331, opened 2017-05-06
feat: Add pagination annotation to baked controllers

### LINKED ISSUE #329: Paginate annotation
Controllers should get this annotation if index method and paginate() is baked:
```php
/**
 * @method \App\Model\Entity\Release[] paginate($object = null, array $settings = [])
 */
class ReleasesController extends AppController
{
```

josegonzalez:
Closes #329


dereuromark:
Thanks a lot for picking this up!

PR #333, opened 2017-05-07
feat: Update fixture generation for small/tiny integers

### LINKED ISSUE #322: Update fixture generation for small/tiny integers
With cakephp supporting `smallinteger`/`tinyinteger` types as of cakephp/cakephp#10377, we should update fixture generation to build these column types and populate them with default values.

josegonzalez:
Closes #322


josegonzalez:
Tests will fail until CakePHP 3.5.0 drops, so this should not be merged until then.

Issue #284, opened 2016-12-02
Detect varchar defined length and add a ->maxLength rule for baked tables


steinkel:
Detect a specific length set in the column type, for example VARCHAR(12) and create a default maxLength validation rule in validationDefault method.

markstory:
Sounds like a good idea to me.

PR #335, opened 2017-05-16
Add a maxLength rule for text columns

dmromanov:
#284
This pull-request depends on cakephp/cakephp/issues/11295 .

### DIFF
--- a/src/View/Helper/BakeHelper.php
+++ b/src/View/Helper/BakeHelper.php
@@ -292,8 +292,19 @@ class BakeHelper extends Helper
         $validationMethods = [];
 
         foreach ($rules as $ruleName => $rule) {
-            if ($rule['rule'] && !isset($rule['provider'])) {
+            if ($rule['rule'] && !isset($rule['provider']) && !isset($rule['args'])) {
                 $validationMethods[] = sprintf("->%s('%s')", $rule['rule'], $field);
+            } elseif ($rule['rule'] && !isset($rule['provider'])) {
+                $formatTemplate = "->%s('%s')";
+                if (!empty($rule['args'])) {
+                    $formatTemplate = "->%s('%s', %s)";
+                }
+                $validationMethods[] = sprintf(
+                    $formatTemplate,
+                    $rule['rule'],
+                    $field,
+                    implode(', ', $this->escapeArguments($rule['args']))
+                );
             } elseif ($rule['rule'] && isset($rule['provider'])) {
                 $validationMethods[] = sprintf(
                     "->add('%s', '%s', ['rule' => '%s', 'provider' => '%s'])",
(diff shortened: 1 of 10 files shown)


src/Template/Bake/Model/table.ctp
```diff
@@ -106,8 +106,10 @@ class <%= $name %>Table extends Table
 foreach ($validation as $field => $rules):
     $validationMethods = [];
     foreach ($rules as $ruleName => $rule):
-        if ($rule['rule'] && !isset($rule['provider'])):
+        if ($rule['rule'] && !isset($rule['provider']) && !isset($rule['args'])):
             $validationMethods[] = sprintf("->%s('%s')", $rule['rule'], $field);
+        elseif ($rule['rule'] && !isset($rule['provider'])):
+            $validationMethods[] = sprintf("->%s('%s', %s)", $rule['rule'], $field, implode(', ', $rule['args']));
```
markstory: Won't this have issues if the arguments are strings?

dmromanov:
@markstory String values are now wrapped with quotes.

Tests are failing because each DBMS has different length limits for string-type fields.

src/View/Helper/BakeHelper.php
```diff
@@ -163,6 +163,23 @@ public function classInfo($class, $type, $suffix)
...
+     * Wrap string arguments with quotes
+     *
+     * @param array $args array of arguments
+     * @return array
+     */
+    public function escapeArguments($args)
+    {
+        return array_map(function ($v) {
+            if (!is_numeric($v)) {
+                $v = "'$v'";
```
markstory: You should probably escape `'` inside strings :smile:

markstory:
Tests will need to be passing before we can get this merged.

dmromanov:
Ok, I'll change the code.

ADmad:
Tests still failing.

dmromanov:
Still working on it.

dmromanov:
My changes fails to pass validation on some DBMS, because of some inconsistency between them. E.g. in SQLite, text fields can have infinite length, while in MySQL, it's not.

To work around this, I am thinking about two possible approaches:
1.  To mock a table schema and build it manually in tests, where a table schema is required.
2. Make DBMS-specific tests, that would be marked as skipped, if they are run under different DMBS.

@ADmad @markstory Which one is better? Or maybe you can suggest a better solutions?

markstory:
Could we test this validator against varchar columns instead of text ones? Those should have more stable lengths across different db vendors.

dmromanov:
Depends on cakephp/cakephp/issues/11295

tests/Fixture/NumberTreesFixture.php
```diff
@@ -0,0 +1,138 @@
...
+namespace Bake\Test\Fixture;
+
+use Cake\TestSuite\Fixture\TestFixture;
+
+/**
+ * NumberTreeFixture
+ *
+ * Generates a tree of data for use testing the tree behavior
+ */
+class NumberTreesFixture extends TestFixture
```
markstory: Why do we need this fixture?

dmromanov: Overall it's a copy of CakePHP's core NumberTreesFixture, but with a change: "name" column now has a length. Since the length attribute is not needed in CakePHP core, and is not covered with tests, it is a subject to change in the future. So I decided to copy it into the repository.

markstory: I didn't see where the fixture was used. My mistake.

dmromanov:
Now with TWIG support.

Issue #316, opened 2017-02-24
Add more typehinting to baked templates


dereuromark:
After https://github.com/cakephp/bake/pull/297
it would be also good to add typehinting for the variables in the add/edit/view and index views:

Right now I manually have to add this typehint for index views for example:
```
/** @var \App\Model\Entity\Mailing[] $mailings */
foreach ($mailings as $mailing):
```

We could just add at the top those ctps:
```
/** 
 * @var \<%= $namespace %>\View\AppView $this
 * @var \<%= $namespace %>\Model\Entity\<%= $name %>[] $<%= $var %> */
```
for index

and the singular version for the other views.

dereuromark:
Refs https://github.com/dereuromark/cakephp-ide-helper/commit/27ba7cb7260402d0e78c69fa4034b84b8cbfb190

PR #337, opened 2017-05-18
Additional typehinting for baked templates

dmromanov:
#316

### DIFF
--- a/src/Template/Bake/Template/index.ctp
+++ b/src/Template/Bake/Template/index.ctp
@@ -16,6 +16,7 @@
 <?php
 /**
   * @var \<%= $namespace %>\View\AppView $this
+  * @var \<%= $namespace %>\Model\Entity\<%= $entityClass %>[]|\Cake\Collection\CollectionInterface $<%= $pluralVar %>
   */
 ?>
 <%
--- a/src/Shell/Task/TemplateTask.php
+++ b/src/Shell/Task/TemplateTask.php
@@ -266,6 +266,7 @@ class TemplateTask extends BakeTask
      *
      * - 'modelObject'
      * - 'modelClass'
+     * - 'entityClass'
      * - 'primaryKey'
      * - 'displayField'
      * - 'singularVar'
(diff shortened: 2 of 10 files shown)


dereuromark:
Thanks for looking into it :)

src/Template/Bake/Template/index.ctp
```diff
@@ -16,6 +16,7 @@
 <?php
 /**
   * @var \<%= $namespace %>\View\AppView $this
+  * @var \<%= $namespace %>\Model\Entity\<%= $entityClass %>[] $<%= $pluralVar %>
```
ndm2: `|\Cake\Collection\CollectionInterface` maybe, given that this is the default for baked controllers?

dereuromark: @ndm2 That is a useless annotation, though. You need the specific entity, otherwise all the properties are yellow (not findable), and so all relations with it.

See my ide helper plugin, and this here works right into that.

dereuromark: Maybe - like my ide helper - it should check if those entities actually exist, and if not fall back to some interface for entities or collections.
In this case you have still at least some annotations, even if you do not want to use concrete entities.

ndm2: @dereuromark  I'd argue that using collection methods has nothing to do with accessing entity properties. While the latter may be part of the further processing of collection method results, this shouldn't invalidate the ability to do the former without having the code being marked as erroneous.

dereuromark: Could both together be the solution to fullfill both sides?

ndm2: Sorry for the confusion, using both is what I ment to suggest... should have explicitly mentioned that instead of going the lazy way with the easy to overlook `|` separator.

dereuromark: I am sry, I did indeed overlook that pipe character in front of your suggestion.
So all good :)

dmromanov:
Added the `|\Cake\Collection\CollectionInterface` hint

dereuromark:
What do you think? Should we only add the specific Entity typehints if those actually exist when running the bake template generators? Otherwise those will be yellow.

dmromanov:
It's not something critical, but sure it would be nice to have.
I'll do it.

dereuromark:
OK, should we merge or wait for the conditional "entity exists" check?

PR #338, opened 2017-05-22
Add typehints only if Entity class exist

dmromanov:
In addition to a pull #337

### DIFF
--- a/src/Template/Bake/Template/index.ctp
+++ b/src/Template/Bake/Template/index.ctp
@@ -16,7 +16,7 @@
 <?php
 /**
   * @var \<%= $namespace %>\View\AppView $this
-  * @var \<%= $namespace %>\Model\Entity\<%= $entityClass %>[]|\Cake\Collection\CollectionInterface $<%= $pluralVar %>
+  * @var \<%= $entityClass %>[]|\Cake\Collection\CollectionInterface $<%= $pluralVar %>
   */
 ?>
 <%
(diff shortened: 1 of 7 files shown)


src/Template/Bake/Template/index.ctp
```diff
@@ -16,7 +16,9 @@
 <?php
 /**
   * @var \<%= $namespace %>\View\AppView $this
+<% if ($entityClass): %>
```
ndm2: Could not defining a view variable at all maybe be a little to hard of a cut? How about falling back to `\Cake\ORM\Entity`?

dereuromark: The entity interface makes total sense 👍 
Once the actual entity exists, the IdeHelper will replace the annotation then automatically anyway.

markstory: :+1: to falling back to EntityInterface.

tests/comparisons/Template/testBakeIndex.ctp
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
   * @var \Bake\Test\App\View\AppView $this
-  * @var \Bake\Test\App\Model\Entity\TemplateTaskComment[]|\Cake\Collection\CollectionInterface $templateTaskComments
+  * @var \Bake\Test\App\Model\Entity\Cake\Datasource\EntityInterface[]|\Cake\Collection\CollectionInterface $templateTaskComments
```
ndm2: That looks a little weird, the interface name shouldn't be prefixed with `\Bake\Test\App\Model\Entity`.

dereuromark: Oh you are right

dmromanov: Will fix.

PR #340, opened 2017-06-03
Add support for association target class annotations.

ndm2:
Alternative approach to #324.

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -278,13 +278,15 @@ class ModelTask extends BakeTask
     {
         $info = [];
 
+        $appNamespace = Configure::read('App.namespace');
+
         foreach ($table->associations() as $association) {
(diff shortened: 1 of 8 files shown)


src/Shell/Task/ModelTask.php
```diff
@@ -255,6 +257,57 @@ public function applyAssociations($model, $associations)
...
+     * @return array A map of association information.
+     */
+    public function getAssociationInfo(Table $table)
+    {
+        $info = [];
+
+        foreach ($table->associations() as $association) {
+            /* @var $association \Cake\ORM\Association */
+
+            $namespace = Configure::read('App.namespace');
```
markstory: This should be above the `foreach` it isn't going to change.

dereuromark:
Thanks a lot for looking into it! Now the combination bake + ide helper is officially 100% more awesome :)

Issue #341, opened 2017-06-03
Do not add a '@property Table' annotation if the table does not exist.


dmromanov:
A controller generated with `$ cake bake controller Foo` command should not contain the following annotation if `FooTable` class does not exist.
```
/**
 * @property App\Model\Table\FooTable $Foo
 */
```

dereuromark:
PR Open.

PR #342, opened 2017-06-03
Do not add a '@property Table' annotation if the table does not exist.

dmromanov:
#341


src/Shell/Task/ControllerTask.php
```diff
@@ -130,13 +130,19 @@ public function bake($controllerName)
         $singularName = $this->_singularName($currentModelName);
         $singularHumanName = $this->_singularHumanName($controllerName);
         $pluralHumanName = $this->_variableName($controllerName);
+
+        $defaultModel = sprintf('%s\Model\Table\%sTable', $namespace, $controllerName);
+        if (!class_exists($defaultModel)) {
+            $defaultModel = false;
```
dereuromark: null would suffice, right?

dereuromark:
Sounds reasonable.

PR #345, opened 2017-06-14
Index & view modified to comply w/ doc blocks.

sitedyno:
That code-sniffer fusses at me when I bake a new controller!

Command used:
`phpcs --colors -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests`

```
----------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
----------------------------------------------------------------------
 17 | WARNING | Function return type is not void, but function has no
    |         | return statement
 34 | WARNING | Function return type is not void, but function has no
    |         | return statement
----------------------------------------------------------------------
```

### DIFF
--- a/src/Template/Bake/Element/Controller/index.ctp
+++ b/src/Template/Bake/Element/Controller/index.ctp
@@ -17,7 +17,7 @@
     /**
      * Index method
      *
-     * @return \Cake\Http\Response|null
+     * @return \Cake\Http\Response|void
      */
     public function index()
     {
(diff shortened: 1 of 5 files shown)


dereuromark:
Hm thats not ideal.
Maybe the sniffer should be relaxed here?

markstory:
Or the docs could be Response|void.

sitedyno:
Ok. Changing doc block instead of adding return statement.

markstory:
Thanks!

src/Template/Bake/Element/Controller/index.ctp
```diff
@@ -17,7 +17,7 @@
     /**
      * Index method
      *
-     * @return \Cake\Http\Response|null
+     * @return \Cake\Http\Response|void
```
dereuromark: Thats semantically wrong.

Issue #348, opened 2017-07-12
Include MIT license by default for baked plugins


dereuromark:
We should add
https://github.com/cakephp/bake/blob/master/src/Template/Bake/Plugin/composer.json.ctp#L20

    license: MIT

by default
as most dont even know that there should be a license defined, and for most MIT suffices, for propriety and others they can just change it pretty quick.
This would prevent people from just by fun using AGPL or other not so suited default ones. 
Refs https://github.com/FriendsOfCake/awesome-cakephp/pull/271

markstory:
Sounds reasonable to me.

PR #349, opened 2017-07-13
Add MIT as default license

dereuromark:
Closes https://github.com/cakephp/bake/issues/348

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "cakephp/bake",
-    "description": "Bake plugin for CakePHP 3.0",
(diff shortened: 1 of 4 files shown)


markstory:
Looks like tests need to be updated.

ADmad:
Should a LICENSE.txt also be added?

inoas:
And why would we make that decision? What about "See License.txt"?

dereuromark:
I fixed the test. This should be mergable as is.
We can have a follow up PR with the license txt file or some other reference to what @ionas suggested.

PR #355, opened 2017-08-11
Allowing an overload of core shell tasks

AlPri78:
Hello,
I purpose this modification that I'm already using.
It allows to stay in the open way of cakephp by letting the possibility to rewrite the core tasks if needed, and I needed that.
I added tests to verify this simple modification (I just moved a line in loadtasks() function), and I also add a test to check an APP task (it wasn't checked). Of course, all BakeShell tests are ok.
What do you think about ?
Al


AlPri78:
I will repost this request better !

PR #359, opened 2017-09-04
Reinstate accessible field generation.

markstory:
This was removed in #134 to make adding new fields to the database easier for new developers. This created a tradeoff with more secure defaults. I'd like to re-add more secure defaults alongside warning log messages in the marshaller during debug mode to help new developers.


dereuromark:
Tests need updating, they use the old travis config for DB.

PR #361, opened 2017-09-15
fix detect Controller name

saeideng:
invalid result  when using 
`bin/cake bake test Controller Users --prefix=Admin`
but when using `controller` instead of  `Controller`
returned true result


markstory:
Thanks!

Issue #222, opened 2016-01-31
SkipFields enhancement


dereuromark:
See https://github.com/dereuromark/cakephp-setup/blob/master/docs/Console/Bake.md#skipping-fields
for current implementation

``` php
public $scaffoldSkipFields = array('password', ...); // For all actions
public $scaffoldSkipFieldsIndex = array('admin_comment'); // Only for index action
```

Ported from https://github.com/cakephp/cakephp/issues/1889

markstory:
Could we skip `_hidden` fields? Or is this different than that. Having two lists of hidden/skipped fields on 2 different classes is a bit confusing to me.

dereuromark:
hidden is different, I think.
Because this part of skip fields also mainly depends on the view type (form != view != index).

But I am fine keeping it in my custom template, if there is no interest in adapt.

markstory:
Its not that I'm not interested. I'm just trying to make it consistent with other features we have. Is the reason you don't want to use CLI flags that these settings should persist beyond a single bake run?

dereuromark:
Yes

A propos persisted. That is currently a big issue with using the shell in general.
All settings like theme and alike should also be persisted, using a Configure `Bake` key, so they are read into the shell as default settings on a per app basis.

markstory:
If there is a need/desire for persistent configuration in bake, I would prefer we do just that vs. adding properties to arbitrary classes.

dereuromark:
Well, we would have to add those new config options one way or another.
But most of those options should not be cli-enabled, as it is just super painful to manually append them to the call. Those should be internal config (either inside class as I did, or config file) only.

markstory:
A standalone config file feels like a better long term solution to me. We could also use this file to allow customisations to other tasks or enabling specific cli flags by default.

josegonzalez:
You could just use bake events for this.

othercorey:
Looks like this isn't getting support for now.

PR #365, opened 2017-10-12
Allow setting default theme through Configure.

ADmad:


dereuromark:
:+1: always bothered me :)

ADmad:
> always bothered me

Wish you had done something about it earlier and saved me the trouble :laughing:

dereuromark:
Haha, I cant take all the credit.
And technically I already did so 2 years ago for my own bake template afaik: https://github.com/cakephp/bake/issues/222

PS: I usually make the whole config array Configure-able by the way, but I guess this is the most pressing one, so maybe this suffices already for most.

ADmad:
> I usually make the whole config array Configure-able by the way..

Yeah, having the possibility in core itself would be nice but I too don't want to get into that right now.

markstory:
Nice work!

markstory:
Should the documentation also be updated for this?

ADmad:
I'll make a PR for the docs

ADmad:
Oh well, this change doesn't work in practice. I had assumed the defaults set for the shell would be passed to the `BakeTemplateTask`, but they don't.

While I can easily fix this issue by reading default through `Configure::read()` in `BakeTemplateTask` too, this is something we should consider fixing in general.

PR #366, opened 2017-10-13
Set default theme in BakeTask.

ADmad:
All other tasks inherit from BakeTask and get their default params from it.
Refs #365.

### DIFF
--- a/src/Shell/Task/BakeTask.php
+++ b/src/Shell/Task/BakeTask.php
@@ -250,6 +250,7 @@ class BakeTask extends Shell
         ])->addOption('theme', [
             'short' => 't',
             'help' => 'The theme to use when baking code.',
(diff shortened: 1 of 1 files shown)


tests/comparisons/BakeTemplate/testGenerateWithConfiguredTemplate.ctp
```diff
@@ -0,0 +1,9 @@
+<?php
+/**
+ * These are my routes. There are many like them but these are my own.
+ */
+use Cake\Routing\Router;
+
+Router::plugin('Special', function ($routes) {
```
saeideng: Router and `.ctp` file or `.php`!!?

ADmad: It's a template file hence ctp.

PR #376, opened 2017-11-12
else removed from controller/delete template

rrd108:
As I see other else statements were removed already, controller's delete method is the last what kept it.

### DIFF
--- a/tests/comparisons/Controller/testBakeActions.php
+++ b/tests/comparisons/Controller/testBakeActions.php
@@ -122,11 +122,10 @@ class BakeArticlesController extends AppController
     {
         $this->request->allowMethod(['post', 'delete']);
         $bakeArticle = $this->BakeArticles->get($id);
-        if ($this->BakeArticles->delete($bakeArticle)) {
-            $this->Flash->success(__('The bake article has been deleted.'));
-        } else {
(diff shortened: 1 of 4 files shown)


tests/comparisons/Controller/testBakeActions.php
```diff
@@ -122,11 +122,10 @@ public function delete($id = null)
     {
         $this->request->allowMethod(['post', 'delete']);
         $bakeArticle = $this->BakeArticles->get($id);
-        if ($this->BakeArticles->delete($bakeArticle)) {
-            $this->Flash->success(__('The bake article has been deleted.'));
-        } else {
+        if (!$this->BakeArticles->delete($bakeArticle)) {
```
dereuromark: That is behavior change here. Both messages are now triggered.

rrd108: Ah, you are right.

rrd108:
I am not sure that adding the redirect into the if would be ok from the viewpont of readibilty. With this in mind I think the original code with the else is better...

dereuromark:
Indeed - we already removed all reasonably wrong else cases. So in light of this, we can close.
Thank you, though, for looking into this!

Issue #377, opened 2017-11-22
Prefix duplicated for tests


dereuromark:
When running "cake bake all Posts" with `--prefix admin`
the tests somehow end up in 

    /tests/TestCase/Controller/Admin/Admin/PostsControllerTest.php

Expected would be

    /tests/TestCase/Controller/Admin/PostsControllerTest.php

PR #379, opened 2017-11-22
Remove duplicate prefix name being applied.

markstory:
In #361 the type check was fixed which resulted in prefixed controllers having their prefixes doubled when tests were generated. By removing the code in ControllerTask tests generate correctly again.

Refs #377


dereuromark:
Looks good.

PR #380, opened 2017-11-23
Corrected fieldValidation annotation

dmromanov:


src/Shell/Task/ModelTask.php
```diff
@@ -691,14 +691,14 @@ public function getValidation($model, $associations = [])
      * @param \Cake\Database\Schema\TableSchema $schema The table schema for the current field.
      * @param string $fieldName Name of field to be validated.
      * @param array $metaData metadata for field
-     * @param string $primaryKey The primary key field
+     * @param array $primaryKey The primary key field
```
dereuromark: Cant primary key also be a string?
Maybe we should do string|array and cast inside the in_array() call?

dereuromark: I see the cast is happening before calling this method: `$primaryKey = (array)$schema->primaryKey();`
Then you can just add the array typehint in the signature IMO.

dmromanov: Done.

src/Shell/Task/ModelTask.php
```diff
@@ -691,14 +691,14 @@ public function getValidation($model, $associations = [])
...
      * @return array Array of validation for the field.
      */
-    public function fieldValidation($schema, $fieldName, array $metaData, $primaryKey)
+    public function fieldValidation($schema, $fieldName, array $metaData, array $primaryKey)
```
ADmad: I don't think this is a backwards compatible change. Type hint change for `$primaryKey` var will cause error if method is overridden in an extending class with old signature.

dereuromark: In this case this is probably fair enough as it was always internally expected to be an array (or crash).
See code. This just puts the error more clearly to the beginning of the method.

ADmad: Even if one is correctly passing an array he would still get an error due to method signature mismatch in child class, which shouldn't happen.

PR #382, opened 2017-11-24
Remove argument typehint.

ADmad:
This ensures method signature's backwards compatibility. Refs #380


dereuromark:
this was a bad change for the reasons i stated in the ticket.

PR #383, opened 2017-11-25
Fix array expectation via cast.

dereuromark:
Reverts cakephp/bake#382

as I explained already to @ADmad in the comment to the original and valid change to add the typehint, the internal code expects the array, thus this revert here was invalid and a bad idea.
BC has nothing to do with this :)

See if (in_array($fieldName, $primaryKey)) { usage.

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -694,7 +694,7 @@ class ModelTask extends BakeTask
      * @param array $primaryKey The primary key field
      * @return array Array of validation for the field.
      */
-    public function fieldValidation($schema, $fieldName, array $metaData, $primaryKey)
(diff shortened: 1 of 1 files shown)


ADmad:
And I already explained to you why maintaining the same signature is necessary to avoid PHP complaining from about signature mismatch, in an existing userland class which extends ModelTask and has overridden this method.

Hopefully @markstory or someone else can explain the same to you again.

dereuromark:
I think you are a bit pedantic here on the bc issue which is highly theoretical.
The actual issue remains..

dereuromark:
I adjusted to code to ahere to the BC issue (which is really not that big of deal IMO on that require-dev plugin).
But I would prefer if this was fixed properly instead of just reverting code that was put there for a reason :)

PR #384, opened 2017-11-27
Made generated templates more easier to localize

martonmiklos:
Added labels containing the humanized field names to the generated templates:
- To the add/edit form's inputs
- To the paginator header links in the index

As the displayed strings in the generated contents are localizable I think the fields names should be localizable as well.

### DIFF
--- a/src/Template/Bake/Template/index.twig
+++ b/src/Template/Bake/Template/index.twig
@@ -42,7 +42,7 @@
         <thead>
             <tr>
 {% for field in fields %}
-                <th scope="col"><?= $this->Paginator->sort('{{ field }}') ?></th>
+                <th scope="col"><?= $this->Paginator->sort('{{ field }}', ['label' => __('{{ fieldHumanNames[field] }}')]) ?></th>
 {% endfor %}
                 <th scope="col" class="actions"><?= __('Actions') ?></th>
             </tr>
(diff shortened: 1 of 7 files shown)


src/Template/Bake/Template/index.twig
```diff
@@ -42,7 +42,7 @@
         <thead>
             <tr>
 {% for field in fields %}
-                <th scope="col"><?= $this->Paginator->sort('{{ field }}') ?></th>
+                <th scope="col"><?= $this->Paginator->sort('{{ field }}', ['label' => __('{{ fieldHumanNames[field] }}')]) ?></th>
```
markstory: This method doesn't have a label option in the second argument. See [the API docs](https://api.cakephp.org/3.5/class-Cake.View.Helper.PaginatorHelper.html#_sort).

martonmiklos: Hmm. But it seems to be working:
kep

What do you recommend? Is this an undocumented feature, which might change in the future?

dereuromark:
You might not know: Everything you changed is already localized. 
You just added extra verbosity. Most would find that unnecessary.

martonmiklos:
Oh. I have just realised it so I will close the PR. What do you think about the documentation of the label parameter of the PaginatorHelper::sort?
I think it would be useful to update the docs because having a different sorter label than the humanized field name is not so uncommon need.

PR #387, opened 2017-12-01
Run phpstan checks.

ADmad:


src/Shell/Task/ModelTask.php
```diff
@@ -1031,14 +1031,14 @@ protected function _getAllTables()
                 'Connections need to implement schemaCollection() to be used with bake.'
             );
 
-            return $this->_stop();
+            $this->_stop();
```
markstory: You might need this return for when stop is mocked. You could also use abort() instead.

ADmad:
@ondrejmirtes Is there a better way to deal with [this](https://travis-ci.org/cakephp/bake/jobs/310083150#L603) error other than just ignoring it?  The issue is caused by use of [@property](https://github.com/cakephp/bake/blob/de561b9674e74c30b6712b45408b90da16d6b838/src/View/Helper/DocBlockHelper.php#L76) tag within description text.

ondrejmirtes:
@ADmad Your problem is not unique, I encountered it in at least three other places. We'll try to ignore phpDocs in the middle of text descriptions, but for now, try to construct as specific regex as possible to ignore it (so that you don't ignore other valid errors).

ADmad:
This is good to go.

src/Shell/Task/ModelTask.php
```diff
@@ -1027,18 +1027,14 @@ protected function _getAllTables()
     {
         $db = ConnectionManager::get($this->connection);
         if (!method_exists($db, 'schemaCollection')) {
-            $this->err(
+            $this->abort(
                 'Connections need to implement schemaCollection() to be used with bake.'
             );
-
-            return $this->_stop();
         }
```
saeideng: > Connections **need to** implement schemaCollection() to be used with bake. 

and continue to use `$db->schemaCollection();`  :)

src/Shell/Task/ModelTask.php
```diff
@@ -1027,18 +1027,14 @@ protected function _getAllTables()
...
         $schema = $db->schemaCollection();
         $tables = $schema->listTables();
         if (empty($tables)) {
-            $this->err('Your database does not have any tables.');
-
-            return $this->_stop();
+            $this->abort('Your database does not have any tables.');
```
saeideng: and here ...

PR #388, opened 2017-12-01
Get FQCN instead of assuming it's a core behavior.

ADmad:


dereuromark:
Looks good.

markstory:
Looks good. Tests will need to be updated though.

ADmad:
Tests are fine, the code needed to be tweaked. My initial patch failed due to the fact that core behaviors are under `ORM/Behavior` instead of `Model/Behavior`.

This makes me wonder if this difference in namespace should be handled inside `App::className()` itself instead of [Behavior::_resolveClassName()](https://github.com/cakephp/cakephp/blob/d7cf3e7a1dde15ed9516b9932f6c436b4cd973ec/src/ORM/BehaviorRegistry.php#L92).

markstory:
@ADmad `App::className()` is already pretty complex, perhaps behavior class resolution could have a public method on the BehaviorRegistry?

ADmad:
> perhaps behavior class resolution could have a public method on the BehaviorRegistry?

And be static too? Wouldn't want to instantiate registry just to generate classname.

markstory:
A static method should be ok.

PR #389, opened 2017-12-02
Use helper to generate behavior config array.

ADmad:


tests/comparisons/Model/testGetBehaviors.php
```diff
@@ -0,0 +1,55 @@
...
+ * @method \App\Model\Entity\Post patchEntity(\Cake\Datasource\EntityInterface $entity, array $data, array $options = [])
+ * @method \App\Model\Entity\Post[] patchEntities($entities, array $data, array $options = [])
+ * @method \App\Model\Entity\Post findOrCreate($search, callable $callback = null, $options = [])
+ *
+ * @mixin \Cake\ORM\Behavior\CounterCacheBehavior
+ * @mixin \Cake\ORM\Behavior\TranslateBehavior
+ */
+class PostsTable extends Table
+{
+
```
dereuromark: We should remove that newline if possible (as per psr2). Or the generated code will always have to be sniffer fixed afterwards.

ADmad: Our coding standard does not report any error for this new line.

ADmad: I checked, neither does PSR2.

dereuromark: But it should, it states that there is no newline at beginning or end.

src/View/Helper/BakeHelper.php
```diff
@@ -78,7 +78,19 @@ public function stringifyList(array $list, array $options = [])
...
             if (!is_numeric($k)) {
-                $v = "'$k' => $v";
+                $nestedOptions = $options;
+                if ($nestedOptions['indent']) {
+                    $nestedOptions['indent'] += 1;
+                }
+                if (is_array($v)) {
+                    $v = sprintf(
+                        "'%s' => [%s]",
+                        $k,
```
markstory: Do we need to escape possible `'` in `$k`?

ADmad: I don't see anyone using single quotes in config key names.

markstory: Ok.

PR #390, opened 2017-12-08
Remove _serialize from templates

cleptric:
cc @chinpei215

### DIFF
--- a/src/Template/Bake/Element/Controller/add.twig
+++ b/src/Template/Bake/Element/Controller/add.twig
@@ -43,5 +43,4 @@
     {%- set compact = compact|merge(["'#{otherPlural}'"]) %}
 {% endfor %}
         $this->set(compact({{ compact|join(', ')|raw }}));
-        $this->set('_serialize', ['{{ singularName }}']);
     }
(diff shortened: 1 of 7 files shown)


markstory:
Is this necessary with the auto serialize removed from the app skeleton?

chinpei215:
I think it is necessary. I think people would easy to forget to remove this code from the baked controller. For instance, our official site has not removed the code: https://github.com/cakephp/cakephp.org/blob/e4471d274571e4c9fc9581816077cb54063609e8/src/Controller/Admin/ProjectsController.php#L34
I don't think it is intended to return JSON/XML from the action. And it could be a security hole if we had authorization based on roles.

chinpei215:
We should recognize this as a potential security risk. In fact, we have already recommended to remove all` _serialize` from actions. So first, I think we should remove `_serialize` as soon as possible. And if you have some ideas about the new option, we can add it and its documentation.

rchavik:
i agree with@chinpei215

markstory:
@rchavik @chinpei215 Ok.

Issue #392, opened 2017-12-18
bake all --everything fails with non-default connection


ewbarnard:
There is a problem with the order of operations for "bake all --everything --connection xxx". In cakephp/bake/src/Shell/Task/ModelTask.php function bake() (around line 109), we bake the table, entity, fixture, and test.

The problem is that when baking the test, we go off looking for associations. The associations can be for models which have not yet been baked. Therefore at

cakephp/cakephp/src/ORM/Locator/TableLocator.php line 210:

`$connectionName = $className::defaultConnectionName();`

returns the default connection, because that model class ($className) has not yet been baked.

I can work around the issue by not baking tests/fixtures on the non-default connection. In ModelTask::bake(), between bakeEntity() and bakeFixture() (line 116), I can insert the following workaround:

```
if (array_key_exists('connection', $this->params) && ($this->params['connection'] !== 'default')) {
  // Associated models may not yet be baked so their default connection will be unknown
  return;
}
```

This is related to #250 but it's not the same issue. I suggest disallowing automatic generation of tests for "bake all everything" on the non-default connection. It's an edge case not worth taking up a lot of time. To do that, insert something like the above code but adding a check for "everything"; add a notice to BakeShell::all() near line 253 that tests/fixtures will not be generated; update documentation for bake all everything.

markstory:
>  I suggest disallowing automatic generation of tests for "bake all everything" on the non-default connection. It's an edge case not worth taking up a lot of time. To do that, insert something like the above code but adding a check for "everything"; add a notice to BakeShell::all() near line 253 that tests/fixtures will not be generated; update documentation for bake all everything.

I think this is a good approach. With alternate connections is is hard to know the scope of 'everything'.

ewbarnard:
Mark, I'll see if I can get a PR together by the end of this week.

saeideng:
a PR merged about this issue

PR #393, opened 2017-12-20
#392 Bake all --everything connection

ewbarnard:
When baking all --everything on the non-default database connection,
bypass baking the unit tests. When baking the unit tests, the code
generator searches for model associations. Since the associated models
may not yet be baked, the generator fails when trying to find the
associated table on the default connection.

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -113,6 +113,15 @@ class ModelTask extends BakeTask
         $data = $this->getTableContext($tableObject, $table, $name);
         $this->bakeTable($tableObject, $data);
         $this->bakeEntity($tableObject, $data);
+        if (array_key_exists('connection', $this->params) &&
+            array_key_exists('everything', $this->params) &&
+            ($this->params['connection'] !== 'default')) {
+            // Cannot bake tests on 'bake all --everything' on non-default
+            // connection because associated models may not yet be baked,
+            // causing us to look for the table on the default connection
+            $this->out('Not baking tests or test fixtures');
+            return;
+        }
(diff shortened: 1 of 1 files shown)


src/Shell/Task/ModelTask.php
```diff
@@ -113,6 +113,16 @@ public function bake($name)
         $data = $this->getTableContext($tableObject, $table, $name);
         $this->bakeTable($tableObject, $data);
         $this->bakeEntity($tableObject, $data);
+        if (array_key_exists('connection', $this->params) &&
```
dereuromark: Why do we need array_key_exists?
Shouldnt

    if ($this->param('connection') && ...)

suffice?

dereuromark:
ping @ewbarnard

src/Shell/Task/ModelTask.php
ewbarnard: @dereuromark Mark you are correct. When I wrote that code I was unaware that $this->param() could be used as you propose. array_key_exists() should be changed. I'm on the road at present but can edit the PR by Monday.

ewbarnard: @dereuromark Mark, thanks for the ping. I see I missed that we need test coverage here. I'll see what I can do on that, but will be at least Monday 2018-08-20 before I can get to it.

dereuromark:
@markstory Do we finish this or close?

ewbarnard:
Sorry, forgot about this. Looks like 2 things are needed - new tests, and remove the unnecessary array_key_exists(). I can hopefully complete this tonight.

ewbarnard:
Ready for re-consideration. 
@dereuromark @markstory

PR #394, opened 2017-12-21
Exclude columns of primary key from "belongsTo"

tporadowski:
When a database table contains a column that ends with "_id" by convention it is treated as a foreign key. I stumbled upon a schema where regular primary keys were named in this manner (i.e. "usr_id" was primary key of "users" table, etc.), which of course misled bake. This change introduces a check to skip such fields from being treated as associations when they are set as primary key (but only when PK is a single column).

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -323,7 +323,7 @@ class ModelTask extends BakeTask
     {
         $schema = $model->getSchema();
         foreach ($schema->columns() as $fieldName) {
-            if (!preg_match('/^.+_id$/', $fieldName) || in_array($fieldName, $schema->primaryKey())) {
+            if (!preg_match('/^.+_id$/', $fieldName) || ([$fieldName] == $schema->primaryKey())) {
                 continue;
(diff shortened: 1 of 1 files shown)


src/Shell/Task/ModelTask.php
```diff
@@ -323,7 +323,7 @@ public function findBelongsTo($model, array $associations)
     {
         $schema = $model->getSchema();
         foreach ($schema->columns() as $fieldName) {
-            if (!preg_match('/^.+_id$/', $fieldName)) {
+            if (!preg_match('/^.+_id$/', $fieldName) || ([$fieldName] == $schema->primaryKey())) {
```
dereuromark: Please use `===`

tporadowski: In this case I suppose it was fine (arrays of strings), but I've just changed it as you asked. I will keep this in mind for any future PRs, thanks!

markstory:
@dereuromark I've added a test before merging in fa49e714a9309672883fb065f2142df8e588e202.

PR #397, opened 2017-12-25
Allow generating validation rules with boolean or array args.

ADmad:


src/Shell/Task/ModelTask.php
```diff
@@ -732,11 +732,11 @@ public function fieldValidation($schema, $fieldName, array $metaData, $primaryKe
         }
 
         if (in_array($fieldName, (array)$primaryKey)) {
-            $rules['allowEmpty'] = ['create'];
+            $rules['allowEmpty'] = ["'create'"];
         } elseif ($metaData['null'] === true) {
             $rules['allowEmpty'] = [];
         } else {
-            $rules['requirePresence'] = ['create'];
+            $rules['requirePresence'] = ["'create'"];
```
markstory: Could this potentially result in working templates no longer working as they require quotes now?

ADmad: Yes it potentially can if someone has overridden this method.

So we can document this change and bump bake version accordingly.

markstory: We'd have to go to 1.6 I think.

ADmad: Yup, fine by me.

PR #399, opened 2017-12-26
Using \Cake\Http\ServerRequest|Response instead of \Cake\Network\*

gotoeveryone:
`bake` task is using the following.

- `\Cake\Network\Request`
- `\Cake\Network\Response`

Changed it to use the following.

- `\Cake\Http\ServerRequest`
- `\Cake\Http\Response`


dereuromark:
Looks good.

PR #400, opened 2017-12-27
Fixes baking controller tests with prefix option

nojimage:
A test case is generated in the incorrect namespace when the `prefix` option is specified when baking the controller.

eg:

```
bin/cake bake controller --prefix Api/Public Posts
```

I expected:

tests/TestCase/Controller/Api/Public/PostsControllerTest.php

```
<?php
namespace App\Test\TestCase\Controller\Api\Public;
```

Actually I got:

```
<?php
namespace App\Test\TestCase\Controller\Api/Public;
```

### DIFF
--- a/tests/TestCase/Shell/Task/TestTaskTest.php
+++ b/tests/TestCase/Shell/Task/TestTaskTest.php
@@ -363,6 +363,19 @@ class TestTaskTest extends TestCase
     }
 
     /**
+     * test resolving class names with prefix
+     *
+     * @return void
+     */
+    public function testGetRealClassnamePrefix()
+    {
+        $this->Task->params['prefix'] = 'Api/Public';
+        $result = $this->Task->getRealClassname('Controller', 'Posts');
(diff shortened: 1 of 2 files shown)


ADmad:
This is a backwards incompatible change. The prefix is supposed to be specified in lower case not CamelCase. So to fix the problem you should check if string contains slash and split the parts camelize and join back.

nojimage:
I change to use `_getPrefix()` method. This method return prefix option that split slashes and camelized.

tests/TestCase/Shell/Task/TestTaskTest.php
```diff
@@ -362,6 +362,19 @@ public function testGetRealClassnamePlugin()
...
     }
 
+    /**
+     * test resolving class names with prefix
+     *
+     * @return void
+     */
+    public function testGetRealClassnamePrefix()
+    {
+        $this->Task->params['prefix'] = 'Api/Public';
```
ADmad: The test needs to be done using lowercase `api/public`.

PR #402, opened 2018-01-01
Fix up non interactive mode using -q quit and -f force params.

dereuromark:
Following https://github.com/cakephp/cakephp/pull/11574 this fixes up the non interactive mode.

Currently -q would prompt for sth but you are not able to see for what. So it just endless runs.

It now properly
- uses -q (quit) as non-interactive mode and silently ignores all existent files (no accidental overwrite anymore).
- uses -f (force) to force overwrite in -q non-interactive mode as specific in help.

This should use 3.5.10 release as before it always silently overwrote the files. Thus the composer constraint.

### DIFF
--- a/src/Shell/BakeShell.php
+++ b/src/Shell/BakeShell.php
@@ -73,6 +73,13 @@ class BakeShell extends Shell
         if (isset($this->params['connection'])) {
             $this->connection = $this->params['connection'];
         }
+
+        if ($this->params['quiet']) {
+            $this->interactive = false;
+            if (isset($this->{$task}) && !in_array($task, ['Project'])) {
+                $this->{$task}->interactive = false;
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Shell/BakeShell.php
```diff
@@ -73,6 +73,13 @@ public function startup()
         if (isset($this->params['connection'])) {
             $this->connection = $this->params['connection'];
         }
+
+        if ($this->params['quiet']) {
+            $this->interactive = false;
+            if (isset($this->{$task}) && !in_array($task, ['Project'])) {
```
markstory: Why are you using `in_array()` with a single element? Couldn't that check be done with `$task !== 'Project'`

dereuromark: Copy and paste from the similar code above.
I thought this must have a reason (extendability) to be this way there.

dereuromark:
We could add a test that proofs it works to e.g. `cake bake all Foos`.

dereuromark:
Is this added test all right?

Issue #235, opened 2016-04-01
missing --prefix option for bake cell


saeideng:
https://github.com/cakephp/cakephp/pull/8541

LubosRemplik:
We can close this now :)

PR #408, opened 2018-01-25
Prefix option for bake cell

LubosRemplik:
I like prefixes for code I don't want to decouple into plugin. I am missing this option in some bake tasks, currently in the cell so I have tried to add it there.

Maybe it would be good idea to implement prefix to ` src/Shell/Task/SimpleBakeTask.php` so all Tasks which extends this class can have this prefix option, but that's too much for me at the moment

Hope all OK

Reference #235

### DIFF
--- a/src/Shell/Task/CellTask.php
+++ b/src/Shell/Task/CellTask.php
@@ -14,6 +14,8 @@
  */
 namespace Bake\Shell\Task;
 
+use Cake\Core\Configure;
+
 /**
  * Task for creating cells.
  *
@@ -54,6 +56,26 @@ class CellTask extends SimpleBakeTask
     }
 
     /**
+     * Get template data.
+     *
+     * @return array
+     */
+    public function templateData()
+    {
+        $prefix = $this->_getPrefix();
+        if ($prefix) {
+            $prefix = '\\' . str_replace('/', '\\', $prefix);
+        }
+
+        $namespace = Configure::read('App.namespace');
+        if ($this->plugin) {
+            $namespace = $this->_pluginNamespace($this->plugin);
+        }
+
+        return compact('namespace', 'prefix');
+    }
+
+    /**
      * Bake the Cell class and template file.
      *
      * @param string $name The name of the cell to make.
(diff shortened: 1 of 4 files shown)


src/Shell/Task/CellTask.php
```diff
@@ -74,13 +96,30 @@ public function bake($name)
...
+     * Gets the option parser instance and configures it.
+     *
+     * @return \Cake\Console\ConsoleOptionParser
+     */
+    public function getOptionParser()
+    {
+        $parser = parent::getOptionParser();
+        $parser
+        ->addOption('prefix', [
+            'help' => 'The namespace/routing prefix to use.'
```
markstory: Is `prefix` intended to be connected to a routing prefixes here?

tests/TestCase/Shell/Task/CellTaskTest.php
```diff
@@ -97,4 +97,45 @@ public function testBakePlugin()
...
+        $this->generatedFiles = [
+            APP . 'View/Cell/Admin/ExampleCell.php',
+            ROOT . 'tests/TestCase/View/Cell/Admin/ExampleCellTest.php',
+            APP . 'Template/Cell/Admin/Example/display.ctp',
+        ];
+        $this->exec('bake cell --prefix Admin Example');
+
+        $this->assertExitCode(Shell::CODE_SUCCESS);
+        $this->assertFilesExist($this->generatedFiles);
+        $this->assertFileContains('class ExampleCell extends Cell', $this->generatedFiles[0]);
```
markstory: You should also check that the classes namespace is correct.

src/Shell/Task/CellTask.php
LubosRemplik: It is not, I copied from different class. Fixed

tests/TestCase/Shell/Task/CellTaskTest.php
LubosRemplik: I have added assert with namespace

markstory:
Thanks!

PR #418, opened 2018-02-02
Adding getOptionParser method to the shell task

### LINKED ISSUE #416: Bake Task getOptionParser
When using `bin/cake bake task MyTask` it generates class with `main` method.  Shouldn't we generate `getOptionParser` by default too? It would be useful even when not used, but usually you add some options to the task.

If agreed, will create PR

LubosRemplik:
Closes #416

### DIFF
--- a/src/Template/Bake/Shell/task.twig
+++ b/src/Template/Bake/Shell/task.twig
@@ -27,7 +27,6 @@ class {{ name }}Task extends Shell
      * Manage the available sub-commands along with their arguments and help
      *
      * @see http://book.cakephp.org/3.0/en/console-and-shells.html#configuring-options-and-generating-help
-     *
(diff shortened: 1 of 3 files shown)


markstory:
You will need to update the tests as well

src/Template/Bake/Shell/task.twig
```diff
@@ -23,6 +23,20 @@ use Cake\Console\Shell;
 class {{ name }}Task extends Shell
 {
 
+    /**
+     * Manage the available sub-commands along with their arguments and help
+     *
+     * @see http://book.cakephp.org/3.0/en/console-and-shells.html#configuring-options-and-generating-help
+     *
```
ADmad: Extra blank line.

PR #421, opened 2018-02-10
baked shell test case superclass to ConsoleIntegrationTestCase

tenkoma:
Follow https://book.cakephp.org/3.0/en/development/testing.html#console-integration-testing

### DIFF
--- a/src/Template/Bake/tests/test_case.twig
+++ b/src/Template/Bake/tests/test_case.twig
@@ -18,13 +18,13 @@
 {% set isController = type|lower == 'controller' %}
 {% set isShell = type|lower == 'shell' %}
 {% if isController %}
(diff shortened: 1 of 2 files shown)


src/Template/Bake/tests/test_case.twig
```diff
@@ -16,11 +16,15 @@
  */
 #}
 {% set isController = type|lower == 'controller' %}
+{% set isShell = type|lower == 'shell' %}
 {% if isController %}
-    {%- set uses = uses|merge(['Cake\\TestSuite\\IntegrationTestCase']) %}
+    {%- set superclassName = 'IntegrationTestCase' %}
```
dereuromark: should this be superClassName maybe?

tenkoma: okey

PR #428, opened 2018-03-01
Do not generate fixture property recursively to avoid too big array

### LINKED ISSUE #427: Model TestCase fixtures
When you bake Model Table Test, it bakes `$fixtures` property with all possible fixtures depending on associations. Sometimes there are too many and I think by default it could bake just fixture for the table itself, for example for `PostsTableTest` it would bake 

```
public $fixtures = [
    'app.posts'
];
```

I am thinking about this because once I bake something, I have to remove fixtures otherwise phpunit test fails 

What do you think?

LubosRemplik:
Closes #427

### DIFF
--- a/src/Shell/Task/TestTask.php
+++ b/src/Shell/Task/TestTask.php
@@ -431,8 +431,7 @@ class TestTask extends BakeTask
     }
 
     /**
-     * Process a model recursively and pull out all the
-     * model names converting them to fixture names.
+     * Process a model, pull out model name + associations converted to fixture names.
      *
      * @param \Cake\ORM\Table $subject A Model class to scan for associations and pull fixtures off of.
      * @return void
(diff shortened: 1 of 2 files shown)


LubosRemplik:
Any hint how to fix PHPSTAN failure?

dereuromark:
Remove that line from neon file (not matched anymore since you deleted the code for it).

LubosRemplik:
Got it, thanks :)

tests/TestCase/Shell/Task/TestTaskTest.php
```diff
@@ -280,22 +280,6 @@ public function testFixtureArrayGenerationFromModel()
...
-     *
-     * @return void
-     */
-    public function testFixtureArrayGenerationIgnoreSelfAssociation()
-    {
-        TableRegistry::clear();
-        $subject = new CategoryThreadsTable();
-        $result = $this->Task->generateFixtureList($subject);
-        $expected = [
-            'app.category_threads',
```
markstory: Does this mean that self associations will attempt to include the child/parent fixtures?

src/Shell/Task/TestTask.php
```diff
@@ -447,20 +446,9 @@ protected function _processModel($subject)
             $assoc = $subject->association($alias);
             $target = $assoc->getTarget();
             $name = $target->getAlias();
-            $subjectClass = get_class($subject);
-
-            if ($subjectClass !== 'Cake\ORM\Table' && $subjectClass === get_class($target)) {
```
dereuromark: This line should probably stay.

LubosRemplik:
@dereuromark & @markstory sorry guys, I thought I am removing test for recursively added fixtures. The code & test which checks self association is back

Example of generated array from my test app for UsersTable
```
    /**
     * Fixtures
     *
     * @var array
     */
    public $fixtures = [
        'app.users',
        'app.tagged',
        'app.tags',
        'app.tokens',
        'app.roles'
    ];
```

Which means it adds fixture of table itself + associations (including the ones from behaviors), but should not add fixtures recursively any more

Issue #425, opened 2018-02-22
Expression is not allowed as field default value


davidyell:
# What I did
Installed bake `1.6.2` into my project and generated a migration against my Packages table.

```bash
$ bin/cake bake fixture -n 1 -r Packages
```

# What happened
My fixture was baked successfully.

# What's wrong?
I opened my fixture in PHP Storm and was presented with the following. Which PHP Storm says is `Expression is not allowed as field default value`

![PHP Storm error](https://i.imgur.com/gTmxluU.png)

When I remove the call to `Cake\I18n\FrozenTime::__set_state()` the errors go away and I can run my tests.

If you try and execute a test with a baked fixture, the test throws a fatal.

```bash
PHP Fatal error:  Constant expression contains invalid operations in /tests/Fixture/PackagesFixture.php on line 52
```

davidyell:
If it adds anything, the `exit_links` dimension of the data is created using contain in a plugin behaviour. As it's a dynamic attachment.

markstory:
Looks like we should be converting those timestamps into string values.

dereuromark:
> I'm wondering if another solution would just be to have Bake set $records in init() instead of when defining the property

That sounds like a great idea, would also allow other more "runtime/dynamic" things out of the box.
The hardcoded property is a bit outdated these days.

markstory:
I'm on-board if we want to start generating new fixtures to use the method based approach. Is anyone interested in putting a pull request together for that?

ethanpooley:
This is basically what we did, but I don't feel at all qualified to say what the repercussions might be for widespread adoption. If this is pretty much it I can certainly PR it for us.

![screenshot from 2018-03-08 17-11-24](https://user-images.githubusercontent.com/856884/37185162-de317fca-22f3-11e8-8724-e9ebfbd1a626.png)

dereuromark:
PR open.

PR #429, opened 2018-03-09
Assign fixture records at run time rather than compile time

ethanpooley:
This PR changes Bake fixture templates so that assignment of a fixture's records happens at run time via an `init()` method, rather than at compile time via a property declaration. The limitations imposed by the current implementation can cause problems for complex data types. See https://github.com/cakephp/bake/issues/425

### DIFF
--- a/tests/comparisons/Fixture/testImportRecordsFromDatabase.php
+++ b/tests/comparisons/Fixture/testImportRecordsFromDatabase.php
@@ -18,11 +18,13 @@ class UsersFixture extends TestFixture
     public $import = ['table' => 'users', 'connection' => 'test'];
 
     /**
-     * Records
+     * Init method
      *
-     * @var array
+     * @return void
      */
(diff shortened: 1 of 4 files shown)


tests/comparisons/Fixture/testImportRecordsFromDatabase.php
```diff
@@ -52,4 +54,6 @@ class UsersFixture extends TestFixture
             'updated' => '2012-06-12 01:24:31'
         ],
     ];
```
ethanpooley: The failed test is due to inadequate indentation of the records here. If anyone has a hint on the proper/best way to address this, I'm all ears.

dereuromark: You need one level more, indeed.
Should be doable on the part where it generates it.

markstory:
Thanks!

PR #431, opened 2018-03-20
Add plugin class creation when baking a plugin.

markstory:
Create the stub plugin class when creating a new plugin. I needed to fix the template name mangling to ensure that Plugin.php is generated correctly.


src/Template/Bake/Plugin/src/Plugin.php.twig
```diff
@@ -0,0 +1,28 @@
...
+namespace {{ namespace }};
+
+use Cake\Core\BasePlugin;
+
+/**
+ * Plugin for {{ namespace }}
+ */
+class Plugin extends BasePlugin
+{
+
```
dereuromark: This newline shouldnt be there.

PR #433, opened 2018-04-20
TableSchema::index() is deprecated. Use TableSchema::getIndex() instead.

ava007:
Error Location:  Shell/Task/FixtureTask.php, line: 276

bake version 1.7.0


markstory:
Thank you!

Issue #437, opened 2018-04-25
bake ShellHelper inflection issue


dereuromark:
```
$ bin/cake bake test
```
You must provide a class type to bake a test for. The valid types are:
```
1. Entity
2. Table
3. Controller
4. Component
5. Behavior
6. Helper
7. Shell
8. Task
9. Shell_helper
10. Cell
11. Form
12. Mailer
```
and error `Error: "ShellHelper" is not a valid value for type`

As can be seen, the expected `ShellHelper` as input does not work, which is weird, as all other types are CamelCased from the looks of it.

PR #438, opened 2018-04-25
Fix inflection in TestTask to properly normalize first for consistent…

dereuromark:
… internal handling.

Resolves https://github.com/cakephp/bake/issues/437 and makes code way simpler and less error prone.

I recommend 1.8.0 release.

### DIFF
--- a/tests/TestCase/Shell/Task/TestTaskTest.php
+++ b/tests/TestCase/Shell/Task/TestTaskTest.php
@@ -334,23 +334,23 @@ class TestTaskTest extends TestCase
     {
         return [
             ['Entity', 'Article', 'App\Model\Entity\Article'],
-            ['entity', 'ArticleEntity', 'App\Model\Entity\ArticleEntity'],
+            ['Entity', 'ArticleEntity', 'App\Model\Entity\ArticleEntity'],
             ['Table', 'Posts', 'App\Model\Table\PostsTable'],
-            ['table', 'PostsTable', 'App\Model\Table\PostsTable'],
+            ['Table', 'PostsTable', 'App\Model\Table\PostsTable'],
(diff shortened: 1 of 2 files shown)


tests/TestCase/Shell/Task/TestTaskTest.php
```diff
@@ -334,23 +334,23 @@ public static function realClassProvider()
...
+            ['Component', 'AuthComponent', 'App\Controller\Component\AuthComponent'],
             ['Shell', 'Example', 'App\Shell\ExampleShell'],
-            ['shell', 'Example', 'App\Shell\ExampleShell'],
+            ['Shell', 'Example', 'App\Shell\ExampleShell'],
+            ['Task', 'Example', 'App\Shell\Task\ExampleTask'],
             ['Task', 'Example', 'App\Shell\Task\ExampleTask'],
-            ['task', 'Example', 'App\Shell\Task\ExampleTask'],
             ['Cell', 'Example', 'App\View\Cell\ExampleCell'],
-            ['cell', 'Example', 'App\View\Cell\ExampleCell'],
+            ['Cell', 'Example', 'App\View\Cell\ExampleCell'],
```
markstory: Looks like we have redundant tests now?

dereuromark: Fixed up and added some more tests.

ADmad:
This change was published in bugfix release 1.7.3 causing failures in community plugins which relied on this particular casing. Example:

https://github.com/HavokInspiration/wrench/blob/master/src/Shell/Task/MaintenanceModeTask.php#L59

markstory:
We can work towards not having public properties in the future.

PR #450, opened 2018-05-25
Updated bake files to use test traits

jeremyharris:
Follows https://github.com/cakephp/cakephp/pull/12072


markstory:
Thank you 👍

PR #453, opened 2018-06-26
Fix CommandTest setup

CauanCabral:
Now that bake task generate below code:

```php
    public function setUp()
    {
        parent::setUp();
        $this->$this->useCommandRunner();
    }
```

As `TestTake` provide _construct_ value with **$this**:

```
$construct = "\$this->useCommandRunner();";
```

Theres no reason to repeat **$this** here.


markstory:
Thank you :clap:

PR #454, opened 2018-07-01
load WyriHaximus/TwigView Plugin through new way

saeideng:


saeideng:
now it works with both way 
when user uses `Plugin::load('Bake');`
```php
         if (!Plugin::loaded('WyriHaximus/TwigView')) {
             Plugin::load('WyriHaximus/TwigView', ['bootstrap' => true]);
         }
```
runs

and when user uses `$this->addPlugin('Bake');` ,  `bootstrap` hook 

there is not conflict between 2 way

src/Plugin.php
```diff
@@ -37,4 +39,19 @@ class Plugin extends BasePlugin
...
+     *
+     * @param \Cake\Core\PluginApplicationInterface $app The host application
+     * @return void
+     */
+    public function bootstrap(PluginApplicationInterface $app)
+    {
+        if (!CorePlugin::getCollection()->has('WyriHaximus/TwigView')) {
+            $app->addPlugin('WyriHaximus/TwigView', ['bootstrap' => false]);
+            $plugin = CorePlugin::getCollection()->get('WyriHaximus/TwigView');
+            $plugin->bootstrap($app);
```
markstory: Why are you disabling the bootstrap hook when calling addPlugin() but then calling it here?

saeideng: just for ensuring that `bootstrap()` calls correctly 
I had a problem with pluginbootstrap here https://github.com/cakephp/cakephp/pull/12245/commits/cbe166dbbc
and we can not calls `pluginBootstrap();` again to cover new plugins

markstory: Ok. Previously when I did testing the iterator implementation in PluginCollection allowed newly added plugins to be part of the loop without requiring additional loops. Perhaps I broke something in the meantime.

PR #461, opened 2018-08-30
Fixed dummy text generation for short text fields

dmromanov:
Depends on cakephp/cakephp/pull/12518, https://github.com/cakephp/cakephp/pull/12552 & https://github.com/cakephp/cakephp/pull/12622

When text field is 1 charcter long generated dummy text in a fixture does not fit into a database resulting in the following error:
```
Exception: Unable to insert fixture "App\Test\Fixture\ExampleFixture" in "App\Test\TestCase\Controller\ExampleControllerTest" test
 case:
SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'value' at row 1 in [.../vendor/cakephp/cakephp/src/TestSuite/Fixture/FixtureManager.php, line 345]
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.6.0",
-        "cakephp/cakephp": "^3.6.0",
+        "cakephp/cakephp": "^3.6.12",
         "cakephp/plugin-installer": "^1.0",
         "wyrihaximus/twig-view": "^4.3.4"
     },
--- a/src/Shell/Task/FixtureTask.php
+++ b/src/Shell/Task/FixtureTask.php
@@ -366,7 +366,7 @@ class FixtureTask extends BakeTask
                         } else {
                             $insert = "Lorem ipsum dolor sit amet";
                             if (!empty($fieldInfo['length'])) {
-                                $insert = substr($insert, 0, (int)$fieldInfo['length'] - 2);
+                                $insert = substr($insert, 0, (int)$fieldInfo['length'] > 2 ? (int)$fieldInfo['length'] - 2 : (int)$fieldInfo['length']);
                             }
                         }
                         break;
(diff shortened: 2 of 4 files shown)


dmromanov:
cakephp/cakephp/pull/12518

dereuromark:
Due to the depend on core release, we need to wait till Core has a new patch tag, and then use that tag here in composer.json as constraint maybe?

    "cakephp/cakephp": "^3.6.0",

would then have to be

    "cakephp/cakephp": "^3.6.x",

dmromanov:
That sounds reasonable. I guess next version will be 3.6.11 - I will add it to `composer.json`

dereuromark:
We could now tag it based on https://github.com/cakephp/cakephp/releases/tag/3.6.11
But your last commit seems to introduce invalid changes

composer.json
```diff
@@ -19,14 +19,24 @@
     },
     "require": {
         "php": ">=5.6.0",
-        "cakephp/cakephp": "^3.6.0",
+        "cakephp/cakephp": "dev-master",
```
markstory: We can't merge with his change present.

composer.json
```diff
@@ -19,14 +19,24 @@
     },
     "require": {
         "php": ">=5.6.0",
-        "cakephp/cakephp": "^3.6.0",
+        "cakephp/cakephp": "dev-schema-fixes",
```
markstory: Can this be `^3.6.11` now?

dmromanov:
Depends on https://github.com/cakephp/cakephp/pull/12552

composer.json
dmromanov: Reverted.

composer.json
dmromanov: Done.

PR #467, opened 2018-10-11
4.x lower case template folders

ADmad:


markstory:
Changes look good to me, but travis is sad.

ADmad:
Hmm.. I don't get any failure locally.

ADmad:
For some reason the folder for the cell's template file isn't being creating on travis and since our `File` class is so very wonderful it merrily carries along even though `File::pwd()` is unable to set proper value for `File::$path` property, instead of throwing an exception.

src/View/BakeView.php
```diff
@@ -26,6 +26,13 @@ class BakeView extends TwigView
     use ConventionsTrait;
     use InstanceConfigTrait;
 
+    /**
+     * Folder containing bake tempalates.
```
dereuromark: typo: templates

Issue #469, opened 2018-11-08
Include nullable into entity propery annotations


dereuromark:
Follow https://github.com/dereuromark/cakephp-ide-helper/issues/94

     * @property \Cake\I18n\FrozenTime $completed

Should be (based on schema "Default NULL")

     * @property \Cake\I18n\FrozenTime|null $completed

dereuromark:
PR open.

PR #470, opened 2018-11-08
Add nullable annotations

dereuromark:
Fixes https://github.com/cakephp/bake/issues/469 and PHPStan problems with nullable DB fields on entity annotations.


dereuromark:
I appreciate help here if someone has a bit of spare time.
I did https://github.com/dereuromark/cakephp-ide-helper/pull/95 already, but for now I dont have too much time to fix all those tests here in the bake plugin.

dereuromark:
Should we only check the keys maybe? seems like the values can differ per DB type.
Or is this a bug/issue?

smarek:
CI not passing here should be fixed by recent changes to .travis I think, is it possible to re-run the test here?

dereuromark:
Those are fails due to my additions. But I am saying that we might not want to check the results so exactly, it could be fine to check only the keys of the array in some cases.

src/Shell/Task/ModelTask.php
```diff
@@ -560,7 +560,9 @@ public function getEntityPropertySchema(Table $model)
 
             $properties[$column] = [
                 'kind' => 'column',
-            ] + $columnSchema;
+                'type' => $columnSchema['type'],
+                'null' => $columnSchema['null'],
+            ];
```
dereuromark: Why not passing all?

markstory: We don't need all the properties right now, and it creates a wider interface than we need with more chances to break behavior.

PR #476, opened 2018-12-01
Use https

dereuromark:


dereuromark:
I think tests are too literal and need updating as well.

PR #478, opened 2018-12-16
3.next - Template renderer

markstory:
I'd like to start updating bake to use Command classes. Before I can do that I need a way to render templates outside of a task. I've kept BakeTemplateTask around as I suspect there are userland tasks that require it.

I'm hoping to do a 3.x compatible release of bake that includes commands. I don't know if I'll be able to convert all of bake to commands without needing to bump a major version as there could potentially be userland tasks that are extending our tasks :cry:


markstory:
Going to close this and apply this work to 4.x. This would be a reasonably high risk change and there isn't a ton of value in doing two breaking changes close together.

PR #480, opened 2018-12-17
4.x maintainence

ADmad:


ADmad:
Do we need to run the test suite for all the dbs? Bake does do db relations introspection but for that just running testsuite on mysql and pgsql should suffice. We can drop the sqlite jobs.

markstory:
Sqlite is nice as it gives a way to clone the repo and run tests with no configuration changes required. We could remove it though as it does result in a bunch of skipped tests.

Issue #485, opened 2018-12-21
Bake 1.9.1 with CakePHP 3.7 generates deprecated code


challgren:
When running `bin/cake bake model <modelName>` the Table class that is generated builds a `validationDefault` with deprecated calls `allowEmpty` and `notEmpty`

```
/**
     * Default validation rules.
     *
     * @param \Cake\Validation\Validator $validator Validator instance.
     * @return \Cake\Validation\Validator
     */
    public function validationDefault(Validator $validator)
    {
        $validator
            ->integer('id')
            ->allowEmpty('id', 'create');

        $validator
            ->scalar('name')
            ->maxLength('name', 255)
            ->requirePresence('name', 'create')
            ->notEmpty('name');

        $validator
            ->boolean('active')
            ->allowEmpty('active');

        $validator
            ->integer('energy_audit_count')
            ->allowEmpty('energy_audit_count');

        return $validator;
    }
```

markstory:
Closing as a pull request is open now.

PR #486, opened 2018-12-21
Remove use of deprecated notEmpty/allowEmpty validator methods.

ADmad:
Refs #485.

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -759,12 +759,12 @@ class ModelTask extends BakeTask
 
         if (in_array($fieldName, $primaryKey)) {
             $validation['allowEmpty'] = [
-                'rule' => 'allowEmpty' . $this->getEmptyMethodType($fieldName, $metaData),
+                'rule' => $this->getAllowEmptyMethod($fieldName, $metaData),
                 'args' => ["'create'"],
             ];
         } elseif ($metaData['null'] === true) {
             $validation['allowEmpty'] = [
-                'rule' => 'allowEmpty' . $this->getEmptyMethodType($fieldName, $metaData),
+                'rule' => $this->getAllowEmptyMethod($fieldName, $metaData),
                 'args' => [],
             ];
         } else {
@@ -772,9 +772,9 @@ class ModelTask extends BakeTask
                 'rule' => 'requirePresence',
                 'args' => ["'create'"],
             ];
-            $validation['notEmpty'] = [
-                'rule' => 'notEmpty' . $this->getEmptyMethodType($fieldName, $metaData),
-                'args' => [],
+            $validation['allowEmpty'] = [
+                'rule' => $this->getAllowEmptyMethod($fieldName, $metaData),
+                'args' => ['false'],
             ];
         }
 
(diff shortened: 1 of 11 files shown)


tests/comparisons/Model/testBakeTableValidation.php
```diff
@@ -44,32 +44,33 @@ public function validationDefault(Validator $validator)
...
             ->requirePresence('name', 'create')
-            ->notEmpty('name')
-            ->maxLength('name', 100, 'Name must be shorter than 100 characters.');
+            ->notEmptyString('name');
 
         $validator
             ->nonNegativeInteger('count')
             ->requirePresence('count', 'create')
-            ->notEmpty('count');
+            ->notEmptyString('count');
```
dereuromark: string for ints is a bit weird.
can we use ...Scalar here etc?

markstory: From what I remember the thinking was that POST data from forms is always strings.

src/Shell/Task/ModelTask.php
```diff
@@ -781,6 +793,34 @@ public function fieldValidation($schema, $fieldName, array $metaData, $primaryKe
...
+            case 'datetime':
+            case 'timestamp':
+                return 'DateTime';
+        }
+
+        if (preg_match('/file|image/', $fieldName)) {
+            return 'File';
+        }
+
+        return 'String';
```
dereuromark: The alternative would be to use normal "camelBacked" (string, dateTime) and ufcirst() the result as suffix.
This way the "type" return here is true. But no big deal either way.

tests/comparisons/Model/testBakeTableValidation.php
dereuromark: But we know that this is not true for this validation process. Marshalled data in general can also come from webservice, and other endpoints, as such they sure can be in a nonstring format already.

ADmad: Yes data from webservices can be in non string format but if a value of valid type is set for a non string field then it is non empty anyway and "not empty" check will pass as expected.

Taking the example of an integer field only way to indicate empty value is passing `null` or empty string for it, which is covered by using `notEmptyString()`. (All `notEmpty*` methods treat `null` as empty as expected).

dereuromark: I didnt mean null, but but the actual value (int, float, ..).
So for me scalar still sounds more correct than string.
We might have to patch the core here.

ADmad: Bake can be updated if/when core is updated. For now this is good to go.

src/Shell/Task/ModelTask.php
```diff
@@ -766,6 +757,27 @@ public function fieldValidation($schema, $fieldName, array $metaData, $primaryKe
...
+                'rule' => 'allowEmpty' . $this->getEmptyMethodType($fieldName, $metaData),
+                'args' => [],
+            ];
+        } else {
+            $validation['requirePresence'] = [
+                'rule' => 'requirePresence',
+                'args' => ["'create'"],
+            ];
+            $validation['notEmpty'] = [
+                'rule' => 'notEmpty' . $this->getEmptyMethodType($fieldName, $metaData),
```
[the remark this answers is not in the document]
ADmad: Woops, don't know how I missed that.

ADmad:
@markstory I have fixed the method calls.

PR #488, opened 2018-12-25
Move PluginTask to a command

markstory:
I've created another utility class for subprocess calls so that we don't end up with another big base class that is a kitchen sink of features.

### DIFF
--- a/src/Command/PluginCommand.php
+++ b/src/Command/PluginCommand.php
@@ -13,10 +13,13 @@ declare(strict_types=1);
  * @since         0.1.0
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
-namespace Bake\Shell\Task;
+namespace Bake\Command;
 
 use Bake\Utility\TemplateRenderer;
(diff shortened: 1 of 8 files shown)


src/Command/PluginCommand.php
```diff
@@ -91,34 +97,36 @@ public function main($name = null)
      * Also update the autoloader and the root composer.json file if it can be found
      *
      * @param string $plugin Name of the plugin in CamelCased format
+     * @param \Cake\Console\Arguments $args The command arguments.
+     * @param \Cake\Console\ConsoleIo $io The console io
      * @return bool|null
      */
-    public function bake($plugin)
+    public function bake($plugin, $args, $io)
```
saeideng: why we could not use  `$this->io` object with assigning these vars withing `execute()` ?

markstory: We could. I've been passing the arguments around in other commands so that the commands have as few properties as possible. I'm also considering making only `execute()` public and all other methods protected.

PR #489, opened 2018-12-28
4.x - Model command

markstory:
Move the model task to a command. I've not finished the `model all` subcommand, but this is already really large and I can do `model all` separately.

I've had to add more fixtures to allow model generation to be isolated from other test case files. As I converted tests to be integration tests the re-use of models was resulting in test stubs being deleted causing other tests to fail.


phpcs.xml.dist
```diff
@@ -5,4 +5,9 @@
     <rule ref="CakePHP" />
 
     <exclude-pattern>*/comparisons/*</exclude-pattern>
+
+    <rule ref="CakePHP">
+        <!-- Exclude unwanted sniffs -->
+        <exclude name="Generic.Commenting.Todo.TaskFound" /> <!-- Excluded because it triggers on Todo* models -->
```
ADmad: This shouldn't be required, the Todo sniff has already been removed from codesniffer's `next` branch.

markstory: I don't think bake is using that version yet.

markstory: Interesting, I was trying to work around the stickler errors. Stickler doesn't have the dev-next branch installed.

PR #490, opened 2018-12-31
Add `model all` command

markstory:
Implement `model all` as a command. I've added a utility class that helps making tests for `all` commands easier. In the past we had to skip all tests often as additional fixtures would be left behind. With the new SubsetSchemaCollection we can augment how schema reflection is done
and only return the exact fixtures we want. This should let us have fewer skipped tests.


src/Command/ModelAllCommand.php
```diff
@@ -0,0 +1,81 @@
...
+     * @param \Cake\Console\ConsoleIo $io The console io
+     * @return null|int The exit code or null for success
+     */
+    public function execute(Arguments $args, ConsoleIo $io)
+    {
+        $this->extractCommonProperties($args);
+        $connection = ConnectionManager::get($this->connection);
+        $scanner = new TableScanner($connection);
+        foreach ($scanner->listUnskipped() as $table) {
+            TableRegistry::getTableLocator()->clear();
```
ADmad: Use the `LocatorAwareTrait` instead of `TableRegistry` perhaps?

PR #504, opened 2019-02-07
Make sure the confirm JavaScript is escaped

groovenectar:
Without the escaping, it generates this:

`<a href="#" onclick="if (confirm("Are you sure you want to delete # x?")) { document.post_token.submit(); } event.returnValue = false; return false;">Delete</a>`

Which of course creates an issue with the double quotes... Adding the escaping produces this:

`<a href="#" onclick="if (confirm(&quot;Are you sure you want to delete # x?&quot;)) { document.post_token.submit(); } event.returnValue = false; return false;">Delete</a>`

Which allows the records to be deleted

### DIFF
--- a/src/Template/Bake/Element/form.twig
+++ b/src/Template/Bake/Element/form.twig
@@ -21,7 +21,7 @@
         <li><?= $this->Form->postLink(
                 __('Delete'),
                 ['action' => 'delete', ${{ singularVar }}->{{ primaryKey[0] }}],
-                ['confirm' => __('Are you sure you want to delete # {0}?', ${{ singularVar }}->{{ primaryKey[0] }})]
+                ['confirm' => __('Are you sure you want to delete # {0}?', ${{ singularVar }}->{{ primaryKey[0] }}), 'escape' => true]
             )
         ?></li>
 {% endif %}
--- a/src/Template/Bake/Template/index.twig
+++ b/src/Template/Bake/Template/index.twig
@@ -71,7 +71,7 @@
                 <td class="actions">
                     <?= $this->Html->link(__('View'), ['action' => 'view', {{ pk|raw }}]) ?>
                     <?= $this->Html->link(__('Edit'), ['action' => 'edit', {{ pk|raw }}]) ?>
-                    <?= $this->Form->postLink(__('Delete'), ['action' => 'delete', {{ pk|raw }}], ['confirm' => __('Are you sure you want to delete # {0}?', {{ pk|raw }})]) ?>
+                    <?= $this->Form->postLink(__('Delete'), ['action' => 'delete', {{ pk|raw }}], ['confirm' => __('Are you sure you want to delete # {0}?', {{ pk|raw }}), 'escape' => true]) ?>
                 </td>
             </tr>
             <?php endforeach; ?>


dereuromark:
Nice catch!

ADmad:
@groovenectar Have you actually encountered the lack of escaping? Because escaping is turned on by default.

dereuromark:
Indeed, only one part is escape=>false: https://github.com/cakephp/cakephp/blob/master/src/View/Helper/FormHelper.php#L1954
But that ought to be this way.

groovenectar:
@ADmad Oh! I didn't actually know that's it's turned on by default... But yes, I run into it all the time...

For more details, I run into it via Croogo admin: https://github.com/croogo/croogo

Which is a fairly pure CakePHP CMS... but perhaps Croogo disables escaping, I'm just not sure where that would happen!

groovenectar:
@dereuromark I see that `escape=>false` is in the `postLink()` method, which is what the Delete links are using...

ADmad:
@groovenectar That's `'escape' => false` is only applicable to the `action` attribute of form.

groovenectar:
@ADmad Hmm, offhand would you happen to know where I could find how it's defaulting to `false`? I don't see the option in the `app.php` config...

ADmad:
I suggest first trying out baking from the default skeleton app and confirm if this indeed an issue.

ADmad:
I just tested using latest skeleton app and the confirm message is escaped properly and the delete link works as expected.

@groovenectar Please make sure you have the latest CakePHP and bake plugin. Maybe there might have been a bug in an older version of core/bake.

groovenectar:
Okay got it, thanks, I will look and see what the issue could be on my end

groovenectar:
I found what's going on.... For some reason Croogo does disable the escaping by default, overriding the Html Helper: https://github.com/croogo/croogo/blob/abcc39b4d1f3e05a2a3faa45a64d71bdd422edc6/Core/src/View/Helper/CroogoHtmlHelper.php#L227

This is must be why it always comes up

groovenectar:
Looking at this further, I still wish it would explicitly enabling escaping. Obviously for the delete action it is necessary, and it would ensure that even if the behavior of the HtmlHelper has been overridden somewhere, the button will function

dereuromark:
It is escaping by default - everywhere.

groovenectar:
@dereuromark Yes that is clearly the default behavior of Cake as it should be, but obviously there are cases where an application is overriding the default.... Croogo has over 700 GH stars https://github.com/croogo/croogo, so it is out there in the wild.. And if it were necessary to specify escaping one way or another, obviously we would make it escape

dereuromark:
You cannot double or triple escape output because of this. 
The source of the issue, this app, should then be fixed instead.

groovenectar:
I agree that the app should be fixed (started a thread there)... Although, of course explicitly specifying the escaping will not not cause the escaping run again or double, it will only ensure that the output is absolutely as intended despite any application quirks..

In my opinion since the OSS can be so widely used, it is okay to consider potential variations in deployments.. but it's only my opinion and not necessarily how things should be

ADmad:
The default bake templates generate code considering the core defaults. If your app/plugin overrides sensible defaults then you should use your own custom bake templates.

PR #505, opened 2019-02-21
Fix error reported by phpstan.

ADmad:


dereuromark:
new sniffs already pay off big time :)

Issue #442, opened 2018-05-04
Use column comments from database for entity annotation


dmromanov:
It would be great if Bake was using column comments from a table in a database for annotating an Entities' properties.

dereuromark:
They would need to be stripped/sanitized to avoid chars accidentally breaking the doc blocks.

dereuromark:
Is someone interested in making a PR here?

dmromanov:
I am interested but currently can not give any estimates.

ravage84:
@dmromanov could you show us an example of how it could look and behave? What is the scope of this feature. How far do you want to go?

dereuromark:
It should not be too difficult to PR this.
Do you want to try it?

dmromanov:
This feature is on my list, but I can not give any estimates.

ravage84:
@dmromanov thank you for your explanation. I stand with my counter-arguments on this feature as mentioned in https://github.com/cakephp/bake/issues/465#issuecomment-428910084.

But if we are going to implement this feature into bake directly and not as an extension as mentioned by @dereuromark in https://github.com/cakephp/bake/issues/465#issuecomment-439724532, I would prefer this behavior to be optional. For example activatable through a CLI parameter and/or an interactive question.

dereuromark:
I would rather go away from interactive or CLI towards the global config. That is way cleaner in terms of consistency in the project as well as easier to use for CLI typing (annoying to type out all of them).
I said that before numerous times. The current config hell via CLI is already a PITA and pitfall.
But then, this would be an easy to opt-in feature :)

dmromanov:
Opened a PR.

Issue #465, opened 2018-10-09
Import comments from tables and fields in models


ADmad:
_From @hervehobbes on October 9, 2018 8:31_

Hello,

This is a:

* [] bug
* [x ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6
* Platform and Target: Apache, Ubuntu, Mysql

### What I'd like
When we bake models, if a table in the database has a comment, write back this comment in the Model/Table/MyTable.php script.
The same thing could be done for comments on fields

I'm not enough confortable with Cake' source code to create a PR.

_Copied from original issue: cakephp/cakephp#12631_

ravage84:
While I see some value on this, I think it is not a really good idea.

I don't know the actual use case of @hervehobbes but we are putting comments into the database on fields and tables because we want to document them. So the DB becomes self-documenting. The proposed feature would __copy__ this documentation to the table classes, which could be good but I see two problems:

1. documentation could get out of date or ouf ot sync, which is worse than no documentation or having to fetch the DB to get the correct documentation
2. It could clutter the table code file with lots of comments which could make it huge and not necesary more readable.

Thus I tend to :-1:

dmromanov:
Related to #442

dereuromark:
No, the other one is about using the comments to generate specific bake templates.
This here is just to sync the comments into doc blocks, if I understand correctly.

Not sure if this should be done though.
Maybe a custom user land extension suffices here as solution.

dereuromark:
Duplicate, closing in favor of other one.

PR #507, opened 2019-03-09
Use column comments for annotations when baking entities

dmromanov:
#442


dmromanov:
Note: column comments are supported in Sqlserver, but CakePHP Schema does not retrieve ones from db.

src/View/Helper/DocBlockHelper.php
```diff
@@ -93,60 +85,41 @@ public function associatedEntityTypeToHintType($type, Association $association)
...
-     *     ...
-     * ]
-     * ```
-     *
-     * @see \Bake\Shell\Task\ModelTask::getEntityPropertySchema
-     *
-     * @param array $propertySchema The property schema to use for generating the type map.
-     * @return array The property DocType map.
-     */
-    public function buildEntityAssociationHintTypeMap(array $propertySchema)
```
markstory: We can't safely remove these methods. The DocblockHelper is part of the public interface of bake as it could be used in userland templates.

dereuromark: Or in https://github.com/dereuromark/cakephp-ide-helper :)

Ali1:
Appreciate [these](https://github.com/cakephp/bake/issues/465#issuecomment-428910084) reservations but I like the idea of keeping Model and DB in sync. Would be really great to look to bake having features to help us do this. i.e. `bin/cake bake model_suggestions` would output a human readable diff-like set of suggestions for doc blocks, validators and db comments as applicable which can be run manually after modifying db schema.

dereuromark:
Yeah, look into how this maybe could also be shipped as plugin solution, similar to what [Setup](https://github.com/dereuromark/cakephp-setup) does for these "validations" and "schema suggestions". Or even the "schema vs fixture validation tool" there.
Or what [IdeHelper](https://github.com/dereuromark/cakephp-ide-helper) does for annotations and diffs vs code base.
Maybe that would help make this already working and applicable as opt-in for apps until we can then stabilize it further to form a core code enhancement from it if the acceptance is high.

src/View/Helper/DocBlockHelper.php
dmromanov: Got it. I will think about another implementation.

Ali1:
Sound like the way to go @dereuromark :) . One day!

dereuromark:
Closing then for now.

PR #509, opened 2019-03-16
Remove empty line in class doc blocks

ravage84:
I know I'm a nitpicker but I always found the empty line in the class doc blocks superfluous.
Finally fixed that.

Before:
```` php
/**
 * Foo class
 *
 */
````
Now:
```` php
/**
 * Foo class
 */
````


dereuromark:
There is a sniff for it, in the Spryker code sniffer package :)

PR #510, opened 2019-03-16
Remove blank line after class opening brackets

ravage84:
Again, I know I nitpick, but I also never liked the blank line after the class opening brackets.

Neither our [code style](https://book.cakephp.org/3.0/en/contributing/cakephp-coding-conventions.html) nor [PSR-2](https://www.php-fig.org/psr/psr-2/) demand such a blank line.

Our code style actually shows [an example where there  is no blank line:](https://book.cakephp.org/3.0/en/contributing/cakephp-coding-conventions.html#careful-when-using-empty-isset)

```` php
class Thing
{
    private $property; // Defined

    public function readProperty()
    {
        // Not recommended as the property is defined in the class
        if (!isset($this->property)) {
            // ...
        }
        // Recommended
        if ($this->property === null) {

        }
    }
}
````

My change inserts only blank lines where and when necessary instead of always prepending one.

### DIFF
--- a/src/Template/Bake/Command/command.twig
+++ b/src/Template/Bake/Command/command.twig
@@ -26,7 +26,6 @@ use Cake\Console\ConsoleOptionParser;
  */
 class {{ name }}Command extends Command
 {
-
     /**
      * Hook method for defining this command's option parser.
      *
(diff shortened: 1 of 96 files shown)


dereuromark:
Psr kind of soft demands it.
Again,  there is a sniffer in spryker standard to be ported to Cake one day :)

ADmad:
> Neither our code style nor PSR-2 demand such a blank line.

Our new standard for 4.x demands the blank line not be present and it's enforceable/auto fixable.

dereuromark:
Psr kind of soft demands it.
Again,  there is a sniffer in spryker standard to be ported to Cake one day :)

PR #512, opened 2019-03-20
Correct return value type hint annotations for Table::saveOrFail()

ravage84:
There is a reason why it is called "orFail"...

https://api.cakephp.org/3.7/class-Cake.ORM.Table.html#_saveOrFail


dereuromark:
I fixed the same in the IdeHelper too now.

ravage84:
Thank you for your work on the IdeHelper, by the way, @dereuromark . :heart:

PR #513, opened 2019-03-25
Use cakephp to load plugin

nedvajz:
Little update for PluginTask to get rid of deprecated `Plugin::load`

### DIFF
--- a/src/Shell/Task/PluginTask.php
+++ b/src/Shell/Task/PluginTask.php
@@ -14,6 +14,7 @@
  */
 namespace Bake\Shell\Task;
 
+use Cake\Console\ShellDispatcher;
 use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Core\Plugin;
@@ -28,13 +29,6 @@ use Cake\Utility\Inflector;
  */
 class PluginTask extends BakeTask
 {
-    /**
-     * Path to the bootstrap file. Changed in tests.
-     *
-     * @var string
-     */
-    public $bootstrap = null;
-
     /**
      * Tasks this task uses.
      *
@@ -59,7 +53,6 @@ class PluginTask extends BakeTask
     public function initialize()
     {
         $this->path = current(App::path('Plugin'));
-        $this->bootstrap = ROOT . DS . 'config' . DS . 'bootstrap.php';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


markstory:
Looks reasonable, but tests need to be updated.

nedvajz:
yep, shame on me that I did not run the tests locally before. Sorted now

src/Shell/Task/PluginTask.php
```diff
@@ -130,32 +123,27 @@ public function bake($plugin)
...
-                "\nPlugin::load('%s', [%s'bootstrap' => false, 'routes' => true]);\n",
-                $plugin,
-                $autoload
-            ));
-            $this->out('');
-            $this->out(sprintf('%s modified', $this->bootstrap));
+
+        $cmd = ['cake', 'plugin', 'load', $plugin];
+        if ($hasAutoloader) {
+            $cmd[] = '--autoload';
```
saeideng: if you want to use only `Application.php` and `addPlugin()` , `--autoload` is useless

nedvajz: Oh, that is true, it is [deprecated since 3.7.0](https://book.cakephp.org/3.0/en/plugins.html#manually-autoloading-plugin-classes)  
  
Code updated

nedvajz:
Any hint what to do with `codecov/patch`?

dereuromark:
You can ignore that one. All others are good.

src/Shell/Task/PluginTask.php
```diff
@@ -130,32 +123,23 @@ public function bake($plugin)
...
-                "\nPlugin::load('%s', [%s'bootstrap' => false, 'routes' => true]);\n",
-                $plugin,
-                $autoload
-            ));
-            $this->out('');
-            $this->out(sprintf('%s modified', $this->bootstrap));
-        }
+
+        $cmd = ['cake', 'plugin', 'load', $plugin];
+        ShellDispatcher::run($cmd);
```
markstory: You can use `$this->runCommand()` instead.

nedvajz: yep, makes sense, updated

saeideng: if you want to use `dispatchShell()`
remove the `use Cake\Console\ShellDispatcher;` line

PR #514, opened 2019-03-27
Fix regression in previous patch

dereuromark:
BC break ( https://github.com/cakephp/bake/pull/509/files#diff-9e729c6fd0a87c9303d114549f819a45R34 ) fixed
See https://travis-ci.org/dereuromark/cakephp-ide-helper/jobs/511918535


ravage84:
:heart:

dereuromark:
Can you maybe provide a test case for this scenario? as regression protection?

PR #515, opened 2019-03-28
DocBlockHelper Regression Test

ravage84:
As asked by @dereuromark at https://github.com/cakephp/bake/pull/514#issuecomment-477235431

Sorry for sneaking in some unrelated improvements.
Hope they are fine, too.

Didn't have time to cover all methods, though. :sleeping_bed:

### DIFF
--- a/src/View/Helper/DocBlockHelper.php
+++ b/src/View/Helper/DocBlockHelper.php
@@ -36,9 +36,9 @@ class DocBlockHelper extends Helper
         }
 
         $previous = false;
-        foreach ($annotations as $ann) {
-            if (strlen($ann) > 1 && $ann[0] === '@' && strpos($ann, ' ') > 0) {
-                $type = substr($ann, 0, strpos($ann, ' '));
+        foreach ($annotations as $annotation) {
+            if (strlen($annotation) > 1 && $annotation[0] === '@' && strpos($annotation, ' ') > 0) {
+                $type = substr($annotation, 0, strpos($annotation, ' '));
                 if ($this->_annotationSpacing &&
                     $previous !== false &&
                     $previous !== $type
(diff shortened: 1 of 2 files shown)


dereuromark:
I better start running IdeHelper tests against bake dev-master :P
Early warning system

src/View/Helper/DocBlockHelper.php
```diff
@@ -13,7 +13,10 @@
  */
 class DocBlockHelper extends Helper
 {
-    protected $_annotationSpacing = true;
+    /**
+     * @var bool Whether to add a blank line between different class annotations
+     */
+    protected $_classAnnotationSpacing = true;
```
dereuromark: changing this should be avoided if we want to make this a patch release.

dereuromark:
Looks good otherwise. And IdeHelper is also happy.

tests/TestCase/View/Helper/DocBlockHelperTest.php
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * Tests the propertyHints method
+     *
+     * @return void
+     * @covers ::propertyHints
+     */
+    public function testPropertyHints()
+    {
+        $this->markTestIncomplete('Not implemented yet');
```
markstory: Can we implement these or delete the tests?

ravage84: Sure we can, either. I just had no time yet. I just wanted to increase the visibility for the lack of coverage in the mean time.

src/View/Helper/DocBlockHelper.php
ravage84: I would be fine to make it a minor release, too.

Refs: https://book.cakephp.org/3.0/en/contributing/backwards-compatibility.html#working-on-cakephp

ADmad: Is there any actual feature change which warrants a minor release? I haven't been following the issue but this seems to be related to just changing the whitespacing in generated code.

dereuromark: I would also not change the name here, the added doc block info suffices IMO.

PR #516, opened 2019-03-28
Test with PHP 7.1 again

ravage84:
Not sure why we do not test in PHP 7.1, anymore.
Added it again.


dereuromark:
We can actually also drop others, we only need lower and upper limit - no need to overdue the middle.

PR #518, opened 2019-04-06
4.x Add trailing commas to generated code

markstory:
Right now the code generated by bake does not pass codesniffer:dev-next which is what will be the 4.x version of our codesniffer rules.


ADmad:
Looks good, just the failing test needs to be fixed.

PR #521, opened 2019-04-15
bake plugin "require cakephp/cakephp^4.0"

saeideng:


markstory:
Nice stuff, but tests might need to be updated.

templates/bake/Plugin/composer.json.twig
```diff
@@ -20,7 +20,7 @@
     "type": "cakephp-plugin",
     "license": "MIT",
     "require": {
-        "cakephp/cakephp": "^3.5"
+        "cakephp/cakephp": "^4.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7.14|^6.0"
```
ADmad: phpunit also needs to be bumped to `^7.0`.

dereuromark:
4.x branch is currently also red for other merge issues: https://travis-ci.org/cakephp/bake/builds/518078286

saeideng:
@dereuromark  yes whole https://github.com/cakephp/bake/pull/512/commits/84b515ac86011a09f1498e6de3bea187689b5d6b should be in 4.x but it is not

PR #526, opened 2019-04-17
Use new method for new empty entity.

### LINKED ISSUE #523: Cake 4 - Update newEntity
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.0-alpha1
* Platform and Target: PHP 7.3.2

### What you did
I baked a controller for my `Questions` table.

### What happened
The `add` method baked with the following line
```php
$question = $this->Questions->newEntity();
```

This throws an exception, because the method requires a param.
```
Too few arguments to function Cake\ORM\Table::newEntity(), 0 passed in /Users/davidyell/Sites/CakePHP4/src/Controller/QuestionsController.php on line 51 and at least 1 expected
```

### What you expected to happen
I thought that bake would use `newEmptyEntity()`

ADmad:
Closes #523

### DIFF
--- a/templates/bake/element/Controller/add.twig
+++ b/templates/bake/element/Controller/add.twig
@@ -21,7 +21,7 @@
      */
     public function add()
     {
-        ${{ singularName }} = $this->{{ currentModelName }}->newEntity();
+        ${{ singularName }} = $this->{{ currentModelName }}->newEmptyEntity();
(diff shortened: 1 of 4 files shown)


ADmad:
I have fixed the CS failure on `4.x` which is causing build failure.

PR #529, opened 2019-04-19
Generate coding standard compliant files.

ADmad:


templates/bake/Controller/controller.twig
```diff
@@ -18,10 +18,9 @@
  */
 #}
 <?php
+declare(strict_types=1);
 namespace {{ namespace }}\Controller{{ prefix }};
 
-use {{ namespace }}\Controller\AppController;
```
dereuromark: Isnt this needed if you are in a prefix etc?

ADmad: True, so will have to add this statement conditionally based on whether `prefix` is empty or not.

templates/bake/Controller/controller.twig
```diff
@@ -18,10 +18,12 @@
  */
 #}
 <?php
+declare(strict_types=1);
 namespace {{ namespace }}\Controller{{ prefix }};
 
+{% if prefix %}
 use {{ namespace }}\Controller\AppController;
-
+{% endif %}
```
saeideng: within a plugin 
with this `if` block , generated controller will use main `AppController` if user does not use `prefix` , and with use of `prefix` controller will use `AppController` of plugin

ADmad: > within a plugin with this if block , generated controller will use main AppController

No, it will use the plugin's `AppController`.

Issue #531, opened 2019-04-22
Better requirePresence detection


Ali1:
I've only used MySQL, so forgive the narrow view.

But currently, a `requirePresence` rule is baked in for fields that aren't nullable. This is regardless of whether a default field value is set for the column in the database.

I feel it makes more sense if `requirePresence` is baked only for fields that aren't nullable AND don't have default value set for the column.

Would be consistent with MySQL Strict Modes. In strict, MySQL would require a field's presence on create if no default value is set for the column otherwise it generates an error. This seems quite intuitive. It encourages developers to set a blank default value when the field's presence is not required when creating the table - ensures developers define their intention for that column at database level.

Thoughts?

dereuromark:
I ran into the same issue!

Ali1:
```
            if ($metaData['default'] === null) {
                $validation['requirePresence'] = [
                    'rule' => 'requirePresence',
                    'args' => ["'create'"],
                ];
            }
```
I was thinking something to wrap the creating of requirePresence with this in a PR. Which branch should I use?

dereuromark:
You can do this for master, it will be ported to 4.x then as well.

dereuromark:
PR open

PR #532, opened 2019-04-22
Stop baking requirePresence validation when db default set

Ali1:
#531 
Added `$metaData['default'] === false` too in case other drivers use this and for future-proofing to what I hope would be future behaviour. Would make more sense for TableSchema to be setting `['default']` to `false` rather than `null` (and use `null` when the column default is actually `NULL`).

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -768,10 +768,12 @@ class ModelTask extends BakeTask
                 'args' => [],
             ];
         } else {
-            $validation['requirePresence'] = [
-                'rule' => 'requirePresence',
-                'args' => ["'create'"],
-            ];
+            if ($metaData['default'] === null || $metaData['default'] === false) {
+                $validation['requirePresence'] = [
+                    'rule' => 'requirePresence',
+                    'args' => ["'create'"],
+                ];
+            }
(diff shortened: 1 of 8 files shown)


dereuromark:
Tests failing

Ali1:
Can anyone help me run tests on my machine?

image

Never run had to run vendor's tests before and I'm getting issues like constant APP not defined and Cake\Core\Exception\MissingPluginException : Plugin Bake could not be found when I clone this repo and run tests.

dereuromark:
You need to run them from within the plugin, run composer install there and then execute vendor/bin/phpunit.

garas:
PHPStorm is executing PHPUnit with `--no-configuration`. Update PHPStorm debug configuration to use `phpunit.xml.dist` from Bake root.

Ali1:
pgsql doesn't like this. Will look into it later. Anyone got ideas off the top of their head?

dereuromark:
Well, those are other tests, you probably only ran one DB specific type locally. You can either switch your DB locally, or adjust as per travis result.

Ali1:
Ok happy with it.
Just for reference, in both PostgreSQL and MySQL, `$metaData['default'] = 0` for booleans with a default of false.

PR #542, opened 2019-05-09
Fix docblock of controller actions to reflect aiid/uuid agnostic case.

dereuromark:
If you use the slevomatic strict cs checks, it will always wrongly annotate them as string (or int) based on the current primary key type in your app.
Keeping this agostic by saying int|string prevents this and also then reflects that this can only be typehinted once the specific type in schema is detected (which currently is not the case for baked actions).

### DIFF
--- a/src/Template/Bake/Element/Controller/delete.twig
+++ b/src/Template/Bake/Element/Controller/delete.twig
@@ -16,7 +16,7 @@
     /**
      * Delete method
      *
-     * @param string|null $id {{ singularHumanName }} id.
+     * @param int|string|null $id {{ singularHumanName }} id.
      * @return \Cake\Http\Response|null Redirects to index.
      * @throws \Cake\Datasource\Exception\RecordNotFoundException When record not found.
      */
--- a/src/Template/Bake/Element/Controller/edit.twig
+++ b/src/Template/Bake/Element/Controller/edit.twig
@@ -19,7 +19,7 @@
     /**
      * Edit method
      *
-     * @param string|null $id {{ singularHumanName }} id.
+     * @param int|string|null $id {{ singularHumanName }} id.
      * @return \Cake\Http\Response|null Redirects on successful edit, renders view otherwise.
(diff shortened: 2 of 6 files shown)


ADmad:
Action methods arguments are always string as they are parsed from the URL. So claiming they can be `int` makes no sense. If static analysers are unhappy with this then relevent checks should be ignored instead of lying in the docblocks.

dereuromark:
They are of type string maybe, but they are, especially without strict types casted to int if we used typehinting for int here. So this is not exactly a lie, but a soft casting issue.
We could add explicit doc blocks or even casting, but that requires more changes to bake. This would be a workaround for the time being reflecting both id type cases and not adding specific typehints for the time being.

dereuromark:
I want to stress again the importance of having this as agnostic default case.
Strict CS rules otherwise try to auto add typehints, which are wrong.
They semantically are possibly still of type int, even if strictly speaking they come in a strings from URL.

As such I still very strongly suggest we merge this now to let people chose themselves what way they want to go in without going too strict in one direction.

ADmad:
You can try and come up with whatever rationalizations you want but the fact remains that controllers action arguments are always string or null.

If tools are adding `?string` typehint to the argument then they are doing the right thing and you should typecast the value before passing to a function with expects integer. That's the price you pay for using strict typing, can't have it both ways.

Finally you can always use your own templates if you don't like the type used by default bake templates.

dereuromark:
> that controllers action arguments are always string or null.

That's not true. You can have whatever middleware or other pre-controller logic to influence this.

I get what you mean though for the default case. I still think it is best to keep this agnostic and not require the inside string casts everywhere due to wrongly/badly added typehints on those - for simplicity in projects.

markstory:
Wouldn't adding int as a type require.more casting than string|null does?

dereuromark:
It is only for docblock and thus only for semantic understanding of primarykey possibilities.
But I am fine keeping this in my custom themes for now.

Issue #544, opened 2019-05-11
Return value ModelCommand::getDisplayField() must be of the type string


ava007:
### Environment

CakePHP 4.x (latest development version 2019-05-10)
Bake 4.x (latest development version 2019-05-10)
PHP 7.3.5

### What I did
bin/cake bake model images

### What I got
Exception: Return value of Bake\Command\ModelCommand::getDisplayField() must be of the type string, null returned in [...vendor/cakephp/bake/src/Command/ModelCommand.php, line 486]

dereuromark:
PR open

PR #545, opened 2019-05-11
getDisplayField return string

ava007:
#544


src/Command/ModelCommand.php
```diff
@@ -483,7 +483,7 @@ public function getDisplayField(\Cake\ORM\Table $model, \Cake\Console\Arguments
             return $args->getOption('display-field');
         }
 
-        return $model->getDisplayField();
+        return (string)$model->getDisplayField();
```
ADmad: Might be better to change the method's return type hint to allow returning `null` instead. So `: string` to `: ?string`.

src/Command/ModelCommand.php
```diff
@@ -477,7 +477,7 @@ public function findBelongsToMany(\Cake\ORM\Table $model, array $associations):
      * @param \Cake\Console\Arguments $args CLI Arguments
      * @return string
```
ADmad: This should now be `@return string|null`

saeideng:
@ADmad `getDisplayField()` can returns array , true ??
https://github.com/cakephp/cakephp/blob/cc12b275c6775c4cf70d30f966dfc66746cef249/src/ORM/Table.php#L613-L615

dereuromark:
@saeideng No, I think that is a wrongly declared array typehint here. only scalar allowed.
fd9f3931109ffcc03c6d49084cebbd6277849e3f introduced that as part of some cleanup, but i am sure this wasnt correct. Even for multi-primary-key cases only one (the first) qualifies.

ADmad:
I'll check the core and confirm.

dereuromark:
Any update on this one?

src/Command/ModelCommand.php
```diff
@@ -475,9 +475,9 @@ public function findBelongsToMany(\Cake\ORM\Table $model, array $associations):
      *
      * @param \Cake\ORM\Table $model The model to introspect.
      * @param \Cake\Console\Arguments $args CLI Arguments
-     * @return string
+     * @return string|null
      */
-    public function getDisplayField(\Cake\ORM\Table $model, \Cake\Console\Arguments $args): string
+    public function getDisplayField(\Cake\ORM\Table $model, \Cake\Console\Arguments $args): ?string
```
ADmad: Return value for core's `Table:getDisplayField()` is `string|string[]` so adding return typehints of `string` or `?string` both would be incorrect here.

Though we should ensure that wherever the return values of this method is consumed handles array values properly too.

PR #550, opened 2019-05-21
Update templates

cleptric:
That's the inherent PR for the updated app template. See cakephp/app#683

A missing task is to find a better approach for the related actions.
I don't want to clutter the whole sidebar with actions to related models, so I'm thinking to go into a menu direction for the models only.

EDIT: Yes, tests need to be updated :)

### DIFF
--- a/templates/bake/Template/index.twig
+++ b/templates/bake/Template/index.twig
@@ -30,7 +30,7 @@
 {% for field in fields %}
                 <th><?= $this->Paginator->sort('{{ field }}') ?></th>
 {% endfor %}
-                <th><?= __('Actions') ?></th>
+                <th class="actions"><?= __('Actions') ?></th>
             </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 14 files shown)


templates/bake/Template/index.twig
```diff
@@ -19,36 +19,22 @@
...
-    <table cellpadding="0" cellspacing="0">
+    <table>
         <thead>
             <tr>
 {% for field in fields %}
-                <th scope="col"><?= $this->Paginator->sort('{{ field }}') ?></th>
+                <th><?= $this->Paginator->sort('{{ field }}') ?></th>
 {% endfor %}
-                <th scope="col" class="actions"><?= __('Actions') ?></th>
+                <th><?= __('Actions') ?></th>
```
dereuromark: Shouldnt we keep those classes? They can be useful for styling maybe.

cleptric:
I don't get what the `testGetContent*` tests actually test.
Some help is appreciated 🙂

markstory:
@cleptric tests have been updated. The get content tests are snapshot tests that help ensure we don't accidentally change bake's output. When templates are intentionally changed one can update the comparison files and review the diff much like a jest snapshot.

cleptric:
Thank you 🙂  PHPstan failure seem unrelated so this should be good to go now.

PR #556, opened 2019-05-31
Resolve #554 Do not output a full stack trace for a missing table

### LINKED ISSUE #554: 4.x - Fail nicely when missing a table
## What you did
`bin/cake bake template --theme CakeBootstrap --prefix Admin Questiojs`

## What happened?
`Exception: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'cakephp4testing.questiojs' doesn't exist in [/Users/davidyell/Sites/CakePHP4/vendor/cakephp/cakephp/src/Database/Schema/Collection.php, line 134]`

## What you expected to happen
I would think this could output a more concise error message, rather than a PDO exception.

davidyell:
I have made a start on trying to [reduce the amount of noise](https://i.imgur.com/qtpM2ar.png) generated by an error when baking templates.

It seems that when the `getPrimaryKey()` method is called, this causes a chain reaction, because the schema doesn't exist etc.

I'm open to suggestions to improve the Exception handling, as I can see some instances where a stack trace might be helpful. Perhaps the error could be logged instead?

Thoughts welcome and I can update the pull request. 

Thanks!

### DIFF
--- a/src/Command/TemplateCommand.php
+++ b/src/Command/TemplateCommand.php
@@ -365,7 +365,7 @@ class TemplateCommand extends BakeCommand
     public function getContent(Arguments $args, ConsoleIo $io, string $action, ?array $vars = null)
     {
         if (!$vars) {
-            $vars = $this->_loadController();
+            $vars = $this->_loadController($io);
         }
 
         if (empty($vars['primaryKey'])) {
--- a/src/Command/TemplateCommand.php
+++ b/src/Command/TemplateCommand.php
@@ -256,7 +256,7 @@ class TemplateCommand extends BakeCommand
      * - 'keyFields'
      * - 'schema'
      *
-     * @param \Cake\Console\ConsoleIo Instance of the ConsoleIO
+     * @param \Cake\Console\ConsoleIo $io Instance of the ConsoleIO
      * @return array Returns variables to be made available to a view template
      */
     protected function _loadController(ConsoleIo $io): array
--- a/src/Command/TemplateCommand.php
+++ b/src/Command/TemplateCommand.php
@@ -270,13 +270,18 @@ class TemplateCommand extends BakeCommand
 
         $namespace = Configure::read('App.namespace');
 
-        $primaryKey = (array)$modelObject->getPrimaryKey();
-        $displayField = $modelObject->getDisplayField();
-        $singularVar = $this->_singularName($this->controllerName);
-        $singularHumanName = $this->_singularHumanName($this->controllerName);
-        $schema = $modelObject->getSchema();
-        $fields = $schema->columns();
-        $modelClass = $this->modelName;
+        try {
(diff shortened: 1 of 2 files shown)


src/Command/TemplateCommand.php
```diff
@@ -270,13 +270,18 @@ protected function _loadController(): array
...
+            $primaryKey = (array)$modelObject->getPrimaryKey();
+            $displayField = $modelObject->getDisplayField();
+            $singularVar = $this->_singularName($this->controllerName);
+            $singularHumanName = $this->_singularHumanName($this->controllerName);
+            $schema = $modelObject->getSchema();
+            $fields = $schema->columns();
+            $modelClass = $this->modelName;
+        } catch (\Exception $exception) {
+            echo $exception->getMessage() . "\n";
+            exit;
```
dereuromark: exit makes the whole thing not testable. also no error code but success code 0 here.
so this is currently worse than before.

davidyell: Please can you provide constructive feedback. You've provided no requested changes or route to move forward.

dereuromark: It was - you just have to either output a 1 exit code (see status code constants in Command class) - or throw some "AbortException", I suspect the code also offers some abort() convenience wrapper for this.

davidyell: Great, thanks. That's much better than 'this is currently worse than before'

markstory: You can use `$this->abort()` here.

src/Command/TemplateCommand.php
```diff
@@ -270,13 +270,18 @@ protected function _loadController(): array
...
+        try {
+            $primaryKey = (array)$modelObject->getPrimaryKey();
+            $displayField = $modelObject->getDisplayField();
+            $singularVar = $this->_singularName($this->controllerName);
+            $singularHumanName = $this->_singularHumanName($this->controllerName);
+            $schema = $modelObject->getSchema();
+            $fields = $schema->columns();
+            $modelClass = $this->modelName;
+        } catch (\Exception $exception) {
+            echo $exception->getMessage() . "\n";
```
markstory: You should use `$io->error()` here. You'd need to pass the `ConsoleIo` into `_loadController()` though. `echo` can't be caught by the test harnesses.

markstory:
> I'm open to suggestions to improve the Exception handling, as I can see some instances where a stack trace might be helpful. Perhaps the error could be logged instead?

I have considered removing the stack trace from the default console exception handling entirely. I agree it can be a bit noisy and hide the error message. The trace could be output if the `--verbose` flag was set though.

davidyell:
Just going to add a test for the new output

davidyell:
Let me know if this needs any more tweaks. Thanks!

PR #557, opened 2019-06-07
Fixed return type issue

rochamarcelo:
Tested on linux - PHP 7.2.13

Error: Return value of Bake\Shell\Task\SimpleBakeTask::main() must be of the type integer or null, none returned

### DIFF
--- a/src/Shell/Task/BakeTask.php
+++ b/src/Shell/Task/BakeTask.php
@@ -133,6 +133,8 @@ class BakeTask extends Shell
         if (isset($this->params['connection'])) {
             $this->connection = $this->params['connection'];
         }
+
+        return static::CODE_SUCCESS;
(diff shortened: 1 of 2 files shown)


markstory:
Looks like you have some trailing whitespace.

saeideng:
we can remove return types from 
```php
BakeTask.php //public function main()
SimpleBakeTask.php //public function main(?string $name = null): ?int
TemplateTask.php //public function main(): ?int
```

`BakeTask` does not have any return types 
this is not force from core , but these should have same return type

rochamarcelo:
We could return static::CODE_SUCCESS instead of null. What do you think?

markstory:
@rochamarcelo That sounds like a good idea to me.

PR #558, opened 2019-06-15
Using `$this->getRequest()` instead of `$this->request` at controller.

gotoeveryone:
`bake` task is using the following at controller.

- `$this->request`

Changed it to use the following.

- `$this->getRequest()`


ADmad:
There's nothing wrong with the controller using it's protected property.

PR #570, opened 2019-07-25
Use RouteBuilder instance instance of static methods of Router.

ADmad:


templates/bake/Plugin/config/routes.php.twig
```diff
@@ -18,12 +18,13 @@ declare(strict_types=1);
...
 
-Router::plugin(
-    '{{ plugin }}',
-    ['path' => '/{{ routePath }}'],
-    function (RouteBuilder $routes) {
-        $routes->fallbacks(DashedRoute::class);
-    }
-);
+$routes
+    ->setRouteClass(DashedRoute::class)
```
markstory: This will change the default class for routes added outside of the plugin as well. You should move this inside the plugin scope to preserve the previous behavior.

ADmad: My bad, fixed.

PR #574, opened 2019-07-28
Use Class Intersection in Associations

### LINKED ISSUE #573: Annotate table associations using & Intersection
from dereuromark/cakephp-ide-helper#133

Maybe it should annotate like this
```php
/**
 * @property \App\Model\Table\BooksTable&\Cake\ORM\Association\HasMany $Books
 */
class AuthorsTable extends Table
{
    // ....
    $this->Books->link($book, [$author1, $author2]);
}
```

With `|` phpstan throws that BooksTable does not have a method `link`, changing to intersection fixes that.

Maybe should it be the default for CakePHP 4.0 app?

raul338:
Closes #573 

Also [Table::save()](https://api.cakephp.org/3.8/class-Cake.ORM.Table.html#_save) returns false instead of bool


dereuromark:
I have one question:
What if you want to use both & and | - is there any specific order or pattern?
Does & bind stronger than | ? I pressume it does. Is it valid even to combine?

raul338:
According the [phpdoc-parser](https://github.com/phpstan/phpdoc-parser/blob/master/tests/PHPStan/Parser/TypeParserTest.php) tests, you can mix them.

PR #576, opened 2019-07-29
Use Plugin methods for bootstrapping and setting routes.

### LINKED ISSUE #575: 4.x - Bake plugins without routes.php
Plugins generally have a small number of routes connected if at all, so a separate `routes.php` file isn't really necessary.

We can instead include the `routes()` method in baked `Plugin.php` with comment that one can setup their routes in that method.

ADmad:
Avoid loading separate files.

Closes #575.


templates/bake/Plugin/src/Plugin.php.twig
```diff
@@ -19,10 +19,43 @@ declare(strict_types=1);
...
+
+    /**
+     * Add routes for the plugin.
+     *
+     * @param \Cake\Routing\RouteBuilder $routes The route builder to update.
+     * @return void
+     */
+    public function routes(RouteBuilder $routes): void
+    {
+        $routes->plugin(
```
markstory: Should we have a comment along the lines of 'if you have many routes you can create config/routes.php and delete this method'?

I personally like not having to sift through dozens of files to figure out how a plugin works but I can see folks also preferring smaller files that only do one thing.

ADmad: Adding a comment sounds fine.

markstory: Comment added. This can be merged once the next CI run passes.

PR #580, opened 2019-08-05
Fixed typo in form.twig

jeremyharris:


src/Template/Bake/Form/form.twig
```diff
@@ -48,7 +48,7 @@ class {{ name }}Form extends Form
     }
 
     /**
-     * Defines what to execute once the From is being processed
+     * Defines what to execute once the Form is being processed
```
ADmad: I think the statement is grammatically incorrect too. The "being" should be removed.

PR #582, opened 2019-09-10
Update calls to App::path()

ADmad:


dereuromark:
Would be nice to link this in the description to the core PR etc. For future reference.

ADmad:
I amended the commit message to ref. the core PR.

PR #585, opened 2019-10-01
Fix tables overflowing out of container.

ADmad:
Refs #584.

### DIFF
--- a/templates/bake/Template/view.twig
+++ b/templates/bake/Template/view.twig
@@ -105,27 +105,29 @@
             <div class="related">
                 <h4><?= __('Related {{ otherPluralHumanName }}') ?></h4>
                 <?php if (!empty(${{ singularVar }}->{{ details.property }})) : ?>
-                <table>
-                    <tr>
-{% for field in details.fields %}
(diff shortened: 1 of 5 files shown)


templates/bake/Template/view.twig
```diff
@@ -105,27 +105,29 @@
...
+                            <th><?= __('{{ field|humanize }}') ?></th>
+    {% endfor %}
+                            <th class="actions"><?= __('Actions') ?></th>
+                        </tr>
+                        <?php foreach (${{ singularVar }}->{{ details.property }} as ${{ otherSingularVar }}) : ?>
+                        <tr>
+    {% for field in details.fields %}
+                            <td><?= h(${{ otherSingularVar }}->{{ field }}) ?></td>
+    {% endfor %}
+    {% set otherPk = '$' ~ otherSingularVar ~ '->' ~ details.primaryKey[0] %}
```
markstory: I think the original intent was to not have extra whitespace in the HTML. But whitespace is fine in HTML.

ADmad: I have removed the extra whitespace nonetheless.

markstory:
Related to #584

PR #589, opened 2019-10-08
Remove use statement for Cake\ORM\Query because it's not used

DBnR1:
This removes `Cake\ORM\Query` from use statements when baking a `Table` class because it's not actually used by default.


markstory:
This import is added as any non trivial table class *will* need to reference the query as a typehint and we were trying to save folks work down the road.

DBnR1:
I thought that might be the reason, though I personally wouldn't have it there. Closing this.

PR #591, opened 2019-10-12
Add type on methods

thomasherisson-scaledev:
It corrects this mistake :

> Fatal error: Declaration of App\Test\TestCase\Model\Table\DnsLinesTableTest::setUp() must be compatible with Cake\TestSuite\TestCase::setUp(): void


ADmad:
Changes seem good but tests need to be updated.

ADmad:
Closing in favor of  #602

PR #604, opened 2019-11-07
Update bake template to remove typehint.

markstory:
Refs cakephp/cakephp#13852


templates/bake/Command/command.twig
```diff
@@ -50,7 +50,7 @@ class {{ name }}Command extends Command
      * @param \Cake\Console\ConsoleIo $io The console io
      * @return null|int The exit code or null for success
```
ADmad: `|void` also needed here?

PR #606, opened 2019-11-09
Make controllers final by default

josegonzalez:
If you are extending controllers, maybe the initial thing shouldn't be a controller, and instead just a set of traits that can be composed into a class.


dereuromark:
I am not a big fan of starting this in a framework context 
If you really want to extend bake, this is unnecceraly making your life miserable. 
No need for overkill protection like this IMO.

dereuromark:
I guess for normal app controllers this is probably fine.
It just shouldnt be done inside plugins or other resources that are then hard to overwrite/modify if needed. And also not for other non-endpoint resources. Controllers and Commands are probably end points of some sorts that would be weird to further extend in normal use cases.

ADmad:
I don't think this is necessary but of we do want to add it it should be only done for app controllers not plugin.

markstory:
Thinking about it more, I'm not sure there is a ton of value in adding `final`. It is more syntax we need to teach users, and add to examples.

PR #609, opened 2019-11-27
4.x email unique

ADmad:
Setting `email` as unique for all models is a bit over zealous.

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -3,4 +3,4 @@ linters:
     standard: ./phpcs.xml.dist
 
 branches:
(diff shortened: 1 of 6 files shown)


.stickler.yml
```diff
@@ -1,3 +1,6 @@
 linters:
   phpcs:
     standard: ./phpcs.xml.dist
+
+branches:
+  ignore: ['2.x']
```
othercorey: Was this supposed to be 4.x?

markstory: yes

PR #620, opened 2019-12-17
Update development.rst

LuizMarin:
Note that I did not correct the part about creating a plugin.

I suggest creating this for the cake 4.0 documentation because it is different from 2.

### DIFF
--- a/docs/en/development.rst
+++ b/docs/en/development.rst
@@ -189,7 +189,7 @@ templates that bake uses. The best way to do this is:
 #. Bake a new plugin. The name of the plugin is the bake 'theme' name
 #. Create a new directory **plugins/[name]/src/Template/Bake/Template/**.
 #. Copy any templates you want to override from 
-   **vendor/cakephp/bake/templates/bake/Template** to matching files in your
+   **vendor/cakephp/bake/templates/bake/** to matching files in your
    plugin.
 #. When running bake use the ``--theme`` option to specify the bake-theme you
    want to use. To avoid having to specify this option in each call, you can also
@@ -301,4 +301,3 @@ command name::
 .. meta::
     :title lang=en: Extending Bake
     :keywords lang=en: command line interface,development,bake view, bake template syntax,twig,erb tags,percent tags
-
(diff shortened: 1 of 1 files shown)


docs/en/development.rst
```diff
@@ -188,8 +188,8 @@ templates that bake uses. The best way to do this is:
 
 #. Bake a new plugin. The name of the plugin is the bake 'theme' name
 #. Create a new directory **plugins/[name]/src/Template/Bake/Template/**.
```
saeideng: this line forgotten

docs/en/development.rst
```diff
@@ -188,8 +188,8 @@ templates that bake uses. The best way to do this is:
...
-#. Copy any templates you want to override from
-   **vendor/cakephp/bake/src/Template/Bake/Template** to matching files in your
+#. Copy any templates you want to override from
```
saeideng: why extra space?

LuizMarin: do not know. I never used these tools

docs/en/development.rst
LuizMarin: how would it be correct?

markstory: Wouldn't the path be `plugins/[name]/templates/bake/`?

docs/en/development.rst
```diff
@@ -188,8 +188,8 @@ templates that bake uses. The best way to do this is:
 
 #. Bake a new plugin. The name of the plugin is the bake 'theme' name
 #. Create a new directory **plugins/[name]/src/Template/Bake/Template/**.
-#. Copy any templates you want to override from
-   **vendor/cakephp/bake/src/Template/Bake/Template** to matching files in your
+#. Copy any templates you want to override from 
+   **vendor/cakephp/bake/templates/bake/Template** to matching files in your
```
markstory:
```suggestion
   **vendor/cakephp/bake/templates/bake/** to matching files in your
```

There are more templates than just view templates.

LuizMarin: @markstory  

> #. Bake a new plugin. The name of the plugin is the bake 'theme' name
> #. Create a new directory **plugins/[name]/src/Template/Bake/Template/**.
> 

I haven't worked with plugins yet, so I can't say anything about it

ADmad:
Thanks for this, I have opened a new PR fixing all the paths.

PR #623, opened 2019-12-20
Updated appveyor to PHP7

othercorey:
Updated based on new phinx appveyor config.

https://github.com/cakephp/phinx/pull/1671


othercorey:
Testing bake commands is failing, but it doesn't show the output/error message.

```
$this->exec('bake plugin SimpleExample', ['y', 'n']);
```

markstory:
@othercorey We could dump the stdout and stderr to see what is going on.

othercorey:
> @othercorey We could dump the stdout and stderr to see what is going on.

composer.phar needed to be in the path.  I changed the config to download it into a separate directly to test that works.

PR #636, opened 2020-01-18
Update command.twig

lpj145:
this causes little problems, like IDE not found autocomplete, some functions not work like initialize, because contract of command changed.
And for surprise, have a deprecate warning about Cake\Console\Command.


markstory:
Thank you 👍

PR #642, opened 2020-02-04
Fix phpstan build.

markstory:


src/Command/TemplateCommand.php
```diff
@@ -268,6 +268,8 @@ protected function _loadController(ConsoleIo $io): array
 
         $namespace = Configure::read('App.namespace');
 
+        $primaryKey = $displayField = $singularVar = $singularHumanName = null;
+        $schema = $fields = $modelClass = null;
```
saeideng: `$this->abort();` will throws the `StopException` in catch block 
why phpstan need these then ?

dereuromark: Because it is buggy i think.

PR #647, opened 2020-03-05
Refactored to make the list of commands a property

geoidesic:
- Having the list of commands as a public property allows it to be modified without needing to override the execute method.
- This allows one to change the scope of the Commands called also, e.g. for plugins.
- I've also flipped the foreach loop in the execute method for performance (i.e. to avoid an extra loop).

### DIFF
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -21,6 +21,9 @@ use Cake\Console\Arguments;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Datasource\ConnectionManager;
+use Bake\Command\ModelCommand;
+use Bake\Command\ControllerCommand;
+use Bake\Command\TemplateCommand;
 
 /**
  * Command for `bake all`
@@ -33,9 +36,9 @@ class AllCommand extends BakeCommand
      * @var string[]
      */
     protected $commands = [
-        'Bake\Command\ModelCommand',
-        'Bake\Command\ControllerCommand',
-        'Bake\Command\TemplateCommand',
+        ModelCommand::class,
+        ControllerCommand::class,
+        TemplateCommand::class,
     ];
     /**
      * Gets the option parser instance and configures it.
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -30,12 +30,12 @@ class AllCommand extends BakeCommand
     /**
      * All commands to call.
      *
-     * @var array
+     * @var string[]
      */
-    public $commands = [
-        'ModelCommand',
-        'ControllerCommand',
-        'TemplateCommand',
+    protected $commands = [
+        'Bake\Command\ModelCommand',
+        'Bake\Command\ControllerCommand',
+        'Bake\Command\TemplateCommand',
     ];
     /**
      * Gets the option parser instance and configures it.
@@ -93,6 +93,7 @@ class AllCommand extends BakeCommand
         }
 
         foreach ($this->commands as $commandName) {
+            /** @var \Cake\Comand\Command $command */
             $command = new $commandName();
             foreach ($tables as $table) {
                 $subArgs = new Arguments([$table], $args->getOptions(), ['name']);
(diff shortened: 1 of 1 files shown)


src/Command/AllCommand.php
```diff
@@ -27,6 +27,16 @@
  */
 class AllCommand extends BakeCommand
 {
+    /**
+     * All commands to call.
+     *
+     * @var array
```
dereuromark: string[] then

src/Command/AllCommand.php
```diff
@@ -82,13 +92,9 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
...
 
-        $commands = [
-            new ModelCommand(),
-            new ControllerCommand(),
-            new TemplateCommand(),
-        ];
-        foreach ($tables as $table) {
-            foreach ($commands as $command) {
+        foreach ($this->commands as $commandName) {
+            $command = new $commandName();
```
dereuromark: Should we inline annotate here? So execute() below becomes clickable?

src/Command/AllCommand.php
geoidesic: Sorry, I don't understand? Are you saying it should be a string, not an array?

src/Command/AllCommand.php
geoidesic: Sounds good but I don't know how?

ADmad:
> Having the list of commands as a public property allows it to be modified without needing to override the execute method.

Why public and not protected?

src/Command/AllCommand.php
dereuromark: An array of strings. See my code fenced example now.

src/Command/AllCommand.php
dereuromark: `/** @var \Cake\Comand\Command $command */`

src/Command/AllCommand.php
jadb:
```suggestion
     * @var string[]
```

geoidesic: Tests are failing with `Error: Class 'ModelCommand' not found`. I'm not sure how to fix that tbh. I tried adding a `use Bake\Command\ModelCommand;` declaration to the test but that didn't help. 
```
Bake\Test\TestCase\Command\AllCommandTest::testExecuteEverything
Error: Class 'ModelCommand' not found
bake/tests/TestCase/Command/AllCommandTest.php:138
```

geoidesic:
Tests are failing with `Error: Class 'ModelCommand' not found`. I'm not sure how to fix that tbh. I tried adding a `use Bake\Command\ModelCommand;` declaration to the test but that didn't help. 
```
Bake\Test\TestCase\Command\AllCommandTest::testExecuteEverything
Error: Class 'ModelCommand' not found
bake/tests/TestCase/Command/AllCommandTest.php:138
```

src/Command/AllCommand.php
```diff
@@ -27,6 +27,16 @@
  */
 class AllCommand extends BakeCommand
 {
+    /**
+     * All commands to call.
+     *
+     * @var string[]
+     */
+    protected $commands = [
+        'Bake\Command\ModelCommand',
```
ADmad: Add `use` statement for each at top of class then then use `ModelCommand::class` instead of `'Bake\Command\ModelCommand'`.

ADmad:
This should target `3.next` IMO.

geoidesic:
> This should target `3.next` IMO.

Ok. Yes, it was an error to target `master`. I'll close and redo the PR.

geoidesic:
Although isn't `3.next` a stale branch? (6 months).
I actually looked at that and decided that it seems like `1.x` was for cakephp3.x and `master` must be for cakephp4.x.

ADmad:
Sorry I meant `2.next` i.e for next minor `2.1`. Now idea why there's a `3.next` branch in the repo :)

geoidesic:
> Sorry I meant `2.next` i.e for next minor `2.1`. Now idea why there's a `3.next` branch in the repo :)

Ok cool. I'm not sure how to do that as the branch doesn't exist on `cakephp/bake` I've created it on my fork but the PR only allows me to target existing branches on `cakephp/bake` as far as I can see. 

What should I do?

dereuromark:
Why next? This is fully BC right?

PR #648, opened 2020-03-05
Help geoidesic move commands to property

jadb:
Supersedes #647

### DIFF
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -28,6 +28,16 @@ use Cake\Datasource\ConnectionManager;
 class AllCommand extends BakeCommand
 {
     /**
+     * All commands to call. 
+     *
+     * @var array
+     */
+    public $commands = [
+        'ModelCommand',
+        'ControllerCommand',
+        'TemplateCommand'
+    ];
+    /**
      * Gets the option parser instance and configures it.
      *
      * @param \Cake\Console\ConsoleOptionParser $parser Option parser to update.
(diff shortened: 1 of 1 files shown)


geoidesic:
Hi. I managed to fix the failing tests as per AdMad's comments. I'll resubmit a PR.

jadb:
@geoidesic, this PR also keeps the order of operations the same as it was instead of inverting what to loop through first (tables / commands).

src/Command/AllCommand.php
```diff
@@ -82,11 +93,8 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
             $tables = [$name];
         }
 
-        $commands = [
-            new ModelCommand(),
-            new ControllerCommand(),
-            new TemplateCommand(),
-        ];
+        $commands = array_map(function ($c) { return new $c; }, $this->commands);
```
dereuromark: $command is better :)

dereuromark: And inline annotation is still useful here.

geoidesic:
> @geoidesic, this PR also keeps the order of operations the same as it was instead of inverting what to loop through first (tables / commands).

@jadb `array_map` introduces an extra loop, which isn't as performant or as concise as the flip. Not sure why this is better?

dereuromark:
@geoidesic Performance on this nano level is not an issue here, especially not for CLI.
The main focus must be readability for developers.
If that answer is yes for the map, then OK. otherwise it should stay simple loop :)

jadb:
@geoidesic, my goal was not to replace your PR which introduced a **good change**, but to extend on it as it was faster than comments back and forth. Never said one is better than the other, I just pointed out the difference between the two. 

I personally (doesn't mean am right) think that keeping things in the right order of operation (the one already defined) is more important than performance when using bake (only in development).

geoidesic:
@jadb Ok sure. I don't know if changing the order could have a BC impact. It's possible I guess. So maybe your way is safer. I've submitted a new PR to master as there isn't a `2.next` branch. I don't mind how this gets refactored or pulled, just so long as the property is available :)

dereuromark:
Other PR was merged. Follow up PRs can be done if needed.

PR #649, opened 2020-03-05
Make the list of commands run by AllCommands a property

geoidesic:
- Having the list of commands as a protected property allows it to be modified without needing to override the execute method.
- This allows one to change the scope of the Commands called also, e.g. for plugins.
- I've also flipped the foreach loop in the execute method for performance (i.e. to avoid an extra loop).

### DIFF
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -93,7 +93,7 @@ class AllCommand extends BakeCommand
         }
 
         foreach ($this->commands as $commandName) {
-            /** @var \Cake\Comand\Command $command */
+            /** @var \Cake\Command\Command $command */
             $command = new $commandName();
             foreach ($tables as $table) {
                 $subArgs = new Arguments([$table], $args->getOptions(), ['name']);
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -16,14 +16,14 @@ declare(strict_types=1);
  */
 namespace Bake\Command;
 
+use Bake\Command\ControllerCommand;
+use Bake\Command\ModelCommand;
+use Bake\Command\TemplateCommand;
 use Bake\Utility\TableScanner;
 use Cake\Console\Arguments;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Datasource\ConnectionManager;
-use Bake\Command\ControllerCommand;
-use Bake\Command\ModelCommand;
-use Bake\Command\TemplateCommand;
 
 /**
  * Command for `bake all`
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -21,6 +21,9 @@ use Cake\Console\Arguments;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Datasource\ConnectionManager;
+use Bake\Command\ModelCommand;
+use Bake\Command\ControllerCommand;
+use Bake\Command\TemplateCommand;
 
 /**
  * Command for `bake all`
@@ -33,9 +36,9 @@ class AllCommand extends BakeCommand
      * @var string[]
      */
     protected $commands = [
-        'Bake\Command\ModelCommand',
-        'Bake\Command\ControllerCommand',
-        'Bake\Command\TemplateCommand',
+        ModelCommand::class,
+        ControllerCommand::class,
+        TemplateCommand::class,
     ];
     /**
      * Gets the option parser instance and configures it.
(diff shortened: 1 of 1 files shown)


src/Command/AllCommand.php
```diff
@@ -82,13 +95,10 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
...
         }
 
-        $commands = [
-            new ModelCommand(),
-            new ControllerCommand(),
-            new TemplateCommand(),
-        ];
-        foreach ($tables as $table) {
-            foreach ($commands as $command) {
+        foreach ($this->commands as $commandName) {
```
dereuromark: You are changing the order from tables => commands to commands => tables
But I think this is actually correct.
As you first need all the models, before you can do the controller and template (to have the relations). :+1:

jadb: Tables are instantiated earlier on in the table scanner or how were the commands previously aware of relations? 

Also not so confident that changing the order is a good thing (BC?) but am not very familiar with the intricacies of bake either.

dereuromark: Just saying.
A template task might need to be more aware of all the tables existing
Thus letting the MVC arrow guide us here to finish each per type (command) rather than class could definitely make it less error prone.

geoidesic:
Hmm.. I don't understand why the Travis build failed, nor how to fix it.

src/Command/AllCommand.php
jadb: In either case, commands are already aware of all tables which are instantiated earlier is what am trying to say.

What concerns me about the order is more BC than anything else.

jadb:
@geoidesic because the imports are unnecessary from a namespace perspective as they belong to the same namespace (`Bake\Command`) as the class itself.

geoidesic:
@jadb ok but the unit tests fail without the imports, so I don't know how to fix both.

geoidesic:
Maybe my original idea will pass?
```
protected $commands = [
    'Bake\Command\ControllerCommand',
    'Bake\Command\ModelCommand',
    'Bake\Command\TemplateCommand',
];
```

geoidesic:
Nope.. that didn't help. I'm stumped.

src/Command/AllCommand.php
```diff
@@ -16,6 +16,9 @@
  */
 namespace Bake\Command;
 
+use Bake\Command\ControllerCommand;
+use Bake\Command\ModelCommand;
+use Bake\Command\TemplateCommand;
```
dereuromark: Just dont use these use statements
They are not needed.

geoidesic: If I take them out, the unit tests fail.

geoidesic: Ah.. never mind. You're right.

geoidesic: Removing them didn't solve the problem though :) same failure.

dereuromark:
"Cake\Comand\Command. "
Make sure you type annotations correctly.

geoidesic:
Hallelujah!

dereuromark:
So are we ok with the switch in order?
Sounds reasonable to me - as such this can now be continued, right?

jadb:
I personally still think calls should be left in the order they initially were (#648)

src/Command/AllCommand.php
dereuromark: @markstory Are you fine with this change?
I think the new order does make sense.

jadb: @dereuromark I fail to understand the logic behind switching the order. 

Aside from running into BC risk, what else does it "add"?

dereuromark: Proper order of layers first, before moving to the next layer.
It makes so much for sense then doing this per model first, possibly needing to redo e.g. templating and controller based on some changed model, which often escalates into multiple ones due to relation changes.
So making this layer=>model is more logical than model=>layer

jadb: > Proper order of layers first, before moving to the next layer.

When I bake, the right order is by type of "resource", specially in CRUD type of applications where I use bake the most. Which is also why I always bake my tables before baking my controllers. If you aren't using `bake all`, in what order do you bake? 
 
> possibly needing to redo e.g. templating and controller based on some changed model, which often escalates into multiple ones due to relation changes

How does switching the order minimize the "redo"-type of work when model(s) change? Or did I misunderstand the "advantage" you are trying to explain?

dereuromark: > minimize the "redo"-type of work when model(s) change?

If it breaks in the controller, you know you are done with model and can also just do controllers now.

This was meant for "all" command mainly. Which is exactly the class we are talking about here.

dereuromark:
@markstory Are we able to continue here?

PR #653, opened 2020-03-08
Switch to cakephp/twig-view 1.0

othercorey:


ADmad:
After switch to the new twigview plugin do we release a new minor for bake or new major?

dereuromark:
A minor should be fine, as it will auto pull the needed dependency.

othercorey:
The only problem with a minor now is we won't be 100% backwards compatible with cakephp/twig-view. The namespaces will change. I don't think there's anything else we're going to change that will affect end users.

dereuromark:
For a require dev dependency thats probably okish.

othercorey:
Can we marked this as replacing wyrihaximus/twig-view in packagist? Will that throw an error for projects that don't update their own require dependency if they use TwigView outside of bake?

dereuromark:
If you use `conflict` you can do that. It can then only install once you remove the other twig dependency.

ADmad:
I think we should use `conflict`, otherwise there will be problems for those who use twigview directly.

dereuromark:
Now that we have a beta, we should switch to the cake one
It otherwise blocks other plugins from switching

    - cakephp/bake dev-master requires wyrihaximus/twig-view ^5.0.1 -> satisfiable by wyrihaximus/twig-view[5.0.1].
    - cakephp/bake dev-master requires wyrihaximus/twig-view ^5.0.1 -> satisfiable by wyrihaximus/twig-view[5.0.1].
    - cakephp/twig-view 1.0.0-beta1 conflicts with wyrihaximus/twig-view[5.0.1].
    - cakephp/twig-view 1.0.0-beta1 conflicts with wyrihaximus/twig-view[5.0.1].
    - dereuromark/cakephp-dto dev-cake-twig requires cakephp/twig-view ^1.0.0 -> satisfiable by cakephp/twig-view[1.0.0-beta1].
    - Installation request for dereuromark/cakephp-dto dev-cake-twig as 1.0.0 -> satisfiable by dereuromark/cakephp-dto[dev-cake-twig].
    - Installation request for cakephp/bake dev-master as 2.0.5 -> satisfiable by cakephp/bake[dev-master].

composer.json
```diff
@@ -41,6 +41,8 @@
             "Cake\\Test\\": "./vendor/cakephp/cakephp/tests/"
         }
     },
+    "prefer-stable": true,
+    "minimum-stability": "dev",
```
ADmad:
```suggestion
    "minimum-stability": "beta",
```

composer.json
```diff
@@ -21,7 +21,7 @@
         "php": ">=7.2",
         "cakephp/cakephp": "^4.0",
         "cakephp/plugin-installer": "^1.0",
-        "wyrihaximus/twig-view": "^5.0.1"
+        "cakephp/twig-view": "^1.0.0"
```
ADmad:
```suggestion
        "cakephp/twig-view": "^1.0"
```

dereuromark: Why? 3 digit is the default, and same meaning :)

dereuromark:
The conflict should not go here, though. Right?
It should go to twig-view repo as there we need to make composer aware it cannot be used together with the legacy one.

ADmad:
> The conflict should not go here, though. Right?

Yes, it has been added to `cakephp/twig-view`.

othercorey:
If we switch to the beta, it requires everyone to move their minimum stability again. Not sure we want to do that.

dereuromark:
Then release a stable plugin version now.
At this point it is as stable as the other one before - and we can add patches for things that are broken.

othercorey:
Can no one try it in a project somewhere once?

dereuromark:
I already did as part of Dto plugin (see link), and the sandbox using it from project level.
But I didn't use it as project only templating. Maybe someone else can
Or you can add a demo of it to the sandbox.

ADmad:
I reverted back to use `dev-master` of TwigView. Given the flurry of changes we are making to TwigView it's better to keep testing against bake regularly, instead of waiting till next beta to discover bugs/BC breaks.

ADmad:
As I had suspected there were issues which I have fixed.

othercorey:
There are missing variable exceptions in the unit tests with `strict_variables => true`.

dereuromark:
prefix + traitName
not too bad

othercorey:
I could not fix this error:

> 1) Bake\Test\TestCase\Command\TemplateCommandTest::testGetContentWithRoutingPrefix
Twig\Error\RuntimeError: Impossible to access an attribute ("type") on a null variable.
>
> /cake/templates/bake/element/form.twig:52

The test is faking data that always loads the $modelObject. However, setting `modelObject` to the model instance causes another failure. I'm not sure if is valid.

templates/bake/Template/add.twig
```diff
@@ -19,4 +19,4 @@
  * @var \{{ entityClass }} ${{ singularVar }}
  */
 ?>
-{% element 'form' %}
+{{ element('form') -}}
```
ADmad, line 22: Whitespacing consistency should be maintained between the add and edit templates.

Here trailing whitespace is removed using `-}}` while it's not in edit.twig.

othercorey: It shouldn't be needed. I accidentally added a newline and was trying to fix it. Thought I updated both.

ewbarnard:
File `docs/en/development.rst`, section "Bake Template Syntax" links to twig 2.x, saying that Bake uses the Twig 2.x template syntax. `cakephp/twig-view` uses current Twig (3.x), correct? Given that we link to current Twig at the top of `development.rst` we could probably revert the link to plain text "Bake template files use the Twig template syntax."

othercorey:
Updated to official 1.0 release.

templates/bake/Template/index.twig
```diff
@@ -39,7 +39,7 @@
                 <tr>
 {% for field in fields %}
 {% set isKey = false %}
-{% if associations.BelongsTo %}
+{% if associations.BelongsTo is defined %}
```
ADmad, line 42: Turning on `strict_variables` by default for Twig will potentially cause errors in user templates. In order to keep upgrade experience smooth for users we should consider turning is off for bake.

othercorey: They are only enabled in debug which is turned on for unit tests. They shouldn't experience it locally ... right?

ADmad: bake is always used locally with debug on.

othercorey:
- Added `Bake.twigStrictVariables` config that's set by unit tests
- Updated controller.twig to generates @property/@method tags that pass phpcs spacing.

Issue #658, opened 2020-03-16
Test case for Controllers: initialize wrong


dereuromark:
Every baked test case for a controller contains this one `testInitialize` test for initialize()
Probably, because it is also public, as per
https://book.cakephp.org/4/en/controllers.html#the-app-controller

```php
    /**
     * Test initialize method
     *
     * @return void
     */
    public function testInitialize(): void
    {
        $this->markTestIncomplete('Not implemented yet.');
    }
```

But I think this specific method should be exluded as it is not a valid action name.

dereuromark:
Resolved in https://github.com/cakephp/bake/pull/660

PR #660, opened 2020-03-16
Remove not needed testInitialize test

dereuromark:
Resolves https://github.com/cakephp/bake/issues/658

I do have a weird thing showing up now though:
```
+\n
+    /**\n
+     * Test initial setup\n
+     *\n
+     * @return void\n
+     */\n
+    public function testInitialization(): void\n
+    {\n
+        $this->markTestIncomplete('Not implemented yet.');\n
+    }\n
 }\n
```

Where does this one now come from?

### DIFF
--- a/src/Command/TestCommand.php
+++ b/src/Command/TestCommand.php
@@ -81,6 +81,15 @@ class TestCommand extends BakeCommand
     ];
 
     /**
+     * Blacklisted methods for controller test cases.
+     *
+     * @var string[]
+     */
+    protected $blacklistedMethods = [
+        'initialize',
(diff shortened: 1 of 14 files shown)


dereuromark:
Ah, it is this part:
```
{% endif %}
    /**
     * Test initial setup
     *
     * @return void
     */
    public function testInitialization(): void
    {
    }
{% endif %}
}
```

dereuromark:
The whole testInitialization is usually noise in tests and not helpful for most business class tests or functional/integration ones.
As such removing these is the best path here.

Issue #645, opened 2020-03-02
AppController of plugin should be optional


dereuromark:
`cake bake all ExposedUsers -p Sandbox` leads to:

```
Exception: Class 'Sandbox\Controller\AppController' not found
In [/home/vagrant/Apps/sandbox.local/plugins/Sandbox/src/Controller/ExposedUsersController.php, line 13]

Error: [Error] Class 'Sandbox\Controller\AppController' not found in /home/vagrant/Apps/sandbox.local/plugins/Sandbox/src/Controller/ExposedUsersController.php on line 13
Stack Trace:
- /home/vagrant/Apps/sandbox.local/vendor/composer/ClassLoader.php:444
- /home/vagrant/Apps/sandbox.local/vendor/composer/ClassLoader.php:322
- spl_autoload_call - [internal], line ??- class_exists - [internal], line ??- /home/vagrant/Apps/sandbox.local/vendor/cakephp/bake/src/Command/TestCommand.php:243
- /home/vagrant/Apps/sandbox.local/vendor/cakephp/bake/src/Command/TestCommand.php:120
- /home/vagrant/Apps/sandbox.local/vendor/cakephp/bake/src/Command/ControllerCommand.php:201
- /home/vagrant/Apps/sandbox.local/vendor/cakephp/bake/src/Command/ControllerCommand.php:147
- /home/vagrant/Apps/sandbox.local/vendor/cakephp/bake/src/Command/ControllerCommand.php:64
- /home/vagrant/Apps/sandbox.local/vendor/cakephp/bake/src/Command/AllCommand.php:93
- /home/vagrant/Apps/sandbox.local/vendor/cakephp/cakephp/src/Console/BaseCommand.php:175
- /home/vagrant/Apps/sandbox.local/vendor/cakephp/cakephp/src/Console/CommandRunner.php:336
- /home/vagrant/Apps/sandbox.local/vendor/cakephp/cakephp/src/Console/CommandRunner.php:171
- /home/vagrant/Apps/sandbox.local/bin/cake.php:12
```

If no AppController is found inside the plugin, it should silently accept the normal app AppController instead (which most use anyway, there is no need to have one per plugin by default).

The controller code generated is
```php
namespace Sandbox\Controller;

class ExposedUsersController extends AppController
```
And missing `use App\Controller\AppController;` here.

PR #661, opened 2020-03-16
Fix plugin controller baking.

dereuromark:
Resolves https://github.com/cakephp/bake/issues/645
and

> Exception: Class 'Controller\AppController' not found
In [/home/vagrant/Apps/sandbox.local/plugins/Sandbox/src/Controller/ExposedUsersController.php, line 14]
2020-03-16 21:42:01 Error: [Error] Class 'Controller\AppController' not found in /home/vagrant/Apps/sandbox.local/plugins/Sandbox/src/Controller/ExposedUsersController.php on line 14

when baking a controller in a plugin


templates/bake/Controller/controller.twig
```diff
@@ -22,8 +22,13 @@ declare(strict_types=1);
 
 namespace {{ namespace }}\Controller{{ prefix }};
 
+{% if plugin %}
+use {{ appNamespace }}\Controller\AppController;
```
markstory: Should we check to see if the plugin has it's own AppController? Right now the plugin command will generate an AppController that will never be used. If plugin app controllers were used when they existed we could handle more situations.

dereuromark: I dont think in 4.x days this is useful here anymore.
This was mainly a 2.x thing, I never use plugin app controllers here, do you?

dereuromark: I agree that we could check this, but that would be somewhat another improvement maybe on top right? As this is not trivial from the template or code probably to check for possible app controller names here.

markstory: > I agree that we could check this, but that would be somewhat another improvement maybe on top right?

That's fair, I can take care of adding the check. As right now we're creating classes in the plugin that will go unused.

> As this is not trivial from the template or code probably to check for possible app controller names here.

We know the namespace of the plugin and could do a `class_exists()` check for the plugin' app controller.

PR #663, opened 2020-03-24
Fix exception. Cleanup.

dereuromark:
```
bin/cake bake migration_snapshot
```

throws

> Exception: Return value of Bake\Command\EntryCommand::execute() must be of the type int or null, bool returned
In [/sandbox.local/vendor/cakephp/bake/src/Command/EntryCommand.php, line 130]

As the Shell return value is still

    @return int|bool|null

So this fixes it

Also fixed phpstan level to 5, as everything below is not a good quality.
Did some minor cleanup along the way.

I don't think anyone is extending the public methods of those classes.
But we can make this a minor if needed.


ADmad:
> I don't think anyone is extending the public methods of those classes.

I do use an extended version of `ModelCommand` for e.g.

Issue #674, opened 2020-04-13
Incorrect calculation for theme choices --theme


ewbarnard:
See https://github.com/cakephp/docs/issues/6574 - however, when I change CommonOptionsTrait.php to look in templates/bake, the list includes `Bake|Migrations|WyriHaximus/TwigView`. Would it be best to exclude Migrations and TwigView from that list? My suggestion is to include the whole list.

I have a PR available. No unit tests are affected (which is why we missed it originally, no doubt).

**EDIT:** The proposed change is lines 84-87 in my fork - https://github.com/ewbarnard/bake/blob/761aab9d8b656e76581702ffeaeaa8d302093b7f/src/Utility/CommonOptionsTrait.php#L84

markstory:
@ewbarnard Those changes look good to me. Would you be able to make a pull request for them?

PR #675, opened 2020-04-13
Use correct path for --theme option with bake

ewbarnard:
Fixes issue #674 - look in templates/bake rather than src/Template/Bake when collecting the list of available themes to use when baking.


markstory:
Thanks :+1:

PR #678, opened 2020-04-17
Focus on primary key protection.

dereuromark:
Too many people trip over this.
I think we should rather focus on the primary key exclusion than having all fields listed, as latter always get out of sync and then especially beginners, but also more experienced devs need forever to find out why.
So switching the check order here.

I always have been using the simplified version for years now, and it absolutely suffices for applications. And keeps dev life simple and balanced here.

### DIFF
--- a/src/View/Helper/BakeHelper.php
+++ b/src/View/Helper/BakeHelper.php
@@ -369,14 +369,14 @@ class BakeHelper extends Helper
         $accessible = [];
 
         if (!isset($fields) || $fields !== false) {
-            if (!empty($fields)) {
-                foreach ($fields as $field) {
-                    $accessible[$field] = 'true';
-                }
-            } elseif (!empty($primaryKey)) {
+            if (!empty($primaryKey)) {
                 $accessible['*'] = 'true';
                 foreach ($primaryKey as $field) {
                     $accessible[$field] = 'false';
+                }
+            } elseif (!empty($fields)) {
+                foreach ($fields as $field) {
+                    $accessible[$field] = 'true';
                 }
             }
         }
--- a/tests/comparisons/Model/testBakeEntityFieldsDefaults.php
+++ b/tests/comparisons/Model/testBakeEntityFieldsDefaults.php
@@ -34,16 +34,7 @@ class TodoItem extends Entity
      * @var array
      */
     protected $_accessible = [
-        'user_id' => true,
-        'title' => true,
-        'body' => true,
-        'effort' => true,
-        'completed' => true,
-        'todo_task_count' => true,
-        'created' => true,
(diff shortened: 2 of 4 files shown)


ADmad:
Making each property accessible explicitly was done as per suggestions of the code review that was done by Mozilla sponsored audit. Having all properties accessible (even besides primary key) by default was considered insecure.

ADmad:
For e.g. for a profile edit form if a developer doesn't use FormProtectionComponent one could sneak in an extra field to change user group/role.

dereuromark:
I wouldnt call that insecure by default. Maybe we need to clarify the need to blacklist fields where needed.
But right now people lose on the other side fields of being saved that shouldn't be protected.
They then blindly add all the fields here, which also doesn't help then either.

> For e.g. for a profile edit form if a developer doesn't use FormProtectionComponent one could sneak in an extra field to change user group/role.

That is the same with the verbose list, as you can now still sneak in all those fields manually whitelisted :) Same thing without form tempering protection.

In the end, I don't mind, as I use my own default here.
But I think this could still prevent a lot of tripups and wtf moments.

ADmad:
What devs do with their code is out of our control but the framework should provide secure defaults wherever possible. I am not in favour of reverting changes done based on a security audit.

steinkel:
It's been a pain for me too sometimes, but I wouldn't remove it for security. Could it be an option if we log a warning when patching an entity with fields not accessible? something like
"Warning, Fields role,id ignored while patching entity User as they are not in listed as accessible"

ADmad:
@steinkel I would be in favor of showing/logging a warning in debug mode when patching non-accessible entities.

dereuromark:
That sounds like a good and helpful compromise.
If you are posting non-accessible fields, you just need to use "fields" key to remove them from getting into the marshaling process. So this sounds like the best way moving forward then.

dereuromark:
But what happens with fields that don't exist? They should silently be OK, I often have extra fields I need for validation rules to do their job.
Only specifically set to "false" should trigger this actually. Which brings us back to my list with the focus on the blacklist than whitelist.

steinkel:
If these additional fields used for validation are going to be patched, then it would make sense to add them into the `accessible` list right? So the logic would check for the fields explicitly added as true in `accessible` and, if not, log a warning message.

dereuromark:
No, not if they are not needed to be put into the entity itself. As they are not fields of the DB table (and the final entity doesn't need them), and as such are not relevant for the "accessible part", only for marshalling to use them.

dereuromark:
Well, I will close then as I guess everyone just uses his own strategy and bake defaults then.

PR #679, opened 2020-04-18
Bake template updates

dereuromark:
Also - there seem to be some files missing for a baked plugin

- gitattributes
- editorconfig


markstory:
Looks like expected output files need to be updated.

PR #680, opened 2020-04-22
Remove manipulation of App.paths.templates

othercorey:
Use the Bake plugin prefix instead of manipulating the global paths.

### DIFF
--- a/src/Command/TaskCommand.php
+++ b/src/Command/TaskCommand.php
@@ -49,6 +49,6 @@ class TaskCommand extends SimpleBakeCommand
      */
     public function template(): string
     {
-        return 'Shell/task';
+        return 'Bake.Shell/task';
     }
 }
--- a/src/View/BakeView.php
+++ b/src/View/BakeView.php
@@ -18,7 +18,6 @@ namespace Bake\View;
 
 use Cake\Core\Configure;
 use Cake\Core\ConventionsTrait;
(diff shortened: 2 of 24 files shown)


src/View/BakeView.php
```diff
@@ -41,21 +40,18 @@ class BakeView extends TwigView
...
-        $paths = (array)Configure::read('App.paths.templates');
-
-        if (!in_array($bakeTemplates, $paths)) {
-            $paths[] = $bakeTemplates;
-            Configure::write('App.paths.templates', $paths);
-        }
-
         $this->_tmpLocation = TMP . 'bake' . DS;
         if (!file_exists($this->_tmpLocation)) {
             mkdir($this->_tmpLocation);
```
othercorey, line 57: btw, I don't see this path used anywhere.

src/Command/TaskCommand.php
```diff
@@ -49,6 +49,6 @@ public function fileName(string $name): string
      */
     public function template(): string
     {
-        return 'Shell/task';
+        return 'Bake.Shell/task';
```
markstory, line 52: This seems fine to me. Plugins offering bake commands already need to do something similar.

othercorey: The goal is to make Bake use config and conventions instead of legacy hacks. This helps ensure we're supporting it properly.

PR #684, opened 2020-04-25
Incorrect method _buildValidator rendered when baking modelless form

sergioalcaraz:
According to the Modelless Forms documentation, the validationDefault method must be used to get an instance of Cake\Validation\Validator


othercorey:
It looks like we don't have a unit test for this command.

markstory:
@othercorey We do just not one that compare the generated file.

PR #687, opened 2020-05-02
Stan

ADmad:


ADmad:
psalm shows pretty weird errors for `EntryCommand`

```
ERROR: InvalidStaticInvocation - src/Command/EntryCommand.php:36:28 - Method Cake\Command\Command::__construct is not static, but is called statically (see https://psalm.dev/014)

class EntryCommand extends Command implements CommandCollectionAwareInterface

ERROR: InaccessibleMethod - src/Command/EntryCommand.php:195:32 - Cannot access protected method Cake\Console\Command\HelpCommand::buildOptionParser from context Bake\Command\EntryCommand (see https://psalm.dev/003)

        $parser = $this->help->buildOptionParser($parser);
```

markstory:
> ERROR: InaccessibleMethod - src/Command/EntryCommand.php:195:32 - Cannot access protected method Cake\Console\Command\HelpCommand::buildOptionParser from context Bake\Command\EntryCommand (see https://psalm.dev/003)

I bet this is because protected methods are *class* based while psalm is inferring them to be instance based.

ADmad:
> I bet this is because protected methods are class based while psalm is inferring them to be instance based.

Yup, that's the case. I just ignored that error.

The other error got sorted out too by updating the `use` statement for `Command` class.

othercorey:
Should we drop an issue in psalm for that?

ADmad:
> Should we drop an issue in psalm for that?

I believe this falls in the category of "allowed by PHP but should be avoided", similar to how PHP vars can be used without proper initialization, yet static analysers do complain about it. So I am unsure if this should be reported.

othercorey:
Up to you. Other languages  use class-based protections so it seems normal.

PR #695, opened 2020-06-29
Update development.rst

Graziel:
NOTE:
i cant make tests to show in command line as hint (in `bin/cake bake test`) so someone would need to rewrite that part


markstory:
Thank you 👍

PR #701, opened 2020-08-02
Fix template override paths

### LINKED ISSUE #698: Documentation for overriding default Bake templates is out of date
Hi,

The documentation states it is possible to override the default bake output:
> If you wish to modify the default output produced by the “bake” command, you can create your own bake templates in your application. This way does not use the --theme option in the command line when baking. The best way to do this is:
> Create a new directory /templates/bake/.
> Copy any templates you want to override from vendor/cakephp/bake/templates/bake/ to matching files in your application.

See https://book.cakephp.org/bake/2/en/development.html#customizing-the-bake-templates. This section says all the default templates you wish to override should reside in `templates/bake`. The templates placed there are however not used, since it should actually be `templates/plugin/Bake/bake`. After updating to those paths, the overrides do work.

What's more, but I'm not sure if this should be it's own issue, but after updating to those paths, determining the `$templateEventName` in `src/View/BakeView::render` fails. It expects to be able to explode the filename for the view on `templates/bake`, but since the actual filename is now `path/to/templates/plugin/Bake/bake`, the explode() yields only 1 result, and you're left with an error complaining about an undefined offset.

All of this was on CakePHP 4.0.9 and bake v2.1.1. I noticed the templates are used in the old location when using v2.0.x, seems to be broken since https://github.com/cakephp/bake/pull/680.

markstory:
The changes in #680 broke compatibility and changed the documented paths that bake would use. I'm reluctant to add another break and go back to what it was as I think the changes to use CakePHP's conventions were generally good ones. This change updates the documentation and removes the `Bake/bake` path from application overrides.

Fixes #698

### DIFF
--- a/src/View/BakeView.php
+++ b/src/View/BakeView.php
@@ -88,11 +88,7 @@ class BakeView extends TwigView
     public function render(?string $view = null, $layout = null): string
     {
         $viewFileName = $this->_getTemplateFileName($view);
-        $templateEventName = str_replace(
-            ['.twig', DS],
-            ['', '.'],
-            explode('templates' . DS . static::BAKE_TEMPLATE_FOLDER . DS, $viewFileName)[1]
-        );
+        $templateEventName = str_replace(DS, '.', $view);
 
         $this->_currentType = static::TYPE_TEMPLATE;
         $this->dispatchEvent('View.beforeRender', [$viewFileName]);
@@ -142,7 +138,10 @@ class BakeView extends TwigView
     {
         $paths = parent::_paths($plugin, false);
         foreach ($paths as &$path) {
-            $path .= static::BAKE_TEMPLATE_FOLDER . DS;
+            // Append 'bake' to all directories that aren't the application override directory.
+            if (strpos($path, 'plugin' . DS . 'Bake') === false) {
+                $path .= static::BAKE_TEMPLATE_FOLDER . DS;
+            }
         }
 
         return $paths;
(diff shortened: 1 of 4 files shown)


src/View/BakeView.php
```diff
@@ -142,7 +138,10 @@ protected function _paths(?string $plugin = null, bool $cached = true): array
     {
         $paths = parent::_paths($plugin, false);
         foreach ($paths as &$path) {
-            $path .= static::BAKE_TEMPLATE_FOLDER . DS;
+            // Append 'bake' to all directories that aren't the application override directory.
+            if (strpos($path, 'plugin' . DS . 'Bake') === false) {
+                $path .= static::BAKE_TEMPLATE_FOLDER . DS;
```
othercorey, line 143: We couldn't agree on a way to extend View to support feature directories to replace this unfortunately. Something to do for cake 5.

othercorey: I think long-term we can actually just remove the bake/ directory in templates in the plugin and avoid the whole issue since templates/plugin/Bake will declutter everyone.

src/View/BakeView.php
```diff
@@ -142,7 +138,10 @@ protected function _paths(?string $plugin = null, bool $cached = true): array
     {
         $paths = parent::_paths($plugin, false);
         foreach ($paths as &$path) {
-            $path .= static::BAKE_TEMPLATE_FOLDER . DS;
+            // Append 'bake' to all directories that aren't the application override directory.
```
ADmad, line 141: So bake themes would still use path `PLUGIN_ROOT/templates/bake/*` right?

othercorey: Yea. It seems we didn't have a test-case for overriding bake template from the application templates. Only theme plugins.

markstory: Yes bake themes and plugin tasks are unchanged by this. This change is only relevant to application overrides.

Issue #598, opened 2019-10-22
4.x Add `--prefix` option to `bake all`


steinkel:
Add `bake all --prefix X` option to quickly bake prefixed controllers/templates.

Ajithp:
I faced the same issue

`bin\cake bake all --prefix admin Users -t AdminLTE`

says
Error: Unknown option `prefix`.

This feature is very helpful and accelrates the development process. Hope to see this in Cakephp official release soon.

M3Y3R:
I made a local fix for myself and it works perfectly for me.

Add in `vendor/cakephp/bake/src/Command/AllCommand.php` the Option:
`->addOption('prefix', [
            'help' => 'The namespace prefix to use.',
            'default' => false,
        ]);`

And in `vendor/cakephp/bake/src/Command/ModelCommand.php` in Line 979 and 1032 after each "getPath"-Call the following code:
`$path = str_replace($this->getPrefix($args) . DIRECTORY_SEPARATOR, '', $path);`

dereuromark:
@M3Y3R Can you make a PR with your suggested changes?
That would help speed up the integration of your addition.

M3Y3R:
@dereuromark I tried to create a pull request, but when I push I get the following message !?

`ERROR: Permission to cakephp / bake.git denied to M3Y3R.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.`

dereuromark:
You need to use normal pull request process:
Fork the repo, add your branch and then create the PR.

PR #708, opened 2020-08-26
Add `--prefix` option to `bake all`

M3Y3R:
#598

### DIFF
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -56,6 +56,9 @@ class AllCommand extends BakeCommand
             'help' => 'Generate code for all tables.',
             'default' => false,
             'boolean' => true,
+        ])->addOption('prefix', [
+            'help' => 'The namespace prefix to use.',
+            'default' => false,
         ]);
 
         return $parser;
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -101,7 +101,13 @@ class AllCommand extends BakeCommand
             /** @var \Cake\Command\Command $command */
             $command = new $commandName();
             foreach ($tables as $table) {
-                $subArgs = new Arguments([$table], $args->getOptions(), ['name']);
+                $options = $args->getOptions();
+                if ($command instanceof ModelCommand) {
+                    if ($args->hasOption('prefix')) {
+                        unset($options['prefix']);
+                    }
+                }
+                $subArgs = new Arguments([$table], $options, ['name']);
                 $command->execute($subArgs, $io);
             }
         }
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -100,13 +100,17 @@ class AllCommand extends BakeCommand
         foreach ($this->commands as $commandName) {
             /** @var \Cake\Command\Command $command */
             $command = new $commandName();
+
+            $options = $args->getOptions();
+            if (
+                $args->hasOption('prefix') &&
+                !($command instanceof ControllerCommand) &&
+                !($command instanceof TemplateCommand)
+            ) {
+                unset($options['prefix']);
+            }
+
             foreach ($tables as $table) {
-                $options = $args->getOptions();
-                if ($command instanceof ModelCommand) {
-                    if ($args->hasOption('prefix')) {
-                        unset($options['prefix']);
-                    }
-                }
                 $subArgs = new Arguments([$table], $options, ['name']);
                 $command->execute($subArgs, $io);
             }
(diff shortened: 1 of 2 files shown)


dereuromark:
@steinkel Do you have any additional tips/help on this one?

steinkel:
> @steinkel Do you have any additional tips/help on this one?

No, after tests are passing we should be OK with the change...

src/Command/ModelCommand.php
```diff
@@ -976,6 +976,7 @@ public function bakeEntity(Table $model, array $data, Arguments $args, ConsoleIo
         $out = $renderer->generate('Bake.Model/entity');
 
         $path = $this->getPath($args);
+        $path = str_replace($this->getPrefix($args) . DIRECTORY_SEPARATOR, '', $path);
```
ndm2, line 979: If there is no prefix, then this will remove all directory separators from the path, hence the failing tests.

However this feels kinda like a hack anyways, I would recommend to maybe instead modify `AllCommand`, so that it only passes the `prefix` option to those commands in the list that actually support it, which currently would be `ControllerCommand` and `TemplateCommand`.

M3Y3R:
Oh, you are right...

That means, i shoul check the CommandName in the foreach and remove the --prefix option from Argument-Class if the CommandName is ModelCommand, right?

To do this, the Argument class would have to provide a method that can remove the option. But this Class is in the repository of CakePhp itself...

What is the best way to do this?

src/Command/AllCommand.php
```diff
@@ -98,7 +101,13 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
             /** @var \Cake\Command\Command $command */
             $command = new $commandName();
             foreach ($tables as $table) {
-                $subArgs = new Arguments([$table], $args->getOptions(), ['name']);
+                $options = $args->getOptions();
+                if ($command instanceof ModelCommand) {
+                    if ($args->hasOption('prefix')) {
+                        unset($options['prefix']);
+                    }
+                }
```
ndm2, line 109:
```suggestion
                if (
                    $args->hasOption('prefix') &&
                    !($command instanceof ControllerCommand) &&
                    !($command instanceof TemplateCommand)
                ) {
                    unset($options['prefix']);
                }
```

It might be better to apply the option only to the commands that are supported, instead of not applying to those that aren't supported, this means less possible failures further down the road (for example in case additional commands are being added).

You can also put it outside the `foreach` loop, as it only needs to be checked once per command.

M3Y3R:
@markstory Should i squash the commits?

markstory:
> Should i squash the commits?

That isn't necessary, I can do that when I merge the changes. It would be good to have a test for this behavior. The AllCommandTest has a few existing tests that you could use as a base to ensure that the controller and templates end up in the right paths. Checking the contents isn't necessary as we have tests for that behavior elsewhere.

dereuromark:
I think we should move forward with this
I also stumbled upon this issue and currently baking is quite limited due to this.

ndm2:
If someone with write access wants to add a test to this so it can get merged.

```php
class AllCommandTest extends TestCase
{
    // ...

    /**
     * Test execute() generating a full stack with prefixes
     *
     * @return void
     */
    public function testExecuteWithPrefix()
    {
        $path = APP;
        $testsPath = ROOT . 'tests' . DS;
        $templatesPath = ROOT . 'templates' . DS;

        $this->generatedFiles = [
            $templatesPath . 'Admin/Products/add.php',
            $templatesPath . 'Admin/Products/edit.php',
            $templatesPath . 'Admin/Products/index.php',
            $templatesPath . 'Admin/Products/view.php',
            $path . 'Controller/Admin/ProductsController.php',
            $path . 'Model/Table/ProductsTable.php',
            $path . 'Model/Entity/Product.php',
            $testsPath . 'Fixture/ProductsFixture.php',
            $testsPath . 'TestCase/Model/Table/ProductsTableTest.php',
            $testsPath . 'TestCase/Controller/Admin/ProductsControllerTest.php',
        ];
        $this->exec('bake all --connection test --prefix admin Products', ['y']);

        $this->assertExitCode(Command::CODE_SUCCESS);
        $this->assertFilesExist($this->generatedFiles);
        $this->assertFileContains(
            'namespace Bake\Test\App\Controller\Admin;',
            $path . 'Controller/Admin/ProductsController.php'
        );
        $this->assertFileContains(
            'use Bake\Test\App\Controller\AppController;',
            $path . 'Controller/Admin/ProductsController.php'
        );
        $this->assertFileContains(
            'class ProductsController extends AppController',
            $path . 'Controller/Admin/ProductsController.php'
        );
        $this->assertFileContains(
            'use Bake\Test\App\Controller\Admin\ProductsController;',
            $testsPath . 'TestCase/Controller/Admin/ProductsControllerTest.php'
        );
        $this->assertOutputContains('Bake All complete');
    }
}
```

dereuromark:
OK, I will add it post merge.

PR #710, opened 2020-09-24
Default to PHP type `string` for unknown/custom database types.

ADmad:
Prior to this fix if a table column was mapped to a custom database type class
the `@property` annotation generated for the field in entity class did not
have a type specified.

### DIFF
--- a/src/View/Helper/DocBlockHelper.php
+++ b/src/View/Helper/DocBlockHelper.php
@@ -165,11 +165,11 @@ class DocBlockHelper extends Helper
      * Converts a column type to its DocBlock type counterpart.
      *
      * This method only supports the default CakePHP column types,
-     * custom column/database types will be ignored.
+     * for custom column/database types `'string'` will be returned.
      *
      * @see \Cake\Database\Type
      * @param string $type The column type.
-     * @return string|null The DocBlock type, or `null` for unsupported column types.
+     * @return string|null The DocBlock type, or `'string'` for unsupported column types.
      */
     public function columnTypeToHintType(string $type): ?string
     {
@@ -215,7 +215,8 @@ class DocBlockHelper extends Helper
                 return '\Cake\I18n\Time';
         }
 
-        return null;
+        // Any unique or custom types will have a `string` type hint
+        return 'string';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/View/Helper/DocBlockHelper.php
```diff
@@ -215,7 +215,7 @@ public function columnTypeToHintType(string $type): ?string
                 return '\Cake\I18n\Time';
         }
 
-        return null;
+        return 'string';
```
othercorey, line 218:
```suggestion
        // Any unique or custom types will have a `string` type hint
        return 'string';
```
The docblock of this function needs to be updated to not say this is "ignored" but defaulted.

tests/comparisons/Model/testBakeEntityWithPropertyTypeHints.php
```diff
@@ -19,7 +19,7 @@
  * @property \Cake\I18n\FrozenTime|null $updated
  * @property array $array_type
  * @property array $json_type
- * @property $unknown_type
+ * @property string $unknown_type
```
dereuromark, line 22: Or should this be mixed?

ADmad: Hmm.. I guess `mixed` would be more technically correct. I used `string` as in most cases that would be case and the dev won't have to edit it :slightly_smiling_face:.

I am open to changing it to `mixed` if that's the consensus.

dereuromark: At least then people would maybe see it and think about what the actual type might be.
With string it could also be harder to spot the issue.
But either way works.

markstory: Unless there are custom type mappings being used unknown column types will be fetched as strings.

dereuromark: I guess that's true :)

othercorey: So bake needs to support custom types for that to work better?

markstory: Supporting custom types would be a great improvement by we'd need a hook point for custom types to indicate their property value type.

ADmad: > Unless there are custom type mappings being used unknown column types will be fetched as strings.

Yup, so since PDO returns everything as string that's a reasonable default.

> Supporting custom types would be a great improvement by we'd need a hook point for custom types to indicate their property value type.

That's beyond the scope of this PR :)

markstory: > That's beyond the scope of this PR :)

Totally agree, I was just thinking out loud.

PR #715, opened 2020-10-19
Clean up docblocks.

dereuromark:


ADmad:
Can you bump up the versions too [here](https://github.com/cakephp/bake/blob/master/composer.json#L53) please.

dereuromark:
Done
It found indeed 2 new issues that weren't detected so far.

PR #722, opened 2020-11-06
Add GA CI workflow

ADmad:


ADmad:
@othercorey Any idea what min. version of twig-view plugin do we need for the `prefer-lowest` build to pass?

othercorey:
> @othercorey Any idea what min. version of twig-view plugin do we need for the `prefer-lowest` build to pass?

We should enforce >= 1.0.2.

othercorey:
> cakephp/plugin-installer (1.0.0)

might be the problem?

ADmad:
Hah, the failure is due to combination of composer v2 (default used by `shivammathur/setup-php`) and plugin-installer v1 (due `--prefer-lowest`).

composer.json
```diff
@@ -19,13 +19,14 @@
     },
     "require": {
         "php": ">=7.2",
-        "cakephp/cakephp": "^4.0",
-        "cakephp/twig-view": "^1.0"
+        "cakephp/cakephp": "^4.1",
```
ADmad, line 22: This was required to fix following failure with `--prefer-lowest`:

```
1) Bake\Test\TestCase\Command\ModelCommandTest::testApplyAssociations
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'Users'
-    1 => 'TodoTasks'
-    2 => 'TodoLabels'
+    0 => 'users'
+    1 => 'todotasks'
+    2 => 'todolabels'
 )
```

othercorey:
Can we disable the double push/pr build?

ADmad:
> Can we disable the double push/pr build?

Done

PR #725, opened 2020-11-11
added timestamptimezone field type

GrandFelix:
With PostgreSQL and timestamptimezone field type you get empty '' record values in fixtures and because of that running phpunit tests fails in some cases.


dereuromark:
Do we also need to adjust https://github.com/cakephp/migrations or https://github.com/cakephp/phinx/ here maybe?

GrandFelix:
@dereuromark I don't think so. It looks migrations and phinx have this already covered.

othercorey (approved):
looks good

PR #727, opened 2020-11-24
Improve bake command names generation.

ADmad:


markstory:
Looks good to me outside of the additional import that isn't needed.

ADmad:
Is someone interested in fixing AppVeyor or we should just drop it?

garas:
There is no `.appveyor.yml` so it doesn't run correctly.

I've enabled `Skip branches without appveyor.yml` in AppVeyor settings.

PR #728, opened 2020-11-24
Allow using "Command/Bake" subfolder for bake commands.

ADmad:
If this subfolder exists, bake plugin will only use commands under that
directory as bake commands by default and will not check the command
files directly under "Command" folder.


ADmad:
@markstory Which the default naming fix I need in previous command, does bake even need to iterator over the commands classes now since `BasePlugin::console()` already does that?

ADmad:
To answer myself base plugin needs it's own command scanner as `BasePlugin`'s scanner adds plugin commands with plugin prefix.

ADmad:
A new minor release would be better suited for this change.

markstory:
> To answer myself base plugin needs it's own command scanner as BasePlugin's scanner adds plugin commands with plugin prefix.

Yep prefix fixing is one of the reasons.

PR #731, opened 2020-11-24
Remove commands for baking deprecated shells and tasks.

ADmad:
I don't see why we need to make using deprecated class types easier :slightly_smiling_face:.


markstory:
I agree with this, but ShellHelpers are still in use by Commands. We should likely keep that one.

ADmad:
I have restored files related to shell helpers.

PR #732, opened 2020-12-07
folder name should be CamelCase

rrd108:
lowercase bake as a folder name is ignored by the bake command

### DIFF
--- a/docs/en/development.rst
+++ b/docs/en/development.rst
@@ -229,7 +229,7 @@ If you wish to modify the default output produced by the "bake" command, you can
 create your own bake templates in your application. This way does not use the
 ``--theme`` option in the command line when baking. The best way to do this is:
 
-#. Create a new directory **/templates/plugin/Bake/**.
+#. Create a new directory **/templates/bake/**.
 #. Copy any templates you want to override from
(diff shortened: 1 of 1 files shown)


dereuromark:
The core ones are not: https://github.com/cakephp/bake/tree/master/templates/bake
Also mine are not : https://github.com/dereuromark/cakephp-setup/tree/master/templates/bake / https://github.com/spryker/cakephp-statemachine/tree/master/templates/bake

There must be some casing issue underneath then. as the lowercase bake should be correct IMO for the filesystem naming.
So this is most likely a bug for app level where the folder lookup doesn't work as documented.

markstory:
The directory name used in the view is also lower cased https://github.com/cakephp/bake/blob/master/src/View/BakeView.php#L33

garas:
`templates/plugin/Bake/` is generic path to [override plugin templates](https://book.cakephp.org/4/en/plugins.html#overriding-plugin-templates-from-inside-your-application).

`templates/bake` is Bake specific path to override templates.

Both methods should work, but preferred is probably the second one.

Docs should be use one of method to avoid confusion.

dereuromark:
That explains it
I would then think we document the latter.

Issue #651, opened 2020-03-06
RFC [Question / Feature]: non-destructive bake all


geoidesic:
So the docs on bake are a bit sketchy. It seems to me that bake is designed to be used once then discarded – generally the approach for scaffolding. I personally don't like that idea. I prefer to use bake as an ongoing tool during development.

__Why?__ 

1. It facilitates easy schema changes
1. I don't want to have to get clever and specific on the command line with bake because it slows me down, so I want to be able to run `bake all --everything -t MyTheme` and have it all just work.
1. It also avoids errors; you know that you'll never overwrite something with bake that you shouldn't have.
1. Makes the separation between boilerplate and custom code crystal clear, which improves maintenance.
1. Facilitates maintenance as you can change boilerplate and re-bake without destroying custom code.
1. DRY

__How?__
To achieve this I think the way that bake outputs should be configurable through overrides or config rather than (or in addition to) via shell arguments.

The architecture I like has the following changes:

__Models__
1. Baked Tables get written to `Models/Table/Baked`. These are overwritten each bake.
1. Extensions to these (which is where you put your custom code), go into `Models/Table/Extension`. Any that are missing are added during bake, no overwriting.
1. `Models/Table` just gets classes that glue the two together. These are overwritten. E.g.
```
include APP . 'Model/Table/Extension/DocumentsTable.php';
``` 

__Controllers__ 
1. Baked Controllers get written to `Controller/Baked`. These are overwritten each bake.
1. `Controller/<Route>` can be configured to add any that are missing (e.g. for tables added newly to the schema). Can configure to add for multiple routes. Newly added are just empty classes extending the bake. This is where you put any custom code. Same applies for main `Controller` folder. Thus the bake method should be generic here, something like `bakeFolder` and then in the bake method one iterates over an `extensions` array which defines the folder and template to be passed to bakeFolder as arguments.

__Entity__
Unchanged

__Templates__
1. Baked templates get written to the `Templates/Baked` folder. Overwrite on re-bake.
1. Extensions get written to the `Templates/Extensions` folder. Only write if missing. The extensions would simply include the corresponding bake by default e.g. 
```
<?php
    include str_replace('Extension', 'Baked', __FILE__);
?>   
```

Let me know if I'm missing something fundamental about bake's philosophy but in my books this is a far more useful arrangement than bake as it currently stands. 

If there's general consensus about this I'll submit a BC PR with these changes.

markstory:
[...]

PR #733, opened 2021-01-21
Persistent classes for Model / Entity

GrandFelix:
added support for persistent classes (Model, Entity). Use --persistent or --remove-persistent to create or remove persistent classes


GrandFelix:
If there will be interest to merge this, I will also fix tests...

tests/comparisons/Model/testBakeTablePersistent.php
```diff
@@ -0,0 +1,125 @@
...
+ * @method \Bake\Test\App\Model\Entity\Item|false save(\Cake\Datasource\EntityInterface $entity, $options = [])
+ * @method \Bake\Test\App\Model\Entity\Item saveOrFail(\Cake\Datasource\EntityInterface $entity, $options = [])
+ * @method \Bake\Test\App\Model\Entity\Item[]|\Cake\Datasource\ResultSetInterface|false saveMany(iterable $entities, $options = [])
+ * @method \Bake\Test\App\Model\Entity\Item[]|\Cake\Datasource\ResultSetInterface saveManyOrFail(iterable $entities, $options = [])
+ * @method \Bake\Test\App\Model\Entity\Item[]|\Cake\Datasource\ResultSetInterface|false deleteMany(iterable $entities, $options = [])
+ * @method \Bake\Test\App\Model\Entity\Item[]|\Cake\Datasource\ResultSetInterface deleteManyOrFail(iterable $entities, $options = [])
+ *
+ * @mixin \Cake\ORM\Behavior\TimestampBehavior
+ */
+class ItemsTable extends ItemsPersistentTable
```
markstory, line 33: I thought the classes that extended the persistent tables wouldn't have bake generated code? Right now it looks like you'd have all your associations defined twice, same with validation rules.

GrandFelix: classes that extends persistent tables are generated for every bake (if approved by user or used --force). Persistent class is baked only once and there are empty functions like validationDefault(), buildRules() etc. But this could be removed and would only empty persistent class. So baked code is generated only in main class, comments in persistent class with property definitions will be removed. As I said here https://github.com/cakephp/bake/issues/651 this approach worked ok for me because at the early stages of development when I add some functions to model and then add/rename some field, add relations I must always watch for those functions, class changes etc.. With that approach I can run bake without worrying that anything would be overwritten.

markstory: Ok, I had it mixed up then. It would be good if the generated persistent classes included some doc block comments on how they differ from the bake managed classes and which methods should go where.

src/Command/ModelCommand.php
```diff
@@ -1021,17 +1078,42 @@ public function bakeTable(Table $model, array $data, Arguments $args, ConsoleIo
...
+                $persistentOut = $renderer->generate('Bake.Model/persistent_table');
+                $io->out("\n" . sprintf('Baking persistent table class for %s...', $name), 1, ConsoleIo::QUIET);
+                $io->createFile($persistentTableFilePath, $persistentOut);
+
+                if (file_exists($persistentTableFilePath)) {
+                    require_once $persistentTableFilePath;
+                }
+            } else {
+                $io->out(
+                    "\n" . sprintf('Skipping creating persistent table class for %s...', $name),
```
markstory, line 1103: Why are there newlines being concatenated here? You could set `$newlines` to `2` instead.

GrandFelix: I guess I did copy paste from lines before https://github.com/cakephp/bake/blob/b83bdd87cd4c0fb3a2e6b246ac147e3d6a8d56d8/src/Command/ModelCommand.php#L1089 But yes, you are right. Will change in coming days.. I just synced my fork because I need it for one of my projects.

Issue #736, opened 2021-02-16
Customizing controller.twig file still loads default when baking


connect232:
Hi,
I have created a custom controller.twig file in the directory templates/bake/Controller.

However, the default controller.twig file is still being used when executing the bake command.

I have found changing $renderer->generate('Bake.Controller/controller) to $renderer->generate('Controller/controller) in vendor/cakephp/bake/src/Command/ControllerCommand.php, makes the command command load my custom controller.twig file.

Can someone advise if the "Bake." in $renderer->generate('Bake.Controller/controller) is needed for other things I am not aware of please?

othercorey:
Please tell us what version of bake and where your controller.twig file is.

connect232:
I am using version 2.3 of bake and I have created a custom controller.twig file in the directory templates/bake/Controller.

markstory:
That template path should be working. I'll see if I can reproduce the problem of templates not being used for controllers.

markstory:
Sorry @connect232 the documentation for application template overrides is wrong. I've opened a pull request to correct them.

othercorey:
Closing this with the documentation changes.

PR #737, opened 2021-02-17
Correct the template override docs.

markstory:
Applications and themes have different paths for overrides. Refs #736

### DIFF
--- a/docs/en/development.rst
+++ b/docs/en/development.rst
@@ -207,32 +207,36 @@ Creating a Bake Theme
 
 If you wish to modify the output produced by the "bake" command, you can
 create your own bake 'theme' which allows you to replace some or all of the
-templates that bake uses. The best way to do this is:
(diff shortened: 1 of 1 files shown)


othercorey:
What path/use did we leave this part of the path manipulation in for? There was one situation in either application override or plugin override that used this even after moving everything to use the `Bake.` prefix.

https://github.com/cakephp/bake/blob/master/src/View/BakeView.php#L143

markstory:
> What path/use did we leave this part of the path manipulation in for? There was one situation in either application override or plugin override that used this even after moving everything to use the Bake. prefix.

That looks like it is for bake theme plugins.

connect232 (commented):
Thanks for looking into this.

I think the directory should be templates/plugin/Bake/ taking away the 's' in plugins.

othercorey:
Fixed the plugin directory name in application override.

PR #746, opened 2021-04-19
Add Common Options to Plugin Command

### LINKED ISSUE #745: Plugin Command missing Common Options
The PluginTask in 1.12.x included the common options, but looks like it was dropped in the conversion to PluginCommand in 2.0.x.

```
~/workspace/php/croogo $ bin/cake bake plugin --theme Croogo/Core Agile
Error: Unknown option `theme`.
```

miketallroth:
Fixes #745

### DIFF
--- a/src/Command/PluginCommand.php
+++ b/src/Command/PluginCommand.php
@@ -369,6 +369,15 @@ class PluginCommand extends BakeCommand
         ])->addOption('composer', [
             'default' => ROOT . DS . 'composer.phar',
             'help' => 'The path to the composer executable.',
+        ])->addOption('force', [
+            'short' => 'f',
+            'boolean' => true,
+            'help' => 'Force overwriting existing files without prompting.',
+        ])->addOption('theme', [
+            'short' => 't',
+            'help' => 'The theme to use when baking code.',
+            'default' => Configure::read('Bake.theme') ?? '',
+            'choices' => $this->_getBakeThemes(),
         ]);
 
         return $parser;
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -115,6 +115,7 @@ class Plugin extends BasePlugin
      * @param string $path The path to look in.
      * @return string[]
      * @phpstan-return class-string<\Bake\Command\BakeCommand>[]
+     * @psalm-suppress InvalidReturnType
      */
     protected function findInPath(string $namespace, string $path): array
     {
@@ -152,6 +153,7 @@ class Plugin extends BasePlugin
             $candidates[$class::defaultName()] = $class;
         }
 
+        /** @psalm-suppress InvalidReturnStatement */
         return $candidates;
     }
 }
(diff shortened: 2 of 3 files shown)


src/Command/PluginCommand.php
```diff
@@ -361,6 +361,8 @@ public function findPath(array $pathOptions, ConsoleIo $io): void
      */
     public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
     {
+        $parser = $this->_setCommonOptions($parser);
```
markstory, line 364: Not all these options are relevant though. For example the `plugin` and `connection` options don't make sense when generating a plugin. Could we add only the theme option?

miketallroth: That works for me. Can I use removeOption() to take the two out we don't need? Otherwise, I'm cutting and pasting to build up $bakeThemes again, or I lose the choices feature.

miketallroth: I'll try refactoring CommonOptionsTrait a bit so I can call the $bakeThemes build from the PluginCommand. That'll keep it DRY and not require adding, then removing.

miketallroth: I've pushed the refactored code which still ends up with some cut and paste from CommonOptionsTrait. Not sure that's better than the alternative of calling setCommonOptions, then removing the two that don't apply.

Let me know your opinion, I don't mind doing it again.

markstory: Removing options works for me.

src/Command/PluginCommand.php
```diff
@@ -369,6 +369,15 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
...
             'help' => 'The path to the composer executable.',
+        ])->addOption('force', [
+            'short' => 'f',
+            'boolean' => true,
+            'help' => 'Force overwriting existing files without prompting.',
+        ])->addOption('theme', [
+            'short' => 't',
+            'help' => 'The theme to use when baking code.',
+            'default' => Configure::read('Bake.theme') ?? '',
+            'choices' => $this->_getBakeThemes(),
```
markstory, line 380: This looks good. Seems like there is one small style issue that needs to be fixed and this is would be good to go.

miketallroth: Updated for my style issue. Still another style issue in the plugin, but not related to this change.

miketallroth: Style issues all resolved / suppressed.

I have another issue that relates to baking plugins, and requires a new option. Should I add that code to this PR, or create a new one? Sorry for bouncing around, I was planning on making that a separate PR because its different functionality, and a different reason, but it's basically all the same code.

New option 'modify' (default to true), to modify the Application.php file. Croogo CMS uses its own Plugin loader, so we don't want the Application file modified automatically. This option would allow us to turn off that functionality.

ADmad: You can open a separate PR for that.

PR #749, opened 2021-04-23
Update version constraint for phpunit in baked composer.json

ADmad:
This matches the current phpunit version constraint of `cakephp/cakephp`.


othercorey:
Looks ok to me.

PR #751, opened 2021-04-27
Add new helper method for exporting vars.

ADmad:
The existing `BakeHelper::stringifyList()` is very simplistic
and only supports an array of strings, which requires crafting
config arrays in a specific way which is quite cumbersome.

### DIFF
--- a/src/View/Helper/BakeHelper.php
+++ b/src/View/Helper/BakeHelper.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 namespace Bake\View\Helper;
 
 use Bake\Utility\Model\AssociationFilter;
+use Brick\VarExporter\VarExporter;
 use Cake\Core\Configure;
 use Cake\Core\ConventionsTrait;
 use Cake\Database\Schema\TableSchema;
@@ -64,6 +65,7 @@ class BakeHelper extends Helper
      * @param array $list array of items to be stringified
      * @param array $options options to use
      * @return string
+     * @deprecated 2.5.0 Use BakeHelper::exportVar() instead.
      */
     public function stringifyList(array $list, array $options = []): string
     {
@@ -123,6 +125,24 @@ class BakeHelper extends Helper
         }
 
         return $start . implode($join, $list) . $end;
+    }
+
+    /**
+     * Export variable to string representation.
+     *
+     * (Similar to `var_export()` but better).
+     *
+     * @param mixed $var Variable to export.
+     * @param int $indentLevel Identation level.
+     * @param int $options Exporting options.
+     * @return string
+     * @see https://github.com/brick/varexporter#options
+     */
+    public function exportVar($var, int $indentLevel = 0, int $options = 0): string
+    {
+        $options |= VarExporter::TRAILING_COMMA_IN_ARRAY;
+
+        return VarExporter::export($var, $options, $indentLevel);
     }
 
     /**
@@ -362,7 +382,7 @@ class BakeHelper extends Helper
      *
      * @param string[]|false|null $fields Fields list.
      * @param string[]|null $primaryKey Primary key.
-     * @return string[]
+     * @return array<string, bool>
      */
     public function getFieldAccessibility($fields = null, $primaryKey = null): array
     {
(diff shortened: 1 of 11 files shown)


ADmad:
The question that remains is are we okay releasing the change as a new minor or we should bump to a new major, since these changes would cause problems to anyone who has extended the commands (as I have).

markstory (approved):
This seems fine for a minor release. We'll need to keep stringifyList() around for a long time as it will be used by userland templates.

src/View/Helper/BakeHelper.php
```diff
@@ -64,6 +65,7 @@ public function arrayProperty(string $name, array $value = [], array $options =
      * @param array $list array of items to be stringified
      * @param array $options options to use
      * @return string
+     * @deprecated 2.5.0 Use BakeHelper::exportVar() instead.
```
othercorey, line 68: Will it be hard to switch to exportVar()? I think we want to add deprecationWarning() at the same time in most cases.

templates/bake/Model/table.twig
```diff
@@ -45,7 +45,7 @@ class {{ name }}Table extends Table
 
 {%- if primaryKey %}
     {%- if primaryKey is iterable and primaryKey|length > 1 %}
-        $this->setPrimaryKey([{{ Bake.stringifyList(primaryKey, {'indent': false})|raw }}]);
+        $this->setPrimaryKey({{ Bake.exportVar(primaryKey, 0, constant('Brick\\VarExporter\\VarExporter::INLINE_NUMERIC_SCALAR_ARRAY'))|raw }});
```
othercorey, line 48: Is there any way we can make this config option easier to use? Not a great example for people to follow.

src/View/Helper/BakeHelper.php
ADmad: It would be actually easier to change all the funky string array to regular ones :)

templates/bake/Model/table.twig
ADmad: Yes it's not pretty but the `$options` arg of `VarExporter::export()` is a bitmask and using constants is better than arbitrary numbers. I am open to suggestions to make it easier / more readable.

othercorey: We could add globals for the config constants.

othercorey: Another idea would be supporting string arguments and convert them to constants in exportVar().

dereuromark: Couldnt it default to a reasonable default in the cake wrapper method?

ADmad: I added a new `exportArray()` with required defaults for inlined arrays.

dereuromark (approved):
I would have inverted it to make it more human ~~reasonable~~readable.
```php
       $options = 0;
        if ($inline) {
            $options = VarExporter::INLINE_NUMERIC_SCALAR_ARRAY;
```
but no big deal

ADmad:
@dereuromark Yup that's better, done.

src/View/Helper/BakeHelper.php
```diff
@@ -125,6 +127,44 @@ public function stringifyList(array $list, array $options = []): string
...
     }
 
+    /**
+     * Export variable to string representation.
+     *
+     * (Similar to `var_export()` but better).
+     *
+     * @param mixed $var Variable to export.
+     * @param int $indentLevel Identation level.
+     * @param int $options Exporting options.
```
othercorey, line 137:
```suggestion
     * @param int $options VarExporter option flags
```

src/View/Helper/BakeHelper.php
```diff
@@ -125,6 +127,44 @@ public function stringifyList(array $list, array $options = []): string
...
+    }
+
+    /**
+     * Export array to string representation.
+     *
+     * (Similar to `var_export()` but better).
+     *
+     * @param array $var Array to export.
+     * @param int $indentLevel Identation level.
+     * @param bool $inline Inline numeric scalar array.
```
othercorey, line 155:
```suggestion
     * @param bool $inline Inline numeric scalar array (adds INLINE_NUMERIC_SCALAR_ARRAY flag)
```

templates/bake/Model/table.twig
othercorey: I don't know anything about the options but definitely looks better now.

ADmad:
Should I add explicit deprecation warning in `BakeHelper::stringifyList()`?

dereuromark:
I wouldn't. We dont know how well this new method works with custom uses of the existing method
And if switching here is even an option in all those cases.

IMO it doesn't hurt to keep the other one around until some major. And then removing it is still easy enough, no?

ADmad:
Works for me.

PR #752, opened 2021-04-28
Update use of deprecated method.

ADmad:
Refs #751.

### DIFF
--- a/tests/comparisons/Model/testBakeTableValidation.php
+++ b/tests/comparisons/Model/testBakeTableValidation.php
@@ -80,7 +80,10 @@ class TestBakeArticlesTable extends Table
             ->allowEmptyString('email');
 
         $validator
-            ->uploadedFile('image', ['optional' => true, 'types' => ['image/jpeg']])
+            ->uploadedFile('image', [
+                'optional' => true,
+                'types' => ['image/jpeg'],
+            ])
             ->allowEmptyFile('image');
 
         return $validator;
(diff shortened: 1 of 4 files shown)


tests/comparisons/Model/testBakeTableValidation.php
```diff
@@ -80,7 +80,10 @@ public function validationDefault(Validator $validator): Validator
             ->allowEmptyString('email');
 
         $validator
-            ->uploadedFile('image', ['optional' => true, 'types' => ['image/jpeg']])
+            ->uploadedFile('image', [
```
ADmad, line 83: Currently `VarExporter` doesn't have an option to force generation of a single line representation of associative array. I'll see if I can they would be interested in adding such an option.

othercorey: They might not want to support things like "maximum line length" which people might ask for with that.

ADmad: Yeah the that might not be feasible, so inline associative arrays should be done only when it's know beforehand that the result won't exceed "max line length".

markstory: Multi line arrays are fine though. I don't think we need to stick to single line output.

ADmad: Yup, my initial comment was more of an informative one just to clarify why the output changed :slightly_smiling_face:.

PR #753, opened 2021-04-28
Add model validation generation test.

ADmad:
This test was lost when bake tasks were converted to commands for CakePHP 4.x.
Without the `ModelCommandTest::testBakeTableValidation()` test the comparison file
`testBakeTableValidation.php` was unused.


ADmad:
Merging as I need this test for #752.

Issue #755, opened 2021-05-05
Annotate `@uses` on test method level


mentisy:
Previously discussed here: 
#577

Back then `@uses` was added at test controller class level. This is useful, but personally I'd like to also have them at method level. This way I can avoid having IDE yell at me about unused elements (methods etc.). This scenario occurs especially with controller actions and table find methods. If you only call finder methods through the ORM `find(string $type = 'all'...)` method, the custom methods won't have been called according to the IDE.

I'm not sure if @dereuromark covers this in his [cakephp-ide-helper](https://github.com/dereuromark/cakephp-ide-helper), maybe [PR #135](https://github.com/dereuromark/cakephp-ide-helper/pull/135), but adding it to the plugin code could be useful.

I have forked this repo and added a branch that adds `@uses \{{class}}::{{method}}()` to all test methods that bake generates. I will happily create a PR from my fork if this is something you'd like to implement.

Fork: https://github.com/cakephp/bake/compare/master...mentisy:uses-docblocks

Example of PostsController and index action:
```
/**
 * Test index method
 *
 * @return void
 * @uses \Bake\Test\App\Controller\PostsController::index()
 */
public function testIndex(): void
{
    $this->markTestIncomplete('Not implemented yet.');
}
```

dereuromark:
Feel free to open the PR to discuss the details.

Yes, the IdeHelper already syncs it for existing code
But you are absolutely right that this should also be added to new files using bake plugin :)

PR #756, opened 2021-05-05
Add `@uses` annotation to method level tests.

mentisy:
This PR adds `@uses` annotations to test methods to indicate which method is being tested, and makes navigation to the code being tested easier. This is especially useful for ORM `find` methods, Command `execute` method and controller actions.

Example of the `index` action on `PostsController`:
```php
/**
 * Test index method
 *
 * @return void
 * @uses \Bake\Test\App\Controller\PostsController::index()
 */
public function testIndex(): void
{
}
```

Solves #755

### DIFF
--- a/tests/TestCase/Command/AllCommandTest.php
+++ b/tests/TestCase/Command/AllCommandTest.php
@@ -188,7 +188,7 @@ class AllCommandTest extends TestCase
     }
 
     /**
-     * Test docblock \@\uses generated for test methods
+     * Test docblock @uses generated for test methods
      *
      * @return void
      */
--- a/tests/TestCase/Command/AllCommandTest.php
+++ b/tests/TestCase/Command/AllCommandTest.php
@@ -186,4 +186,46 @@ class AllCommandTest extends TestCase
         );
         $this->assertOutputContains('Bake All complete');
     }
+
+    /**
+     * Test docblock \@\uses generated for test methods
+     *
+     * @return void
+     */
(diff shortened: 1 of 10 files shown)


mentisy:
The https://github.com/cakephp/bake/pull/756/commits/ec16bed36d6bb5e90df1124ee3c7e1fa5fe621a6 commit was because my IDE unfortunately had `Optimize imports` enabled for commits and I noticed in while looking at the diff

tests/TestCase/Command/AllCommandTest.php
```diff
@@ -186,4 +186,46 @@ public function testExecuteWithPrefix()
         );
         $this->assertOutputContains('Bake All complete');
     }
+
+    /**
+     * Test docblock \@\uses generated for test methods
```
markstory, line 191: Why are there \ in this?

mentisy: My IDE tried to annotate the word following the `@uses`. Codesniffer doesn't care apparently, so I can remove the slashes. Yes?

mentisy: Found other instances of `@uses` in code when not actually annotating. So I pushes a commit removing them.

markstory:
Thank you 👍

dereuromark:
This is primarily useful for controller and command testing, as there the tests are involved indirectly.
But I guess it doesn't hurt to have it for all classes.

PR #757, opened 2021-06-10
Update plugin creation for 4.3

markstory:
These change would go into 2.next branch to be release alongside CakePHP 4.3

### DIFF
--- a/templates/bake/Plugin/tests/bootstrap.php.twig
+++ b/templates/bake/Plugin/tests/bootstrap.php.twig
@@ -55,3 +55,18 @@ if (file_exists($root . '/config/bootstrap.php')) {
 
     return;
 }
+
+/**
+ * Load schema from a SQL dump file.
+ *
+ * If your plugin does not use database fixtures you can
+ * safely delete this.
+ *
+ * If you want to support multiple databases, consider
+ * using migrations to provide schema for your plugin,
+ * and using \Migrations\TestSuite\Migrator to load schema.
+ */
+use Cake\TestSuite\Schema\SchemaManager;
+
+// Load a schema dump file.
+SchemaManager::create('test', './tests/schema.sql');
--- a/templates/bake/Plugin/tests/bootstrap.php.twig
+++ b/templates/bake/Plugin/tests/bootstrap.php.twig
@@ -69,4 +69,4 @@ if (file_exists($root . '/config/bootstrap.php')) {
 use Cake\TestSuite\Schema\SchemaManager;
 
 // Load a schema dump file.
-SchemaManager::create('test', './tests/schema.sql');
+SchemaManager::create('test', 'tests/schema.sql');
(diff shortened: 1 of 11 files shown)


templates/bake/Plugin/tests/bootstrap.php.twig
```diff
@@ -55,3 +55,18 @@ if (file_exists($root . '/config/bootstrap.php')) {
...
+
+/**
+ * Load schema from a SQL dump file.
+ *
+ * If your plugin does not use database fixtures you can
+ * safely delete this.
+ *
+ * If you want to support multiple databases, consider
+ * using migrations to provide schema for your plugin,
+ * and using \Migrations\TestSuite\Migrator to load schema.
```
markstory, line 67: I contemplated adding an option to include database fixture setup which would include:

* Adding migrations as dev-dep
* Using the migrator instead.

If folks prefer this approach I can do that as a follow up.

tests/comparisons/Plugin/SimpleExample/tests/bootstrap.php
```diff
@@ -38,3 +38,18 @@
...
+ * safely delete this.
+ *
+ * If you want to support multiple databases, consider
+ * using migrations to provide schema for your plugin,
+ * and using \Migrations\TestSuite\Migrator to load schema.
+ */
+use Cake\TestSuite\Schema\SchemaManager;
+
+// Load a schema dump file.
+SchemaManager::create('test', './tests/schema.sql');
```
othercorey, line 55:
```suggestion
SchemaManager::create('test', 'tests/schema.sql');
```
Is the `./` required for any sort of security? I think we should just use relative paths from base like most places.

tests/comparisons/Plugin/Company/Example/tests/bootstrap.php
```diff
@@ -38,3 +38,18 @@
...
+ * safely delete this.
+ *
+ * If you want to support multiple databases, consider
+ * using migrations to provide schema for your plugin,
+ * and using \Migrations\TestSuite\Migrator to load schema.
+ */
+use Cake\TestSuite\Schema\SchemaManager;
+
+// Load a schema dump file.
+SchemaManager::create('test', './tests/schema.sql');
```
othercorey, line 55:
```suggestion
SchemaManager::create('test', 'tests/schema.sql');
```

templates/bake/Plugin/tests/bootstrap.php.twig
othercorey: I'm not so sure about the name `Migrator` after seeing it in docs.

markstory: We can still rename it if you have a more appropriate name in mind.

templates/bake/Plugin/tests/bootstrap.php.twig
```diff
@@ -55,3 +55,18 @@ if (file_exists($root . '/config/bootstrap.php')) {
...
+ */
+use Cake\TestSuite\Schema\SchemaManager;
+
+// Load a schema dump file.
+SchemaManager::create('test', './tests/schema.sql');
```
othercorey, line 72:
```suggestion
SchemaManager::create('test', 'tests/schema.sql');
```

Issue #759, opened 2021-07-24
Missing relationship docblock variables on `add` and `edit` templates


mentisy:
When an `add` or `edit` template is generated, there's a docblock on top describing the outside variables, like `AppView` and the related entity.

## How it is:

It can look like this:
```php
<?php
/**
 * @var \Bake\Test\App\View\AppView $this
 * @var \App\Model\Entity\Author $author
 */
?>
```

When looking down at the generated form, you could have these form inputs:
```php
echo $this->Form->control('role_id', ['options' => $roles]);
echo $this->Form->control('name');
echo $this->Form->control('description');
```

In my IDE (PHPStorm), I will now get yelled at stating that `$roles` is undefined.

-----------------

## How it should be:

The Bake template should add these relationship variables to the generated template, so it looks like this

```php
<?php
/**
 * @var \Bake\Test\App\View\AppView $this
 * @var \App\Model\Entity\Author $author
 * @var array $roles
 */
?>
```

Now `$roles` is defined as an array, which it will be as the select box uses an array of key/value pairs.

--------------------------

## Suggested changes to fix issue

 I created a fork and a branch over here with suggested changed: https://github.com/cakephp/bake/compare/master...mentisy:association-variables-add-edit-templates?expand=1

othercorey:
Can you create a PR with your changes?

mentisy:
@othercorey See #761

PR #761, opened 2021-07-24
Add association variables to add and edit templates

mentisy:
Adds proper docblock variables for associations in `add` and `edit`templates.

See #759

### DIFF
--- a/templates/bake/Template/add.twig
+++ b/templates/bake/Template/add.twig
@@ -17,6 +17,19 @@
 /**
  * @var \{{ namespace }}\View\AppView $this
  * @var \{{ entityClass }} ${{ singularVar }}
+        {{- "\n" }}
+{%- if associations.BelongsTo is defined %}
+    {%- for assocName, assocData in associations.BelongsTo %}
+ * @var array ${{ assocData.variable }}
+        {{- "\n" }}
+    {%- endfor %}
+{% endif %}
+{%- if associations.BelongsToMany is defined %}
+    {%- for assocName, assocData in associations.BelongsToMany %}
+ * @var array ${{ assocData.variable }}
+        {{- "\n" }}
+    {%- endfor %}
+{% endif %}
  */
 ?>
 {{ element('Bake.form') }}
--- a/templates/bake/Template/edit.twig
+++ b/templates/bake/Template/edit.twig
@@ -17,6 +17,19 @@
 /**
  * @var \{{ namespace }}\View\AppView $this
  * @var \{{ entityClass }} ${{ singularVar }}
+        {{- "\n" }}
+{%- if associations.BelongsTo is defined %}
+    {%- for assocName, assocData in associations.BelongsTo %}
+ * @var array ${{ assocData.variable }}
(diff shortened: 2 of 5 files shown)


dereuromark:
I think IdeHelper does this already afterwards. But doesnt hurt to generate it right away 
👍

markstory:
Thank you 👍

templates/bake/Template/add.twig
```diff
@@ -17,6 +17,19 @@
 /**
  * @var \{{ namespace }}\View\AppView $this
  * @var \{{ entityClass }} ${{ singularVar }}
+        {{- "\n" }}
+{%- if associations.BelongsTo is defined %}
+    {%- for assocName, assocData in associations.BelongsTo %}
+ * @var array ${{ assocData.variable }}
```
ADmad, line 23: Lies!! This var is a result of `find('list')` which is a `Query` instance, not an array. So the type should be `iterable`.

dereuromark: The concrete type isnt that important usually just the existence:)
Could usually also be string[] etc
As those are used for options.

dereuromark: The concrete type isnt that important usually just the existence:)
Could usually also be string[] etc
As those are used for options.

ADmad: > The concrete type isnt that important...

I disagree, incorrect types are misleading and cause confusion for new users.

mentisy: @ADmad : I was thinking about this when adding the variable. On the index templates, you have this docblock:
```php
/**
 * @var \App\View\AppView $this
 * @var \App\Model\Entity\File[]|\Cake\Collection\CollectionInterface $files
 */
```

We could do something similar with add/edit.

```php
/**
 * @var \App\View\AppView $this
 * @var string[]|\Cake\Collection\CollectionInterface $files
 */
```

What do you think? I can create a PR for this if you'd like.

ADmad: `\Cake\Collection\CollectionInterface<string>` would be the most accurate but if that's not supported by IDEs(PhpStorm) yet, then `string[]|\Cake\Collection\CollectionInterface` is fine.

othercorey: Please use `\Cake\Collection\CollectionInterface|array<string>` or `\Cake\Collection\CollectionInterface|string[]` to follow our new standard.

mentisy: Even if it was, wouldn't it be a good idea to follow the conventions in other templates until all templates are changed?

I couldn't get it to work with PhpStorm 2021.1.4 at least.

While `\Cake\Collection\CollectionInterface<string>` doesn't throw any errors/warnings, say we tried `\Cake\Collection\CollectionInterface<\Cake\Datasource\EntityInterface>` for index it wouldn't suggest any properties/methods in `EntityInterface` when looping through it. So it doesn't seem to be fully supported yet.

I say we go with the `string[]|\Cake\Collection\CollectionInterface` way.

PR #762, opened 2021-07-25
Add more specific docblock typehinting for `add` and `edit` templates

mentisy:
See conversation #761.

This PR adds a more specific typehint to association variables in the `add` and `edit` templates.

While @othercorey suggested `\Cake\Collection\CollectionInterface|string[]`, I used `string[]|\Cake\Collection\CollectionInterface` as to match the order used in `index` template.


markstory:
Thank you 👍

PR #765, opened 2021-09-03
Make field generation in fixtures optional

markstory:
New style fixtures don't require the $fields property anymore so we can skip generating it by default. Users can use the `--fields` option to generate fixtures compatible with the old fixture system.


dereuromark:
Existing apps still require them, though
And by default those would not work anymore now with the current way of them being baked

I wonder if this could be better autodetected somehow

markstory:
> I wonder if this could be better autodetected somehow

Perhaps, we would need additional options to override the autodetection though. I was aiming to set the defaults up so that new applications and applications that upgraded would be the 'default'.

PR #769, opened 2021-09-27
feat: Add stub for console() to bake templates

rkaiser0324:
Using `console()` is described in [the Book](https://book.cakephp.org/4/en/plugins.html#commands) (although it's missing the `CommandCollection` return type hint there; I'll open a PR in that repo to update that).  Since this method wasn't added as part of `bake` or the examples here, it took awhile to track down how to use it in practice.  This PR adds it to all 3 locations.

### DIFF
--- a/templates/bake/Plugin/src/Plugin.php.twig
+++ b/templates/bake/Plugin/src/Plugin.php.twig
@@ -22,6 +22,7 @@ use Cake\Core\BasePlugin;
 use Cake\Core\PluginApplicationInterface;
 use Cake\Http\MiddlewareQueue;
 use Cake\Routing\RouteBuilder;
+use Cake\Console\CommandCollection;
 
 /**
  * Plugin for {{ namespace }}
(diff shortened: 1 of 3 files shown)


templates/bake/Plugin/src/Plugin.php.twig
```diff
@@ -76,4 +77,17 @@ class Plugin extends BasePlugin
...
+
+    /**
+     * Add commands for the plugin.
+     *
+     * @param \Cake\Console\CommandCollection $commands The command collection to update.
+     * @return \Cake\Console\CommandCollection
+     */
+    public function console(CommandCollection $commands) : CommandCollection
+    {
+        // Add your commands here
```
markstory, line 89: Should we have the parent call here so that plugins continue to autodiscover commands?

rkaiser0324: Good idea; I added that as well.

PR #771, opened 2021-10-02
Disable and clear connection metadata cache before baking model

othercorey:
Closes https://github.com/cakephp/bake/issues/724

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -16,12 +16,14 @@ declare(strict_types=1);
  */
 namespace Bake\Command;
 
+use Bake\Utility\SubsetSchemaCollection;
 use Bake\Utility\TableScanner;
 use Bake\Utility\TemplateRenderer;
 use Cake\Console\Arguments;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Core\Configure;
+use Cake\Database\Connection;
 use Cake\Database\Exception;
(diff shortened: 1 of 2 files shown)


othercorey:
@ADmad Do you think this will clear the cache properly?

ADmad:
Disabling schema cache will probably make it just not check it (which is fine as far as bake goes), but it won't clear it. When using the generated model it would most likely still use the stale schema cache.

othercorey:
> Disabling schema cache will probably make it just not check it (which is fine as far as bake goes), but it won't clear it. When using the generated model it would most likely still use the stale schema cache.

Right, sorry. I meant will this avoid the cache problem. I was thinking the user would only bake locally, but they're probably testing on the same instance.

src/Command/ModelCommand.php
```diff
@@ -78,6 +80,16 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
             return static::CODE_SUCCESS;
         }
 
+        // Disable caching before baking with connection
+        $connection = ConnectionManager::get($this->connection);
+        if ($connection instanceof Connection) {
+            $collection = $connection->getSchemaCollection();
+            if (!$collection instanceof SubsetSchemaCollection) {
+                $connection->cacheMetadata(false);
+                $connection->getCacher()->clear();
```
othercorey, line 89: Added call to clear cache explicitly.

markstory (commented):
This looks like it would work to me.

Theaxiom:
Thank you! 👍

Issue #717, opened 2020-10-23
Do not bake password or hidden fields


dereuromark:
Right now, bake template also adds hidden fields and the field password into index and view templates:

> Password | $2y$10$gB6bNUibyAgahzdnm7ZVje2/SZAmzYnOJher2xdSorDgCPUwcGfwx

I think we should exclude the list of hidden fields and the specific field "password" here from being auto generated into the templates.

    /**
     * Fields that are excluded from JSON versions of the entity.
     *
     * @var array
     */
    protected $_hidden = [
        'password',
    ];

What do you think?

Personally, I always display a boolean value instead (Yes, No) as "Is Password Set" for password field.
But I think people can still easily add that part back in.

markstory:
I think omitting fields that are hidden in the entities makes sense.

PR #773, opened 2021-10-03
Exclude hidden fields from index and view template generation

othercorey:
Closes https://github.com/cakephp/bake/issues/717

Defaults to the same hidden fields as ModelCommand if the entity class is empty.

### DIFF
--- a/src/Command/TemplateCommand.php
+++ b/src/Command/TemplateCommand.php
@@ -65,6 +65,13 @@ class TemplateCommand extends BakeCommand
     public $scaffoldActions = ['index', 'view', 'add', 'edit'];
 
     /**
+     * Actions that exclude hidden fields
+     *
+     * @var string[]
+     */
+    public $excludeHiddenActions = ['index', 'view'];
+
+    /**
      * AssociationFilter utility
      *
      * @var \Bake\Utility\Model\AssociationFilter|null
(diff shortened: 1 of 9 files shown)


src/Command/TemplateCommand.php
```diff
@@ -278,6 +285,7 @@ protected function _loadController(ConsoleIo $io): array
             $singularHumanName = $this->_singularHumanName($this->controllerName);
             $schema = $modelObject->getSchema();
             $fields = $schema->columns();
+            $hidden = $modelObject->newEmptyEntity()->getHidden() ?: ['token', 'password', 'passwd'];
```
markstory, line 288: Should we have a constant for the list of fields hidden by default? That would prevent divergence in the future.

markstory (approved):
Nice work!

src/Command/TemplateCommand.php
othercorey: I was just going to rely on the Entity class, but we don't have them in the test suite. Should we exclude anything if the user doesn't have an entity class defined yet?

markstory: > Should we exclude anything if the user doesn't have an entity class defined yet?

You currently have the entity taking ownership of the hidden fields list which is :+1:. We can, and should provide reasonable fallbacks if the developer hasn't defined any fields as hidden. The scenario where a developer willingly wants to show `password` will be exceptionally rare, and defaulting to keeping the developer from making a big mistakes is a defensible position for us to take.

Issue #758, opened 2021-07-22
Baking fails for tables with a numeric field name.


VARGA-Peter:
For a table having a field name which is pure numeric [`1` for example] baking fails:

```
Error: [TypeError] Cake\Database\Schema\TableSchema::getColumn():
Argument cakephp/cakephp#1 ($name) must be of type string, int given, called in
.../vendor/cakephp/bake/src/Command/FixtureCommand.php on line 254 in
.../vendor/cakephp/cakephp/src/Database/Schema/TableSchema.php on line 351

Error: [TypeError] preg_match(): Argument cakephp/cakephp#2 ($subject) must be of type string, int given in
.../vendor/cakephp/bake/src/Command/ModelCommand.php on line 301
```

This table doesn't follow **ANY of** the CakePHP table naming conventions [it's a 3rd party product which tables are needed in the CakePHP project and read by manually created `SELECT` statements].

My opinion is that the baking should be stable as the table/field name can be called what ever a programmer wants using the backticks.

May be the CakePHP developers reconsider it and fix this _bagatelle_.

_My personal note_: I think it's unbelievable _stupid_ to use a number as a field name but the world isn't perfect.

VARGA-Peter:
> Do you have identifier quoting enabled for your connection? Without that backticks will not be used.

Please advise. How should I enable the quoting for the `bake` script? In the project there is no problem.

markstory:
https://book.cakephp.org/4/en/orm/database-basics.html#identifier-quoting

If the columns are working ok in the application code, there might just need to be some additional casts added to make column names into strings as they are coming back as integers.

VARGA-Peter:
> https://book.cakephp.org/4/en/orm/database-basics.html#identifier-quoting
> 
> If the columns are working ok in the application code, there might just need to be some additional casts added to make column names into strings as they are coming back as integers.

I found the quoting documentation earlier but I don't understand how should I enable it for a connection in the baking script as it's per connection. Do I misunderstand something?

I wrote

_[it's a 3rd party product which tables are needed in the CakePHP project and read by manually created SELECT statements]_

so there is no need for quoting.

markstory:
> I found the quoting documentation earlier but I don't understand how should I enable it for a connection in the baking script as it's per connection. Do I misunderstand something?

[...]

PR #774, opened 2021-10-07
Validate column names before baking model or fixture

othercorey:
Closes https://github.com/cakephp/bake/issues/758

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -16,15 +16,16 @@ declare(strict_types=1);
  */
 namespace Bake\Command;
 
-use Bake\Utility\SubsetSchemaCollection;
 use Bake\Utility\TableScanner;
 use Bake\Utility\TemplateRenderer;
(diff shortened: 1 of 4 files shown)


src/Command/ModelCommand.php
```diff
@@ -84,9 +85,9 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
         $connection = ConnectionManager::get($this->connection);
         if ($connection instanceof Connection) {
             $collection = $connection->getSchemaCollection();
-            if (!$collection instanceof SubsetSchemaCollection) {
-                $connection->cacheMetadata(false);
+            if ($collection instanceof CachedCollection) {
                 $connection->getCacher()->clear();
+                $connection->cacheMetadata(false);
```
othercorey, line 90: Calling `clear()` if the model cache doesn't exist throws an exception. Instead we check if the cache is used based on the collection type.

Clear before disable cache.

PR #779, opened 2021-10-10
1.x: Make email field unique only for Users and Accounts

othercorey:
Closes https://github.com/cakephp/bake/issues/689

### DIFF
--- a/src/Shell/Task/ModelTask.php
+++ b/src/Shell/Task/ModelTask.php
@@ -842,9 +842,14 @@ class ModelTask extends BakeTask
             return [];
         }
 
+        $uniqueColumns = ['username', 'login'];
+        if (in_array($model->getAlias(), ['Users', 'Accounts'])) {
+            $uniqueColumns[] = 'email';
+        }
+
         $rules = [];
         foreach ($fields as $fieldName) {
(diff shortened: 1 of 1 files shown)


othercorey:
Local tests pass.

src/Shell/Task/ModelTask.php
```diff
@@ -842,9 +842,14 @@ public function getRules($model, array $associations)
             return [];
         }
 
+        $uniqueColumns = ['username', 'login'];
+        if (in_array($model->getAlias(), ['Users', 'Accounts'])) {
```
ADmad, line 846: Wouldn't a better way would be to check for unique constraint for the field in the schema?

othercorey: This is just backporting the snippet from Bake 2.x to satisfy the issue in Bake 1.x.

ADmad: ok

othercorey: The logic you're looking for exists below this.

othercorey: If this isn't worth it for 1.x, then we can just close this.

othercorey:
Closing due to no interest.

Issue #197, opened 2015-12-11
2.0.0 Bake should detect composite unique constraints


htstudios:
Possible BC breaks.
Ref: https://github.com/cakephp/bake/pull/196

PR #780, opened 2021-10-10
Add support for multiple field unique constraints

othercorey:
Closes https://github.com/cakephp/bake/issues/197
Refs https://github.com/cakephp/bake/issues/34

This adds a `fields` key to the rule if there are multiple fields. We could always add `fields` if the rule expects an array.

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -887,10 +887,7 @@ class ModelCommand extends BakeCommand
             if ($constraint['type'] !== TableSchema::CONSTRAINT_UNIQUE) {
                 continue;
             }
-            if (count($constraint['columns']) > 1) {
-                continue;
-            }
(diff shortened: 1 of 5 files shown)


src/Command/ModelCommand.php
```diff
@@ -887,10 +887,7 @@ public function getRules(Table $model, array $associations, Arguments $args): ar
             if ($constraint['type'] !== TableSchema::CONSTRAINT_UNIQUE) {
                 continue;
             }
-            if (count($constraint['columns']) > 1) {
-                continue;
-            }
-            $rules[$constraint['columns'][0]] = ['name' => 'isUnique'];
+            $rules[$constraint['columns'][0]] = ['name' => 'isUnique', 'fields' => $constraint['columns']];
```
markstory, line 890: This will be totally backwards compatible. Nicely done :clap:

PR #783, opened 2021-10-12
Add support for multiple field unique validation

othercorey:
Closes https://github.com/cakephp/bake/issues/34

This implements multi-column unique index supports that matches the support added in application rules: https://github.com/cakephp/bake/pull/780

A unique rule is added for each field in a multi-field constraint. I'm not sure if we need to validate both "sets" of fields separately. The IsUnique rule should probably fail properly if the field isn't set in the entity and the column isn't nullable in the schema.

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -798,7 +798,7 @@ class ModelCommand extends BakeCommand
 
         foreach ($schema->constraints() as $constraint) {
             $constraint = $schema->getConstraint($constraint);
-            if (!in_array($fieldName, $constraint['columns'], true) || count($constraint['columns']) > 1) {
+            if (!in_array($fieldName, $constraint['columns'], true)) {
                 continue;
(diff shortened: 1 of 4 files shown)


othercorey:
This generates a call to Table::validateUnique() with `$options` set to `['scope' => [other fields]]`.

The scope is combined with the original validation field and then passed to IsUnique as a multi-field rule:

```php
        $fields = array_merge(
            [$context['field']],
            isset($options['scope']) ? (array)$options['scope'] : []
        );
        $values = $entity->extract($fields);
        foreach ($values as $field) {
            if ($field !== null && !is_scalar($field)) {
                return false;
            }
        }
        $class = static::IS_UNIQUE_CLASS;
        /** @var \Cake\ORM\Rule\IsUnique $rule */
        $rule = new $class($fields, $options);
```

othercorey:
Closing due to no interest.

PR #784, opened 2021-10-15
Add allowMultipleNulls to unique rules for nullable fields

othercorey:
This adds the appropriate default for databases where unique indexes allow multiple nulls.

`allowMultipleNulls` is set to true for all non-sql server dbs.

### DIFF
--- a/templates/bake/Model/table.twig
+++ b/templates/bake/Model/table.twig
@@ -120,7 +120,11 @@ class {{ name }}Table extends Table
     {
 {% for field, rule in rulesChecker %}
 {% set fields = rule.fields is defined ? Bake.exportArray(rule.fields) : Bake.exportVar(field) %}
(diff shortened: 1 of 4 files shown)


templates/bake/Model/table.twig
```diff
@@ -120,7 +120,11 @@ class {{ name }}Table extends Table
     {
 {% for field, rule in rulesChecker %}
 {% set fields = rule.fields is defined ? Bake.exportArray(rule.fields) : Bake.exportVar(field) %}
-        $rules->add($rules->{{ rule.name }}({{ fields|raw }}{{ (rule.extra is defined and rule.extra ? (", '#{rule.extra}'") : '')|raw }}), ['errorField' => '{{ field }}']);
+{% set options = '' %}
+{% for optionName, optionValue in rule.options %}
+    {%~ set options = (loop.first ? '[' : options) ~ "'#{optionName}' => " ~ Bake.exportVar(optionValue) ~ (loop.last ? ']' : ', ') %}
+{% endfor %}
```
othercorey, line 126: VarExporter doesn't support inline associative array so this is a simple custom exporter just for options.

https://github.com/brick/varexporter/pull/20

PR #786, opened 2021-10-27
remove parent::buildOptionParser()

saeideng:
I think calling parent is redundant
I will fix testCases

### DIFF
--- a/templates/bake/Command/command.twig
+++ b/templates/bake/Command/command.twig
@@ -37,7 +37,6 @@ class {{ name }}Command extends Command
      */
     public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
     {
-        $parser = parent::buildOptionParser($parser);
 
         return $parser;
     }


dereuromark:
Dont we always do that? this makes it easier to extend by adding something in between, you never know what the parent added.

saeideng:
> Dont we always do that?

always ? :thinking:  nope 
```php
use Cake\Command\Command;
class MyCommand extends Command
```
`MyCommand` first extends `Cake\Command\Command` that has not `buildOptionParser()` and then 
 `Cake\Console\BaseCommand` 

`parent::buildOptionParser()` in baked commands refers to `Cake\Console\BaseCommand` 
```php
    protected function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
    {
        return $parser;
    }
```

dereuromark:
My point is:
It is the same as with events and other extension points.
Calling parent is just a good default as it will also take into consideration any addition there, at the point in time or later.
Finding the issues after the fact is much harder if we remove those IMO.

ADmad:
As stated by @dereuromark calling parent's method is a good default, there's no need to change that.

PR #787, opened 2021-10-27
Update testsuite to use new fixtures system.

ADmad:


othercorey:
Looks good to me. We should follow up by cleaning out the fixture fields unless we have to support users referencing bake fixtures.

ADmad:
> We should follow up by cleaning out the fixture fields unless we have to support users referencing bake fixtures.

It's a possibility, so let's just keep them for now.

othercorey:
I might have merged this too soon, but can fix in a follow up. The core fixtures that are referenced in tests should be copied into Bake since the schema is no longer defined by the fixture.

ADmad:
Ah right, I forgot that too.

PR #791, opened 2021-10-31
Allow column names to start with underscore

### LINKED ISSUE #790: StopException message not shown
Note: I don't know if this is a cakephp or bake bug.

When baking a fixture of tables that have columns name starting with underscore, it fails silently

```
$ bin/cake bake fixture ArosAcos
$
```

The exception is thrown in 

https://github.com/cakephp/bake/blob/cbeae91f35ee82656f76ee2f7aca9b0fba2c4768/src/Command/FixtureCommand.php#L220-L230

But the CommandRunner doesn't show the Exception Message

https://github.com/cakephp/cakephp/blob/101b179920a7b8ab68866554539d65c64c6bfd8d/src/Console/CommandRunner.php#L331-L338

I'm trying to bake fixture for [acl plugin](https://github.com/cakephp/acl/) that has a table `aros_acos` with `_create`, `_read`, `_update`, `_delete`columns.
Should the code be adapted to include such cases?

othercorey:
Fixes https://github.com/cakephp/bake/issues/790

### DIFF
--- a/src/Command/FixtureCommand.php
+++ b/src/Command/FixtureCommand.php
@@ -22,7 +22,6 @@ use Brick\VarExporter\VarExporter;
 use Cake\Console\Arguments;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
-use Cake\Console\Exception\StopException;
 use Cake\Core\Configure;
 use Cake\Database\Exception;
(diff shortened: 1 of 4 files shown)


raul338:
The code should also be in FixtureCommand (or move it to comply DRY)

PR #793, opened 2021-11-09
Persistent classes for Model / Entity

GrandFelix:
added support for persistent classes (Model, Entity). Use --persistent or --remove-persistent to create or remove persistent classes. https://github.com/cakephp/bake/pull/733 replaced with this PR. I did need to reset master branch so I can make some new PR-s.


dereuromark:
I feel like this requires some additional docs, as for most people (including me) it is not very clear what persistent here means, and what kind of behavior change that triggers.

GrandFelix:
> I feel like this requires some additional docs, as for most people (including me) it is not very clear what persistent here means, and what kind of behavior change that triggers.

If user runs `bake model table_name --persistent` it will generate `TableNamePersistentTable.php` which is generated only once! A lot of users run `bake model table_name` when they add new fields etc. and if they have some custom functions, custom validations etc. in those files everything get overwritten. With these feature you can add those functions, validations etc. to this persistent file. Everything is almost the same. The only difference is that if `--persistent` is used the *Table class extends this persistent class. With `--remove-persistent` the persistent class is removed.

dereuromark:
https://github.com/cakephp/bake/tree/2.x/docs/en

GrandFelix:
> https://github.com/cakephp/bake/tree/2.x/docs/en

Ohhh 🤦‍♂️🤦‍♂️ will do it.

templates/bake/Model/table.twig
```diff
@@ -33,8 +36,6 @@ class {{ name }}Table extends Table
      */
     public function initialize(array $config): void
     {
-        parent::initialize($config);
```
ADmad, line 36: Why is this parent call moved down?

tests/comparisons/Model/testBakeAssociationDetectionCategoriesProductsTable.php
```diff
@@ -5,8 +5,8 @@
 
 use Cake\ORM\Query;
 use Cake\ORM\RulesChecker;
-use Cake\ORM\Table;
 use Cake\Validation\Validator;
+use Cake\ORM\Table;
```
ADmad: The `use` statement need to remain in alphabetical order.

templates/bake/Model/table.twig
GrandFelix: with the intention that code from persistent class is always overwritten. so user can override code from main class also when he runs bake model again initialize code will overwrite generated code if needed.

ADmad:
This needs to target `2.next` instead of `2.x`.

GrandFelix:
@dereuromark about docs. Where I should put that? Because there is no section where any of possible arguments is explained. All that it is, is that user use bake with  `--help` where he can see all options. So where I should really put this?

dereuromark:
Let's wait with docs and more refactor until we have a clear consensus on this proposed changeset.

For me it is a bit weird that this only applies to the model layer and for example not other classes around it.
Wouldnt it be easier instead to just have annotations like `@keep` or `@skip-bake` or alike per class file?
Then the tool would be much simpler to check and skip, and would do so for all possible baked classes, not just a small subset.

A -f (--force) flag would still be possible to overwrite, maybe, in case you do want to re-bake all.

ADmad:
Yeah, I don't really like this either. Having multiple classes is very likely to cause confusion especially for new users as it wouldn't be easily apparent which classes are loaded, plus figuring out which ones they can edit and which ones they should not.

For e.g. the table auto loading feature caused lot of confusion for users.

GrandFelix:
> Let's wait with docs and more refactor until we have a clear consensus on this proposed changeset.
> 
> For me it is a bit weird that this only applies to the model layer and for example not other classes around it. Wouldnt it be easier instead to just have annotations like `@keep` or `@skip-bake` or alike per class file? Then the tool would be much simpler to check and skip, and would do so for all possible baked classes, not just a small subset.

Indeed. Was thinking the same about VC layer. For the controller I think there is no need. Views are hard to extend because there is so much cases when user already modifies view, specially forms... Only place where this was useful to me during development was model layer. And it was useful a lot of times. All business logic goes to models and its a pain if you need to handle copy/paste to preserve code every time you run bake (I know, this is happening more or less at the beginning of the project but its PITA). It was so much PITA  that I have done this :)

table auto loading stays the the same, nothing changes. For persistent class --force flag is not taken into account (if --force would overwrite it I did't do nothing then)

dereuromark:
But you do see my point, right? That a `@skip-bake` flag would even allow customized bake views to remain now, or controllers.
So in the end you get support for all classes that could be relevant to a user of bake (not just you individually).

What parts are getting realistically added by each run?
Only validation rules, right?
We could also think about other ways we could add those based on the current schema in a less intrusive way
See e.g. IdeHelper on how it can modify existing code base.

ADmad:
Is a file being overwritten even an issue for those using git / vcs? One can easily restore the required parts using the IDE / diff tools.

>  All business logic goes to models ...

Maybe that's the root cause of your trouble :slightly_smiling_face:. Don't bloat models with business logic. If your logic is complex enough than use service classes.

GrandFelix:
By each run there are validations, associations,  etc. And when you add new relations / fields is much quicker to run bake than doing all by hand. And if I need to make diff every time after that... No. Ok, maybe once :) About business logic. Custom finders, complex validations which are only usable for one model... al the things that are strongly tied to one model. No abusing of model classes. About `@skip-bake` thats the same if I choose `n` for specific file during bake run. Also model relations etc. with `@skip-bake` will not be updated, which is not what I want. I think Propel ORM is doing the same. It create s stub classes which are created only once and there goes custom code.

dereuromark:
This needs to be first done as custom community plugin addon or as outlined above.

PR #799, opened 2021-12-24
Fix PHP 8.1 shell helper fatal error

### LINKED ISSUE #798: Generation of shell helper: Fatal error
> bin/cake bake shell_helper BazBaz

in PHP 8.1:

```
PHP Fatal error:  Declaration of App\Shell\Helper\BazBazHelper::output($args) must be compatible with Cake\Console\Helper::output(array $args): void in /var/work/devilbox/data/www/sandbox/src/Shell/Helper/BazBazHelper.php on line 20

Fatal error: Declaration of App\Shell\Helper\BazBazHelper::output($args) must be compatible with Cake\Console\Helper::output(array $args): void in /var/work/devilbox/data/www/sandbox/src/Shell/Helper/BazBazHelper.php on line 20
```

LordSimal:
fixes #798

### DIFF
--- a/templates/bake/Shell/helper.twig
+++ b/templates/bake/Shell/helper.twig
@@ -32,7 +32,7 @@ class {{ name }}Helper extends Helper
      *
      * @return void
      */
-    public function output($args)
+    public function output(array $args)
     {
     }
 }
--- a/templates/bake/Shell/helper.twig
+++ b/templates/bake/Shell/helper.twig
@@ -32,7 +32,7 @@ class {{ name }}Helper extends Helper
      *
      * @return void
      */
-    public function output(array $args)
+    public function output(array $args): void
     {
     }
 }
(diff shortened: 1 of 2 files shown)


templates/bake/Shell/helper.twig
```diff
@@ -32,7 +32,7 @@ class {{ name }}Helper extends Helper
      *
      * @return void
      */
-    public function output($args)
+    public function output(array $args)
```
dereuromark: Should we add void return too?

LordSimal: ah right, would be nice as well 😄

dereuromark:
This should probably have been fixed in 2.x instead.

LordSimal:
oh, sorry didn't know that

dereuromark:
2.next is for New features or bc breaking issues

PR #800, opened 2021-12-24
fix --quiet option not being quiet

### LINKED ISSUE #796: Bake models / fixtures with --quiet output text anyway
```
$ bin/cake bake model -q Users
Baking table class for Users...

Baking entity class for Users...

Baking test fixture for Users...

Baking test case for App\Model\Table\UsersTable ...
```

https://github.com/cakephp/bake/blob/4746ffa5fbe0700a4334aebafb361350984e0fd0/src/Command/ModelCommand.php#L1072

https://github.com/cakephp/bake/blob/4746ffa5fbe0700a4334aebafb361350984e0fd0/src/Command/FixtureCommand.php#L268

Is there any reason to force output in quiet instead of normal ?

LordSimal:
fixes #796

### DIFF
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -116,7 +116,7 @@ class AllCommand extends BakeCommand
             }
         }
 
-        $io->out('<success>Bake All complete.</success>', 1, ConsoleIo::QUIET);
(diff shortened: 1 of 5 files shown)


dereuromark:
same , 2.x needs this fix already

PR #803, opened 2021-12-30
Add support for detecting/generating `hasOne` associations.

ndm2:
The code is pretty much identical for `findHasOne()` and `findHasMany()`, let me know if you'd rather like to see this split out further into some shared method(s).

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -226,6 +226,7 @@ class ModelCommand extends BakeCommand
 
         $associations = [
             'belongsTo' => [],
+            'hasOne' => [],
             'hasMany' => [],
             'belongsToMany' => [],
         ];
@@ -241,6 +242,7 @@ class ModelCommand extends BakeCommand
             return $associations;
         }
 
+        $associations = $this->findHasOne($table, $associations);
         $associations = $this->findHasMany($table, $associations);
         $associations = $this->findBelongsToMany($table, $associations);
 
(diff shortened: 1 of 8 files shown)


ndm2:
Sorry, wrong repo, I wanted to test in the fork. I hate that Github creates the PR in the upstream repo when clicking the link in the forked repo :(

Anyways, now it's here.

src/Command/ModelCommand.php
```diff
@@ -404,6 +406,112 @@ public function findTableReferencedBy(TableSchemaInterface $schema, string $keyF
...
+    public function isPossibleBelongsToManyRelation(string $sourceTable, string $targetTable): bool
+    {
+        $tables = $this->listAll();
+
+        $pregTableName = preg_quote($sourceTable, '/');
+        $pregPattern = "/^{$pregTableName}_|_{$pregTableName}$/";
+
+        if (preg_match($pregPattern, $targetTable) === 1) {
+            $possibleBTMTargetTable = preg_replace($pregPattern, '', $targetTable);
+            if (in_array($possibleBTMTargetTable, $tables)) {
```
dereuromark: can we use true as 3rd arg here always?

dereuromark (approved):
Pretty cool! :)

src/Command/ModelCommand.php
ndm2: Probably, it shouldn't make a difference with the return value of `preg_replace()` here, it will be either a string or `null`.

src/Command/ModelCommand.php
```diff
@@ -404,6 +406,112 @@ public function findTableReferencedBy(TableSchemaInterface $schema, string $keyF
...
+                        'alias' => $otherModel->getAlias(),
+                        'foreignKey' => $fieldName,
+                    ];
+                } elseif ($otherTableName === $tableName && $fieldName === 'parent_id') {
+                    $className = $this->plugin ? $this->plugin . '.' . $model->getAlias() : $model->getAlias();
+                    $assoc = [
+                        'alias' => 'Child' . $model->getAlias(),
+                        'className' => $className,
+                        'foreignKey' => $fieldName,
+                    ];
```
markstory: Would hasOne association be used for parent/child associations? I could see it being used for the parent side but not the child side.

ndm2: Admittedly, I'm having a hard time imagining a use case, but certainly associating a single self-referencing child could be a thing. A skewed binary tree has only one child per node, but that's probably a rather bad example 🙃

I didn't wanted to limit people, if someone wants a one-to-one parent-child association, who am I to judge. But of course starting limited, and expanding later if necessary is reasonable too, maybe even more reasonable.

ndm2: Or wait... maybe I misunderstood you. In this case the associated record would be a child, but the source side of the association would be the parent. Just like with `hasMany`, the target side would be the child with a `belongsTo` association.

othercorey:
Should probably target 2.next.

ndm2:
Most definitely, that's where it was supposed to go.

src/Command/ModelCommand.php
markstory: I'm having a hard time coming up with a scenario where I would want a child relationship to be a hasOne. Like you said that kind of relationship would be hasMany ChildXyz. I wouldn't include this code until we have a use-case for it.

ndm2: No problem, I'll cut it down for now, and make it pick it up for `hasMany` again.

markstory (approved):
Looks good to me.

othercorey:
Not sure why this wasn't merged.

PR #805, opened 2021-12-30
omit primary key validation and re-add FK validation rules while baking the model

### LINKED ISSUE #804: Omit validation rule generation for primary keys
DB Schema:

```
CREATE TABLE categories (
    id INT AUTO_INCREMENT PRIMARY KEY,
    name VARCHAR(255) NOT NULL,
    created DATETIME,
    modified DATETIME
);
```

creating all files via `bin/cake bake all categories` creates a validation rule like
```
    $validator
        ->integer('id')
        ->allowEmptyString('id', null, 'create');
```

even though those validation rules for primary keys never really do anything.

LordSimal:
Fixes #804

This PR adds checks so that the detected primary key field doesn't get validation rules added as well.

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -694,7 +694,7 @@ class ModelCommand extends BakeCommand
                 continue;
             }
             $field = $schema->getColumn($fieldName);
-            $validation = $this->fieldValidation($schema, $fieldName, $field);
+            $validation = $this->fieldValidation($schema, $fieldName, $field, $primaryKey);
             if ($validation) {
                 $validate[$fieldName] = $validation;
             }
@@ -709,12 +709,14 @@ class ModelCommand extends BakeCommand
      * @param \Cake\Database\Schema\TableSchemaInterface $schema The table schema for the current field.
      * @param string $fieldName Name of field to be validated.
      * @param array $metaData metadata for field
+     * @param array<string> $primaryKey The primary key field. Unused because PK validation is skipped
      * @return array Array of validation for the field.
      */
     public function fieldValidation(
         TableSchemaInterface $schema,
         string $fieldName,
-        array $metaData
+        array $metaData,
+        array $primaryKey
     ): array {
         $ignoreFields = ['lft', 'rght', 'created', 'modified', 'updated'];
         if (in_array($fieldName, $ignoreFields, true)) {
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -682,15 +682,9 @@ class ModelCommand extends BakeCommand
 
         $validate = [];
         $primaryKey = $schema->getPrimaryKey();
-        $foreignKeys = [];
-        if (isset($associations['belongsTo'])) {
-            foreach ($associations['belongsTo'] as $assoc) {
-                $foreignKeys[] = $assoc['foreignKey'];
-            }
-        }
         foreach ($fields as $fieldName) {
-            // Skip primary key and foreign key validations
-            if (in_array($fieldName, $primaryKey, true) || in_array($fieldName, $foreignKeys, true)) {
+            // Skip primary key
+            if (in_array($fieldName, $primaryKey, true)) {
                 continue;
             }
             $field = $schema->getColumn($fieldName);
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -689,11 +689,12 @@ class ModelCommand extends BakeCommand
             }
         }
         foreach ($fields as $fieldName) {
-            if (in_array($fieldName, $foreignKeys, true)) {
+            // Skip primary key and foreign key validations
+            if (in_array($fieldName, $primaryKey, true) || in_array($fieldName, $foreignKeys, true)) {
                 continue;
             }
             $field = $schema->getColumn($fieldName);
-            $validation = $this->fieldValidation($schema, $fieldName, $field, $primaryKey);
+            $validation = $this->fieldValidation($schema, $fieldName, $field);
             if ($validation) {
                 $validate[$fieldName] = $validation;
             }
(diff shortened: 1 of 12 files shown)


src/Command/ModelCommand.php
```diff
@@ -708,14 +709,12 @@ public function getValidation(Table $model, array $associations, Arguments $args
...
      * @param array $metaData metadata for field
-     * @param array<string> $primaryKey The primary key field
      * @return array Array of validation for the field.
      */
     public function fieldValidation(
         TableSchemaInterface $schema,
         string $fieldName,
-        array $metaData,
-        array $primaryKey
+        array $metaData
```
ADmad: We can't change public API without a new major. Just update the description that the param is unused and continue passing the value.

src/Command/ModelCommand.php
```diff
@@ -689,11 +689,12 @@ public function getValidation(Table $model, array $associations, Arguments $args
             }
         }
         foreach ($fields as $fieldName) {
-            if (in_array($fieldName, $foreignKeys, true)) {
+            // Skip primary key and foreign key validations
+            if (in_array($fieldName, $primaryKey, true) || in_array($fieldName, $foreignKeys, true)) {
```
ADmad: Why are you skipping foreign key validation?

LordSimal: This was present before as well, don't know why 🤷🏻

src/Command/ModelCommand.php
LordSimal: Oh right 😅  Will add that back

src/Command/ModelCommand.php
dereuromark: It is there from the start
https://github.com/cakephp/bake/commit/12dea6d0e168031c6f939ce6adffa29943f5783e#diff-b2a7f0a392cef7628e2952ea340e286e957a809321cbe9cb98047925b2e83e8bR667
Maybe the primary key part was meant all along?

ADmad: @LordSimal Sorry, missed the fact that it already existed.

Though I wonder why, since fKs can be part of mass assignment and should be validated.

ADmad: @dereuromark It's the same in the old task too, so if it's a mistake it's existed for a very long time.
https://github.com/cakephp/bake/blob/33e8ee8419ba36c13fa4074c208c93352b5530cf/src/Shell/Task/ModelTask.php#L679

LordSimal: If you desire I can change that to only skip on PK but not on FK

LordSimal: Its funny because there is an extra Test just to check that FK are not included in the generated validation 😂 
https://github.com/cakephp/bake/blob/2.x/tests/TestCase/Command/ModelCommandTest.php#L1088

dereuromark:
Looks good to me. @markstory might want to way in regarding the foreign key topic.

LordSimal:
A little update what happened in the last commits:
We talked about the fact, that FK Validation generation should have never even been skipped in the first place but there even was a Test present for that.
https://github.com/cakephp/bake/blob/2.x/tests/TestCase/Command/ModelCommandTest.php#L1088

Had to remove that of course.

The only other custom change I had to do was the fact, that SQLite now seems to generate a `-> nonNegativeInteger()` as a FK validator. See https://github.com/cakephp/bake/pull/805/commits/d18fea1e16915f9fae129aa814e059268e434348

LordSimal:
The only other thing I could probably adjust is the place where the FK validation rules are being added to the `validationDefault()` method.
Currently they are added at the start but I personally think they should be at the end since I also like to put all my FK columns at the end of the table.
What are you opinions?

LordSimal:
I guess it doesn't matter because the validations are added in the order the columns are present in the table. So we shouldn't sort it ourselves.

markstory:
Foreign key validation rules aren't generated because when the ORM is creating or updating records foreign keys are either not present in the request data and are generated by the ORM or are already set on existing entities and not part of the mass property assignment.

LordSimal:
So should the "old" logic still keep as it is so that FK validation rules are not being added?

dereuromark:
> and not part of the mass property assignment.

Yeah, but down the road they are or at least can be, especially since more and more apps are not form based, but with some JS framework and therefore more API driven.
So it seems we might be better adding them by default, even if they are not yet used that much in default app usage.

tests/TestCase/Command/ModelCommandAssociationDetectionTest.php
```diff
@@ -160,6 +160,7 @@ public function testBakeAssociationDetectionProductVersionsTable()
     {
         $driver = ConnectionManager::get('test')->getDriver();
         $this->skipIf($driver instanceof Mysql, 'Incompatible with mysql');
+        $this->skipIf($driver instanceof Sqlite, 'Incompatible with sqlite');
```
othercorey, line 163: Why is this now incompatible with sqlite?

tests/TestCase/Command/ModelCommandTest.php
```diff
@@ -1146,9 +1107,10 @@ public function testGetValidationExcludeForeignKeysSigned()
                 'integer' => ['rule' => 'integer', 'args' => []],
                 'notEmpty' => ['rule' => 'notEmptyString', 'args' => []],
             ],
-            'id' => [
+            'user_id' => [
                 'integer' => ['rule' => 'integer', 'args' => []],
-                'allowEmpty' => ['rule' => 'allowEmptyString', 'args' => [null, 'create']],
+                'requirePresence' => ['rule' => 'requirePresence', 'args' => ['create']],
```
ADmad, line 1112: We should have a test for nullable foreignKey field too and ensure `requirePresence()` is not used for it.

tests/TestCase/Command/ModelCommandAssociationDetectionTest.php
LordSimal: As my comment states above: 
The only other custom change I had to do was the fact, that SQLite now seems to generate a `->nonNegativeInteger()` as a FK validator. See d18fea1
Thats why that tests needs to be skipped for the unsigned version of the test but done for the signed one

ADmad:
> when the ORM is creating or updating records foreign keys are either not present in the request data and are generated by the ORM

Good point, I missed that. The fKs are populated later by the ORM when doing saves with associations. So "requirePreence()" at least should always be skipped for fKs.

LordSimal:
So we should add a test containing nullable foreign keys as well and check if the generated model files are correct, right? 😅

markstory:
> So we should add a test containing nullable foregin keys as well and check if the generated model files are correct, right? 

It would be good to have more tests :smile:

LordSimal:
Articles already has a nullable `author_id` here https://github.com/cakephp/bake/blob/2.x/tests/schema.php#L70

So therefore the nullable foreign key check has already been done in this test https://github.com/cakephp/bake/blob/71853c607feb8e788e9a4af3f3c93408467c12b6/tests/TestCase/Command/ModelCommandTest.php#L1352

But now I added an extra test anyway.

PR #806, opened 2022-01-05
change validateNames logic to prevent special chars in column names

LordSimal:
This story started with the following discourse entry: https://discourse.cakephp.org/t/bake-cant-handle-column-name-containing/10046

The changes made here further prevent users to bake code which doesn't work like having a column name `data:3` and therefore we generate code like `$individual->data:3` which obviously doesn't work.

The Regex I used here basically states that:

- the column name has to start with a character (lower or upper case) or an underscore and
- further characters are allowed to be either lower or upper case characters, numbers or underscores

Anything else like
- any other special character
- umlauts
- whatever the user thinks is a good idea to use in a table name (emojis? 😂 )

is prevented with this check.

### DIFF
--- a/src/Command/BakeCommand.php
+++ b/src/Command/BakeCommand.php
@@ -166,13 +166,13 @@ abstract class BakeCommand extends Command
     }
 
     /**
-     * Check if a column name is valid
+     * Check if a column name is valid.
      *
      * The Regex used here basically states that:
      * - the column name has to start with an ASCII character (lower or upper case) or an underscore and
-     * - further characters are allowed to be either lower or upper case ASCII characters, numbers or underscores
+     * - further characters are allowed to be either lower or upper case ASCII characters, numbers or underscores.
      *
-     * @param string $name The name of the column
+     * @param string $name The name of the column.
      * @return bool
      */
     protected function isValidColumnName(string $name): bool
--- a/src/Command/BakeCommand.php
+++ b/src/Command/BakeCommand.php
@@ -164,4 +164,27 @@ abstract class BakeCommand extends Command
             $io->out(sprintf('<success>Deleted</success> `%s`', $path), 1, ConsoleIo::NORMAL);
         }
     }
+
+    /**
+     * The Regex used here basically states that:
+     *
+     * the column name has to start with a character (lower or upper case) or an underscore and
+     * further characters are allowed to be either lower or upper case characters, numbers or underscores
+     *
+     * Anything else like
+     * - any other special character
+     * - umlauts
+     * - whatever the user thinks is a good idea to use in a table name (emojis?)
+     *
+     * is prevented with this check.
+     *
+     * @param string $name The name of the column
+     * @return bool true, if it is valid, false if not
+     */
+    protected function isValidColumnName(string $name): bool
+    {
+        $re = '/^[a-zA-Z_][a-zA-Z0-9_]*$/';
+
+        return (bool)preg_match($re, $name);
+    }
 }
--- a/src/Command/BakeCommand.php
+++ b/src/Command/BakeCommand.php
@@ -166,8 +166,9 @@ abstract class BakeCommand extends Command
     }
 
     /**
-     * The Regex used here basically states that:
+     * Check if a column name is valid
      *
+     * The Regex used here basically states that:
      * the column name has to start with a character (lower or upper case) or an underscore and
      * further characters are allowed to be either lower or upper case characters, numbers or underscores
      *
--- a/src/Command/BakeCommand.php
+++ b/src/Command/BakeCommand.php
@@ -169,23 +169,14 @@ abstract class BakeCommand extends Command
      * Check if a column name is valid
      *
      * The Regex used here basically states that:
-     * the column name has to start with a character (lower or upper case) or an underscore and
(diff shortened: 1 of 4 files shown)


src/Command/ModelCommand.php
```diff
@@ -127,9 +127,12 @@ public function bake(string $name, Arguments $args, ConsoleIo $io): void
     public function validateNames(TableSchemaInterface $schema, ConsoleIo $io): void
     {
         foreach ($schema->columns() as $column) {
-            if (!is_string($column) || (!ctype_alpha($column[0]) && $column[0] !== '_')) {
+            $re = '/^[a-zA-Z_][a-zA-Z0-9_]*$/m';
```
othercorey: Why do we need the `m` modifier?

src/Command/ModelCommand.php
```diff
@@ -127,9 +127,12 @@ public function bake(string $name, Arguments $args, ConsoleIo $io): void
...
+            preg_match_all($re, $column, $matches, PREG_SET_ORDER);
```
othercorey: Why can't we just use preg_match?

dereuromark:
```
The Regex I used here basically states that:

the column name has to start with a character (lower or upper case) or an underscore and
further characters are allowed to be either lower or upper case characters, numbers or underscores
Anything else like

any other special character
umlauts
whatever the user thinks is a good idea to use in a table name (emojis? joy )
is prevented with this check.
```
If you make it a protected method isValidColumnName() or alike you can state this whole part in the docblock for transparency in the future.

src/Command/ModelCommand.php
LordSimal: right, we don't need multiline here so I will remove it

src/Command/ModelCommand.php
LordSimal: will do that

LordSimal:
I just realised that the same logic was also present in the `src/Command/FixtureCommand.php`
Therefore I put the `isValidColumnName()` into the base `BakeCommand.php` class.

src/Command/BakeCommand.php
```diff
@@ -164,4 +164,27 @@ protected function deleteEmptyFile(string $path, ConsoleIo $io): void
             $io->out(sprintf('<success>Deleted</success> `%s`', $path), 1, ConsoleIo::NORMAL);
         }
     }
+
+    /**
+     * The Regex used here basically states that:
```
othercorey: Please add a summary line at the top.

dereuromark: I didnt think u would copy paste 1:1 :)
This could need a bit fine tuning

LordSimal: done 👍🏻

LordSimal: @dereuromark I like to be straight forward 😛

src/Command/BakeCommand.php
```diff
@@ -164,4 +164,28 @@ protected function deleteEmptyFile(string $path, ConsoleIo $io): void
...
+     *
+     * Anything else like
+     * - any other special character
+     * - umlauts
+     * - whatever the user thinks is a good idea to use in a table name (emojis?)
+     *
+     * is prevented with this check.
+     *
+     * @param string $name The name of the column
+     * @return bool true, if it is valid, false if not
```
dereuromark: I am not a big Fan of stating obvious things in description 
But no need for nitpick here :)

LordSimal: Well I tried to follow basically all other return type docs present in at least that file 😅

src/Command/BakeCommand.php
```diff
@@ -164,4 +164,28 @@ protected function deleteEmptyFile(string $path, ConsoleIo $io): void
...
     }
+
+    /**
+     * Check if a column name is valid
+     *
+     * The Regex used here basically states that:
+     * the column name has to start with a character (lower or upper case) or an underscore and
+     * further characters are allowed to be either lower or upper case characters, numbers or underscores
+     *
+     * Anything else like
```
ADmad: The valid charset is stated above so listing out various invalid ones again is redundant. It will unnecessarily make the API docs longer.

src/Command/BakeCommand.php
```diff
@@ -164,4 +164,28 @@ protected function deleteEmptyFile(string $path, ConsoleIo $io): void
...
+     * - whatever the user thinks is a good idea to use in a table name (emojis?)
+     *
+     * is prevented with this check.
+     *
+     * @param string $name The name of the column
+     * @return bool true, if it is valid, false if not
+     */
+    protected function isValidColumnName(string $name): bool
+    {
+        $re = '/^[a-zA-Z_][a-zA-Z0-9_]*$/';
```
ADmad: The regex is short enough, so can directly use as argument instead of assigning to a variable first.

ADmad (commented):
Since this adds a new protected method it needs to target `2.next`.

src/Command/BakeCommand.php
markstory: :+1: We gotta save those op-codes 💸

markstory (commented):
Looks good once the doc block is cleaned up.

PR #809, opened 2022-02-27
added services method to baked plugin template

### LINKED ISSUE #808: Creation of services method in plugin bake process
### Description

It would be great to add the services method hook to the plugin bake command, as well as the bootstrap, middleware and routes ones.
It would also be great to add the services plugin hook documentation to the cookbook plugin hook  section, which is actually missing.

### CakePHP Version

_No response_

LordSimal:
Fixes #808


othercorey:
Looks ok o me

PR #810, opened 2022-02-27
changed method visibility for baked setUp and tearDown tests to protected

### LINKED ISSUE #801: Baked setUp/tearDown visibility
Currently baked tests contain
```php
    /**
     * setUp method
     *
     * @return void
     */
    public function setUp(): void
    {
```

Based on parent method in Cake as well as PHPUnit those should be protected though.

LordSimal:
Fixes #801

### DIFF
--- a/templates/bake/tests/test_case.twig
+++ b/templates/bake/tests/test_case.twig
@@ -88,7 +88,7 @@ class {{ className }}Test extends TestCase
      *
      * @return void
      */
-    public function setUp(): void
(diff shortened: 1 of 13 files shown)


othercorey:
Looks ok to me. Doesn't make a big difference though.

LordSimal:
I guess its more being consistent with the parent method also being `protected`

https://github.com/cakephp/cakephp/blob/4.x/src/TestSuite/TestCase.php#L246

PR #811, opened 2022-02-27
don't use Number helper if baked field is nullable

### LINKED ISSUE #807: Number format helper displays NULL as 0
When baking an app, NULL-values are shown as a zero because of the use of the Number-helper:

$this->Number->format(null) gives the zero ([issue](https://github.com/cakephp/cakephp/issues/8961)).

When there's a NULL in the database, expect it to be displayed as a blank.

Added a test if field is NULL, and works like expected ([topic](https://discourse.cakephp.org/t/database-nullable-integer-field-has-value-null-cakephp-displaying-0/10082)).

LordSimal:
Fixes #807 

This will prevent the bake command to use the Number helper `$this->Number->format()` from being used if the given field is nullable.

### DIFF
--- a/templates/bake/Template/index.twig
+++ b/templates/bake/Template/index.twig
@@ -49,8 +49,10 @@
 {% endif %}
 {% if isKey is not same as(true) %}
 {% set columnData = Bake.columnData(field, schema) %}
-{% if columnData.type not in ['integer', 'float', 'decimal', 'biginteger', 'smallinteger', 'tinyinteger'] or columnData.null is same as(true)  %}
+{% if columnData.type not in ['integer', 'float', 'decimal', 'biginteger', 'smallinteger', 'tinyinteger'] %}
                     <td><?= h(${{ singularVar }}->{{ field }}) ?></td>
+{% elseif columnData.null %}
+                    <td><?= ${{ singularVar }}->{{ field }} === null ? '' : $this->Number->format(${{ singularVar }}->{{ field }}) ?></td>
 {% else %}
                     <td><?= $this->Number->format(${{ singularVar }}->{{ field }}) ?></td>
 {% endif %}
--- a/templates/bake/Template/index.twig
+++ b/templates/bake/Template/index.twig
@@ -49,7 +49,7 @@
 {% endif %}
 {% if isKey is not same as(true) %}
 {% set columnData = Bake.columnData(field, schema) %}
-{% if columnData.type not in ['integer', 'float', 'decimal', 'biginteger', 'smallinteger', 'tinyinteger'] %}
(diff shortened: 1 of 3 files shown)


templates/bake/Template/index.twig
```diff
@@ -49,7 +49,7 @@
 {% endif %}
 {% if isKey is not same as(true) %}
 {% set columnData = Bake.columnData(field, schema) %}
-{% if columnData.type not in ['integer', 'float', 'decimal', 'biginteger', 'smallinteger', 'tinyinteger'] %}
+{% if columnData.type not in ['integer', 'float', 'decimal', 'biginteger', 'smallinteger', 'tinyinteger'] or columnData.null is same as(true)  %}
```
othercorey: Isn't the value the issue not the nullability?

templates/bake/Template/view.twig
```diff
@@ -67,7 +67,12 @@
 {% for field in groupedFields.number %}
                 <tr>
                     <th><?= __('{{ field|humanize }}') ?></th>
+{% set columnData = Bake.columnData(field, schema) %}
+{% if columnData.null is same as(true) %}
```
ADmad: Isn't `{% if columnData.null %}` enough?

templates/bake/Template/index.twig
ADmad: Yup, this probably isn't the correct fix. The value itself needs to be checked at "runtime".

So for nullable numeric columns the generated PHP code should be 
```php
<td><?= $entity->field === null ? '' : $this->Number->format($entity->field) ?></td>
```

LordSimal: alright, then I misunderstood the problem. will adjust soon

templates/bake/Template/view.twig
LordSimal: indeed 😅

templates/bake/Template/index.twig
```diff
@@ -51,6 +51,8 @@
 {% set columnData = Bake.columnData(field, schema) %}
 {% if columnData.type not in ['integer', 'float', 'decimal', 'biginteger', 'smallinteger', 'tinyinteger'] %}
                     <td><?= h(${{ singularVar }}->{{ field }}) ?></td>
+{% elseif columnData.null %}
```
othercorey, line 54: We could probably skip this check to make the code simpler.

LordSimal: I thought its more desirable to generate cleaner code so not all fields will get the long version
```
<td><?= ${{ singularVar }}->{{ field }} === null ? '' : $this->Number->format(${{ singularVar }}->{{ field }}) ?></td>
```

othercorey: Sure, I guess its fine if our code has the check.

PR #812, opened 2022-02-27
fix tests not using reference to command code constant via the interface

LordSimal:
fixes the following "warning"

image


ADmad:
That's one whiny IDE :)

LordSimal:
there is more to come soon 😉

PR #813, opened 2022-02-27
minor cleanup

LordSimal:
Here are some minor changes my IDE recommended to do.

### DIFF
--- a/src/Command/TestCommand.php
+++ b/src/Command/TestCommand.php
@@ -451,6 +451,7 @@ class TestCommand extends BakeCommand
             $this->_processController($subject);
         }
 
+        /** @psalm-suppress RedundantFunctionCall */
         return array_values($this->_fixtures);
     }
 
--- a/src/Command/TestCommand.php
+++ b/src/Command/TestCommand.php
@@ -416,6 +416,7 @@ class TestCommand extends BakeCommand
      *
      * @param string $className Name of class to look at.
      * @return string[] Array of method names.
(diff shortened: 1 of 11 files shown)


ADmad:
Please update psalm & phpstan versions in composer's `stan-setup` script too.

LordSimal:
As you see in the phpstan output I now get 
```
------ -------------------------------------------------------------------- 
  Line   Plugin.php                                                          
 ------ -------------------------------------------------------------------- 
  145    Dead catch - ReflectionException is never thrown in the try block.  
 ------ --------------------------------------------------------------------
```

even though (in my opinion) it is needed 
https://github.com/cakephp/bake/blob/2.x/src/Plugin.php#L144

ADmad:
That looks like a bug in phpstan, need to ignore that error.

src/Command/TestCommand.php
```diff
@@ -450,6 +451,7 @@ public function generateFixtureList($subject): array
             $this->_processController($subject);
         }
 
+        /** @psalm-suppress RedundantFunctionCall */
```
LordSimal, line 454: Had to add that here, otherwise I got
```
ERROR: RedundantFunctionCall - src/Command/TestCommand.php:454:16 - The call to array_values is unnecessary, array<empty, empty> is already a list (see https://psalm.dev/280)
        return array_values($this->_fixtures);
```

ADmad (approved):

PR #821, opened 2022-04-14
ubuntu 20.04 for github actions

saeideng:


ADmad:
Given the failure we can hold off on this until Cake 4.4 is release which is PHP 7.4+.

PR #830, opened 2022-06-10
3.x - Remove deprecated shell tasks

othercorey:
Removing deprecated classes to avoid dependencies before feature work begins.


src/Command/EntryCommand.php
```diff
@@ -1,231 +0,0 @@
...
-use Cake\Console\Exception\ConsoleException;
-use Cake\Console\Shell;
-use Cake\Core\Configure;
-use Cake\Core\Plugin as CorePlugin;
-use Cake\Utility\Inflector;
-
-/**
- * Command that provides help and an entry point to bake tools.
- */
-class EntryCommand extends Command implements CommandCollectionAwareInterface
```
ADmad, line 36: What will be the output for `cake bake` if this command is removed?

othercorey: We probably need a help command. I didn't notice it performed both functions.

ADmad: Well the command's name doesn't really matter. We can just remove the unneeded parts of this command and avoid unnecessary code churn.

othercorey: Agreed.

othercorey:
Realized that unless Bake 3 is CakePHP 5+, we can't prevent the shell tasks from showing up in the command list.

PR #831, opened 2022-06-10
Remove access to impossible array offset

othercorey:
This fixes a psalm error since `$assoc['className']` can never exist.

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -494,7 +494,7 @@ class ModelCommand extends BakeCommand
                         'foreignKey' => $fieldName,
                     ];
                 }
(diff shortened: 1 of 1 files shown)


othercorey:
@ndm2 Was this supposed to do something else with `className`?

ndm2:
It's a leftover from dropping self-reference support https://github.com/cakephp/bake/compare/c29d4d7ba97b6d25ea518e5b9dcf5bc2e36cb03c..8d3e08f11ac1389efb025e04ed4796e012876df9#diff-b2a7f0a392cef7628e2952ea340e286e957a809321cbe9cb98047925b2e83e8bL495-L502

ndm2 (approved):

PR #836, opened 2022-06-25
Update for Cake 5.

ADmad:


ADmad:
@othercorey Slevomat CS is barfing on `tests/TestCase/Command/TemplateCommandTest.php`, any idea how to fix/figure out the cause?

EDIT: Figured it out. It blew up because a docblock had `@param $name` instead of `@param string $name`.

othercorey:
> @othercorey Slevomat CS is barfing on `tests/TestCase/Command/TemplateCommandTest.php`, any idea how to fix/figure out the cause?
> 
> EDIT: Figured it out. It blew up because a docblock had `@param $name` instead of `@param string $name`.

phpstan decided to change the return type which broke slevomat.

src/BakePlugin.php
```diff
@@ -21,34 +21,30 @@
...
 /**
  * Plugin class for bake
  */
-class Plugin extends BasePlugin
+class BakePlugin extends BasePlugin
 {
     /**
      * Plugin name.
      *
      * @var string
```
othercorey:
```suggestion
     * @var string|null
```
if you meant for this to be nullable

othercorey:
looks good.

src/BakePlugin.php
ADmad: I wish we didn't have to add these redundant docblock types anymore :(.

othercorey: We need to do a test with the slevomat setting that detects redundant blocks. I don't know if it forces all @param annotations to be removed even if there's a description you want to keep. Should we just not have comments on parameters annotations anymore?

ADmad: I don't think we should drop parameter comments but I doubt it would allow param description without the type.

othercorey: Need to test if a comment prevents it from being marked useless.

dereuromark: I would rather check why the sniffs are missing for it:
https://github.com/spryker/code-sniffer/blob/master/Spryker/Sniffs/Commenting/DocBlockReturnNullSniff.php
or
https://github.com/spryker/code-sniffer/blob/master/Spryker/Sniffs/Commenting/DocBlockReturnNullableTypeSniff.php
could be the right direction here.

As long as they are properly synced, which is easier, all is well.

PR #838, opened 2022-07-06
Update command's template.

ADmad:


ADmad:
This can even be backported as narrowing the return type is covariant.

templates/bake/Command/command.twig
```diff
@@ -46,10 +46,11 @@ class {{ name }}Command extends Command
...
      * @param \Cake\Console\Arguments $args The command arguments.
-     * @param \Cake\Console\ConsoleIo $io The console io
-     * @return null|void|int The exit code or null for success
+     * @param \Cake\Console\ConsoleIo $io The console io.
+     * @return int The exit code.
      */
-    public function execute(Arguments $args, ConsoleIo $io)
+    public function execute(Arguments $args, ConsoleIo $io): int
     {
+        return static::CODE_SUCCESS;
```
markstory, line 54: We didn't require this in the past as it adds boilerplate to all commands. When we can easily infer that null = success.

ADmad: Yes but that prevents us from specifying return type for the method. We also have to suppress errors thrown by tools for the wacky `null|void|int` in the docblock which isn't exactly as valid union of types. I don't think 1 line of pre-generated boilerplate code should be an issue.

This is in reference to https://github.com/cakephp/cakephp/pull/16609 for which other teams members have reacted positively to.

markstory: Ok, I'll make a note of this as something we should try to fix with rector if possible. I personally will want this to be an automated change when we change the core typehint.

tests/comparisons/Command/testBakePlugin.php
```diff
@@ -31,10 +31,11 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
      * Implement this method with your command's logic.
      *
      * @param \Cake\Console\Arguments $args The command arguments.
-     * @param \Cake\Console\ConsoleIo $io The console io
-     * @return null|void|int The exit code or null for success
+     * @param \Cake\Console\ConsoleIo $io The console io.
+     * @return int The exit code.
```
othercorey, line 35: I think we avoid periods in these comments since it's so inconsistent

othercorey:
Do we need to continue with this change?

PR #841, opened 2022-07-14
Don't enforce FKs presence on create to allow e.g. hasMany association save calls to run through

LordSimal:
Caused by #805 

As just discussed in discord my implementation of requiring not nullable FKs in the validator causes issues when you want to create new entities via an e.g. hasMany association save call.

See https://youtu.be/d499_oCjq1s?t=340

The main cause behind all of this is the `->requirePresence()` which can be removed to allow that functionality again.

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -794,9 +794,9 @@ class ModelCommand extends BakeCommand
             if (in_array($fieldName, $primaryKey, true)) {
                 continue;
             }
-            $isForeignKey = in_array($fieldName, $foreignKeys, true);
             $field = $schema->getColumn($fieldName);
-            $validation = $this->fieldValidation($schema, $fieldName, $field, $primaryKey, $isForeignKey);
+            $field['isForeignKey'] = in_array($fieldName, $foreignKeys, true);
+            $validation = $this->fieldValidation($schema, $fieldName, $field, $primaryKey);
             if ($validation) {
                 $validate[$fieldName] = $validation;
             }
@@ -811,16 +811,14 @@ class ModelCommand extends BakeCommand
      * @param \Cake\Database\Schema\TableSchemaInterface $schema The table schema for the current field.
      * @param string $fieldName Name of field to be validated.
      * @param array $metaData metadata for field
-     * @param array<string> $primaryKey The primary key field. Unused because PK validation is skipped
-     * @param bool $isForeignKey Indicates if the currently precessed field is a foreign key
+     * @param array<string> $primaryKey The primary key field. Unused because PK validation is skippe
      * @return array Array of validation for the field.
      */
     public function fieldValidation(
         TableSchemaInterface $schema,
         string $fieldName,
         array $metaData,
-        array $primaryKey,
-        bool $isForeignKey = false
+        array $primaryKey
     ): array {
         $ignoreFields = ['lft', 'rght', 'created', 'modified', 'updated'];
         if (in_array($fieldName, $ignoreFields, true)) {
@@ -886,7 +884,7 @@ class ModelCommand extends BakeCommand
             ];
         } else {
             // FKs shouldn't be required on create to allow e.g. save calls with hasMany associations to create entities
-            if (($metaData['default'] === null || $metaData['default'] === false) && !$isForeignKey) {
+            if (($metaData['default'] === null || $metaData['default'] === false) && !$metaData['isForeignKey']) {
                 $validation['requirePresence'] = [
                     'rule' => 'requirePresence',
                     'args' => ['create'],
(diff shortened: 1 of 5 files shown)


src/Command/ModelCommand.php
```diff
@@ -804,14 +811,14 @@ public function getValidation(Table $model, array $associations, Arguments $args
...
-     * @param array<string> $primaryKey The primary key field. Unused because PK validation is skipped
+     * @param bool $isForeignKey True if the current is a foreign key
      * @return array Array of validation for the field.
      */
     public function fieldValidation(
         TableSchemaInterface $schema,
         string $fieldName,
         array $metaData,
-        array $primaryKey
+        bool $isForeignKey
```
ADmad: We don't change public API in minor releases.

LordSimal: I have now changed the implementation to a property

src/Command/ModelCommand.php
```diff
@@ -60,6 +60,13 @@ class ModelCommand extends BakeCommand
      */
     protected $_tables = [];
 
+    /**
+     * True if the currently processed field is a foreign key
+     *
+     * @var bool
+     */
+    protected $isForeignKey = false;
```
othercorey: This needs a more specific name. It sounds like the model is a foreign key.

Seems odd to save this off for a single field as a global state.

LordSimal: The associations are only available in the `getValidation()` method but I need the info if the current field is a FK or not in the `fieldValidation()` method.
Since I am not allowed to change the arguments passed to the method to not break BC how else can I pass data from one method to another?

othercorey: I understand. Can we just name it something more specific?

LordSimal: I was just asking because maybe I wasn't aware of some simple way to achieve that.

othercorey: Maybe when we re-write it. No worried about it for now.

src/Command/ModelCommand.php
```diff
@@ -783,11 +790,18 @@ public function getValidation(Table $model, array $associations, Arguments $args
...
+            foreach ($associations['belongsTo'] as $assoc) {
+                $foreignKeys[] = $assoc['foreignKey'];
+            }
+        }
         foreach ($fields as $fieldName) {
             // Skip primary key
             if (in_array($fieldName, $primaryKey, true)) {
                 continue;
             }
+            $this->isCurrentFieldForeignKey = in_array($fieldName, $foreignKeys, true);
```
markstory: I don't love this as a property. Couldn't this be a parameter to `fieldValidation()`?  As a property it will sometimes be stale. Is there a reason you went with a property?

LordSimal:
I now changed the implementation to an optional, additional argument in `fieldValidation()`
I guess this is non-BC-breaking?
Otherwise I am fine with just doing this in the upcoming 3.x release for Cake5

src/Command/ModelCommand.php
```diff
@@ -819,13 +812,15 @@ public function getValidation(Table $model, array $associations, Arguments $args
...
+     * @param bool $isForeignKey Indicates if the currently precessed field is a foreign key
      * @return array Array of validation for the field.
      */
     public function fieldValidation(
         TableSchemaInterface $schema,
         string $fieldName,
         array $metaData,
-        array $primaryKey
+        array $primaryKey,
+        bool $isForeignKey = false
```
ADmad: This is again change in method signature and will break people's overridden method, me being one of them.

Personally I can simply update my code for the new minor but we shouldn't break our backwards compatibility promise for minor versions.

LordSimal:
I will re-do this PR on 3.x because neither the property way seems to be good nor can I adjust the method arguments for `fieldValidation()` which I need (at least without breaking BC)

src/Command/ModelCommand.php
```diff
@@ -783,13 +783,20 @@ public function getValidation(Table $model, array $associations, Arguments $args
...
+        }
         foreach ($fields as $fieldName) {
             // Skip primary key
             if (in_array($fieldName, $primaryKey, true)) {
                 continue;
             }
+            $isForeignKey = in_array($fieldName, $foreignKeys, true);
             $field = $schema->getColumn($fieldName);
-            $validation = $this->fieldValidation($schema, $fieldName, $field, $primaryKey);
+            $validation = $this->fieldValidation($schema, $fieldName, $field, $primaryKey, $isForeignKey);
```
ndm2: Maybe passing the flag in the field's metadata could've been an option?

```php
$field = $schema->getColumn($fieldName);
$field['isForeignKey'] = in_array($fieldName, $foreignKeys, true);

$validation = $this->fieldValidation($schema, $fieldName, $field, $primaryKey);
```

ADmad (approved):

Issue #834, opened 2022-06-13
Table with *_id column that isn't a foreign key doesn't generate validation


othercorey:
When baking a model for a table with a column that "looks" like a foreign key such as `external_id` but has no target table, no validation is generated for the column. Instead an association to a non-existent model is generated.

After removing the invalid association, the user might not notice the validation is missing.

ajibarra:
@othercorey Hi, I was trying to reproduce this bug using a table with a field `theme_id` and no themes table but the validation was generated correctly:

```
 $validator
            ->nonNegativeInteger('theme_id')
            ->requirePresence('theme_id', 'create')
            ->notEmptyString('theme_id');
```

Could you please provide more details about how to reproduce this bug? Please include Bake version and a DB schema if possible.

dereuromark:
Is this still valid and reproducable?

dereuromark:
Seems solved with https://github.com/cakephp/bake/pull/842 for now?

PR #842, opened 2022-07-15
Avoid generating association if table object or database table exist

### LINKED ISSUE #839: RFC: Do not generate associations for `_id` fields if no Table object or database table present
### Description

When baking a model, if you database table has a field with `_id` suffix it generates a belongsTo association. It is usually useful but sometimes it is not an actual association.

i.e A `users` table with an `external_id` to store the user id in another app/service.

Currently Bake would generate a belongsTo association with `Externals` as alias, even if ExternalsTable does not exist and `externals` table does not exist onto database.

I would propose to only generate the association if any of the above conditions is met.

I have a partial solution already but I would like to know your thoughts.

ajibarra:
This PR fixes #839 and #834 . Please check and let me know if any change is needed.

Thank you

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -30,6 +30,7 @@ use Cake\Database\Schema\TableSchema;
 use Cake\Database\Schema\TableSchemaInterface;
 use Cake\Datasource\ConnectionManager;
 use Cake\ORM\Table;
+use Cake\ORM\TableRegistry;
 use Cake\Utility\Inflector;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Command/ModelCommand.php
```diff
@@ -346,11 +346,19 @@ public function findBelongsTo(Table $model, array $associations): array
                 ];
             } else {
                 $tmpModelName = $this->_modelNameFromKey($fieldName);
-                if (!in_array(Inflector::tableize($tmpModelName), $this->_tables, true)) {
+                $this->getTableLocator()->get($tmpModelName);
+                $genericInstances = $this->getTableLocator()->genericInstances();
+                $tables = $this->listAll();
+                /** Check if association model could not be instantiated as a subclass but a generic Table instance instead. */
```
othercorey:
```suggestion
                // Check if association model could not be instantiated as a subclass but a generic Table instance instead
```

src/Command/ModelCommand.php
```diff
@@ -346,11 +346,19 @@ public function findBelongsTo(Table $model, array $associations): array
                 ];
             } else {
                 $tmpModelName = $this->_modelNameFromKey($fieldName);
-                if (!in_array(Inflector::tableize($tmpModelName), $this->_tables, true)) {
+                $this->getTableLocator()->get($tmpModelName);
+                $genericInstances = $this->getTableLocator()->genericInstances();
```
othercorey: Could also use get_class() to check the type returned by the above get() call.

src/Command/ModelCommand.php
```diff
@@ -346,11 +346,18 @@ public function findBelongsTo(Table $model, array $associations): array
                 ];
             } else {
                 $tmpModelName = $this->_modelNameFromKey($fieldName);
-                if (!in_array(Inflector::tableize($tmpModelName), $this->_tables, true)) {
+                $associationTable = $this->getTableLocator()->get($tmpModelName);
+                $tables = $this->listAll();
```
markstory, line 350: You should be able to use `$this->_tables` here as it was used here before.

ADmad: listAll() returns the value of $_tables if it's already populated else fetches the list from the db. I think using the method is fine here. That way it doesn't need to rely on listAll() being called before this method is invoked.

ajibarra: That's correct. I initially used $this->_tables but found unit tests failed when findBelongsTo was called isolated. generateAssociations calls listAll as well as findHasMany and findBelongsToMany methods but findBelongsTo didn't.

ADmad (approved):

PR #843, opened 2022-07-16
Add parser to find existing class details for non-destructive baking

othercorey:
This doesn't populate the attributes array until I decide what we need from them.

The code and docblocks remain indented. The expected indent size is 4 spaces. We can add more validation to the file formatting as needed.

### DIFF
--- a/tests/TestCase/Parse/CodeParserTest.php
+++ b/tests/TestCase/Parse/CodeParserTest.php
@@ -1,0 +1,116 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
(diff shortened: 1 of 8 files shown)


markstory:
Pretty neat stuff. Is the plan to use have bake generate the file as it does today with twig templates into a temporary location and then parse both the newly generated file and the current file and 'merge' the two structures together? With the merged structure I'm assuming we'll then create a dumper for the parsed class/method structure?

tests/TestCase/Parse/CodeParserTest.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->assertSame($code, $class->methods['initialize']->code);
+
+        $doc = <<<'DOCMARKER'
+            /**
+             * Initialize method
+             *
+             * @param array $config The configuration for the Table.
+             * @return void
+             */
+        DOCMARKER;
```
markstory, line 71: Do we also preserve the file header, and any other globally defined symbols in the file? Things like constants, `require`, or other classes/functions are things I could see userland code doing.

othercorey: I'm going to add a template for the file header to allow users to customize that. We'll re-generate the file structure but allow more customization by the user.

Other code that need to be saved off are user-specific properties, but I haven't worked that out yet.

othercorey:
> Pretty neat stuff. Is the plan to use have bake generate the file as it does today with twig templates into a temporary location and then parse both the newly generated file and the current file and 'merge' the two structures together? With the merged structure I'm assuming we'll then create a dumper for the parsed class/method structure?

I'm planning to split the twig templates into sections including file header, generated properties/methods and app-specific properties/methods.

I will merge the generated use statements and what I can detect of app-specific use statements into a sorted list.

PR #844, opened 2022-07-16
Redo FK Validation fix

LordSimal:
See #841

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -783,13 +783,20 @@ class ModelCommand extends BakeCommand
 
         $validate = [];
         $primaryKey = $schema->getPrimaryKey();
+        $foreignKeys = [];
+        if (isset($associations['belongsTo'])) {
+            foreach ($associations['belongsTo'] as $assoc) {
+                $foreignKeys[] = $assoc['foreignKey'];
+            }
+        }
         foreach ($fields as $fieldName) {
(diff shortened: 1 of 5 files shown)


src/Command/ModelCommand.php
```diff
@@ -804,14 +811,14 @@ public function getValidation(Table $model, array $associations, Arguments $args
...
-     * @param array<string> $primaryKey The primary key field. Unused because PK validation is skipped
+     * @param bool $isForeignKey Indicates if the currently precessed field is a foreign key
      * @return array Array of validation for the field.
      */
     public function fieldValidation(
         TableSchemaInterface $schema,
         string $fieldName,
         array $metaData,
-        array $primaryKey
+        bool $isForeignKey = false
```
markstory, line 821: Won't this run into BC errors with userland extensions? Should we put the `isForeignKey` attribute into `metaData` instead?

ADmad: This is targeting `3.x`, so we can make breaking changes. But yes we should target `2.x` and put the `isForeignKey` in the `metaData`.

LordSimal: So #841 should be fine in this regard

LordSimal:
Since we already put that info into the `$field` array in https://github.com/cakephp/bake/pull/841 I will close this

PR #845, opened 2022-07-17
Add file_header template

othercorey:
The namespace is included because spacing between namespace and file comment can be custom.

This converts only the Model templates until the format is approved.

### DIFF
--- a/templates/bake/element/file_header.twig
+++ b/templates/bake/element/file_header.twig
@@ -1,0 +1,4 @@
+<?php
+declare(strict_types=1);
+
+namespace {{ namespace }};
--- a/templates/bake/Model/behavior.twig
+++ b/templates/bake/Model/behavior.twig
@@ -13,10 +13,7 @@
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 #}
-<?php
-declare(strict_types=1);
-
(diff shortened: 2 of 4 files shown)


templates/bake/element/file_header.twig
```diff
@@ -0,0 +1,4 @@
+<?php
+declare(strict_types=1);
+
+namespace {{namespace}};
```
othercorey, line 4: Need to add space around variable

ADmad:
What's the benefit / larger picture for this change?

othercorey:
> What's the benefit / larger picture for this change?

It's part of an effort to create bake configuration. This just allows users to have custom file header spacing and file comment to match their project.

Any non-destructive baking would want to keep the original header.

othercorey:
If this change is ok, I'm going to update the remaining file templates

markstory (commented):
Seems ok to me. Should this template also include the `use` statements as a list?

othercorey:
> Seems ok to me. Should this template also include the `use` statements as a list?

I haven't finished that part of the prototype yet where I generate the use statements instead of hard-coded list in the twig file.

I'll provide the list via a template variable, but users would need to add their custom use statements in the command so I can sort them.

I don't know that actual php generation needs to be customized, but I could add to the header regardless.

PR #846, opened 2022-07-25
Use generic syntax

dereuromark:
Should we start adapting the newer syntax also in template files that are baked?
They would reflect it better

Also, the current syntax by some tools (e.g. code sniffer) could be changed to
```
array<\{{ entityClass }}>|\Cake\Collection\CollectionInterface
```
which would be wrong, as the object is now not part of the collection anymore.
Only the legacy syntax supported this understanding (as per legacy fallback interpretation).

Follows https://github.com/cakephp/cakephp/pull/16651

### DIFF
--- a/templates/bake/Template/index.twig
+++ b/templates/bake/Template/index.twig
@@ -16,7 +16,7 @@
 <?php
 /**
  * @var \{{ namespace }}\View\AppView $this
- * @var array<\{{ entityClass }}>|\Cake\Collection\CollectionInterface<\{{ entityClass }}> ${{ pluralVar }}
+ * @var iterable<\{{ entityClass }}> ${{ pluralVar }}
  */
 ?>
 <div class="{{ pluralVar }} index content">
--- a/templates/bake/Template/index.twig
+++ b/templates/bake/Template/index.twig
@@ -16,7 +16,7 @@
 <?php
 /**
  * @var \{{ namespace }}\View\AppView $this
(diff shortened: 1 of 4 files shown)


templates/bake/Template/index.twig
```diff
@@ -16,7 +16,7 @@
 <?php
 /**
  * @var \{{ namespace }}\View\AppView $this
- * @var \{{ entityClass }}[]|\Cake\Collection\CollectionInterface ${{ pluralVar }}
+ * @var array<\{{ entityClass }}>|\Cake\Collection\CollectionInterface<\{{ entityClass }}> ${{ pluralVar }}
```
ADmad: As seen [here](https://github.com/cakephp/cakephp/pull/16651/files#diff-beab97e6f7fcea692dab3a2b6ce9b208aca80beb7490820345e88a28f9fb7c39R2181) phpcs doesn't like the item/class type interfaces even if they extend an iterable type.

dereuromark: Refs https://github.com/cakephp/cakephp/pull/16651#pullrequestreview-1053719318

ADmad (approved):

dereuromark:
This is now also reflected like this in https://github.com/dereuromark/cakephp-ide-helper/releases/tag/1.18.0 for all existing code to be adjusted accordingly - if configured as such.

PR #851, opened 2022-09-05
Add non-destructive baking for tables to ModelCommand

othercorey:
This is the code that affects the Table class generation. These are the only changes to the template and command code so far. 

This does not support importing existing properties. That can be added after this is reviewed.


templates/bake/Model/table.twig
```diff
@@ -14,13 +14,9 @@
...
 {% set annotations = DocBlock.buildTableAnnotations(associations, associationInfo, behaviors, entity, namespace) %}
-<?php
-declare(strict_types=1);
-
-namespace {{ namespace }}\Model\Table;
-
-{% set uses = ['use Cake\\ORM\\Query;', 'use Cake\\ORM\\RulesChecker;', 'use Cake\\ORM\\Table;', 'use Cake\\Validation\\Validator;'] %}
-{{ uses|join('\n')|raw }}
+{% set generatedFunctions = ['initialize'] %}
+{% set classImports = fileBuilder.classImports(['Cake\\ORM\\Query', 'Cake\\ORM\\RulesChecker', 'Cake\\ORM\\Table', 'Cake\\Validation\\Validator']) %}
```
othercorey, line 18: This always add the imports even if the functions aren't generated below. However, this is how it was done previously.

templates/bake/element/file_header.twig
```diff
@@ -0,0 +1,5 @@
+<?php
+declare(strict_types=1);
+
+namespace {{ namespace }};
+{{- Bake.concat("\n", [Bake.classUseStatements(imports.class ?? []), Bake.functionUseStatements(imports.function ?? []), Bake.constUseStatements(imports.const ?? [])], '\n\n') }}
```
othercorey: The individual use statements could be rendered using an element instead of compiling with BakeHelper. 

The same goes for imported methods.

src/View/BakeView.php
```diff
@@ -45,7 +45,8 @@ class BakeView extends TwigView
     public function initialize(): void
     {
         $this->setConfig('environment', [
-          'cache' => false,
+        'autoescape' => false,
```
ADmad: Indentation looks off, surprising that phpcs didn't catch it.

templates/bake/Model/table.twig
```diff
@@ -81,6 +77,7 @@ class {{ name }}Table extends Table
 {{- "\n" }}
 
 {%- if validation %}
+{% set generatedFunctions = generatedFunctions|merge(['validationDefalt']) %}
```
ADmad: Typo in `validationDefalt`, should have been caught by a test case.

othercorey: Will need many more tests once we decide on the template changes.

templates/bake/Model/table.twig
othercorey: This is where I was talking about setting the imports in the command instead of the template.

src/Command/ModelCommand.php
```diff
@@ -1142,6 +1151,9 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
                 'You can use Plugin.name to bake plugin models.',
         ])->addOption('table', [
             'help' => 'The table name to use if you have non-conventional table names.',
+        ])->addOption('update', [
+            'boolean' => true,
+            'help' => 'Update existing file with generated code only.',
```
```suggestion
            'help' => 'Update generated methods in existing files. If the file doesn't exist it will be created.',
```

templates/bake/element/file_header.twig
markstory: I don't know if we need templates for use statements. There isn't much variance in how they are used.

tests/TestCase/Command/ModelCommandTest.php
```diff
@@ -1701,6 +1701,59 @@ public function testBakeWithRulesUnique()
...
+        class TodoItemsTable extends Table
+        {
+            /**
+             * Returns a rules checker object that will be used for validating
+             * application integrity.
+             *
+             * @param \Cake\ORM\RulesChecker $rules The rules object to be modified.
+             * @return \Cake\ORM\RulesChecker
+             */
+            public function buildRules(RulesChecker $rules): RulesChecker
```
markstory, line 1739: It would be good to have coverage on custom methods like a finder. It would also be good to have coverage for overwriting existing methods like replacing `initialize()`.

templates/bake/element/file_header.twig
othercorey: The question isn't customization but what's the best way to maintain this. I wasn't sure if writing code gen into BakeHelper was right. You see I need to do some specific whitespace manipulation with it to support 0 statements.

src/View/BakeView.php
othercorey: These were not indented 4 spaces before for some reason which messed up my editor.

othercorey:
Moved code gen into separate helper and reduced the public API for generation code. Removed options like ignoring certain functions that can be added when something requires it.

Cleaned up template generation so feels less cluttered.

Added several tests for the various cases of overwriting and keeping different functions.

src/View/Helper/CodeGenHelper.php
```diff
@@ -0,0 +1,141 @@
...
+ * Bake code generation helper
+ */
+class CodeGenHelper extends Helper
+{
+    /**
+     * Builds php code from class, function and const imports.
+     *
+     * @param array<string|int, string> $classImports Class imports
+     * @param array<string|int, string> $functionImports Class imports
+     * @param array<string|int, string> $constImports Class imports
```
ADmad:
```suggestion
     * @param array<string|int, string> $functionImports Function imports
     * @param array<string|int, string> $constImports Constant imports
```

src/View/Helper/CodeGenHelper.php
```diff
@@ -0,0 +1,141 @@
...
+        $constUses = $this->getUseStatements('use const', $constImports);
+
+        return $this->concat("\n", [$classUses, $functionUses, $constUses], $prefix, $suffix);
+    }
+
+    /**
+     * Builds array of php use statements from imports.
+     *
+     * @param string $prefix Prefix to put before import (function, const)
+     * @param array<string|int, string> $imports Class imports
```
ADmad:
```suggestion
     * @param array<string|int, string> $imports Imports
```

PR #852, opened 2022-09-07
non-destructive: Port to bake 2.x and php 7

othercorey:
This required fixing a number of psalm errors because psalm wasn't running in cake5 brnach.


othercorey:
I don't know what's causing the final psalm error

PR #858, opened 2022-09-09
Use file_header element in all file generation templates

othercorey:
This also adds support for no namespace to file_header template.

### DIFF
--- a/templates/bake/Command/command.twig
+++ b/templates/bake/Command/command.twig
@@ -13,15 +13,15 @@
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 #}
-<?php
-declare(strict_types=1);
-
-namespace {{ namespace }}\Command;
-
-use Cake\Command\Command;
-use Cake\Console\Arguments;
-use Cake\Console\ConsoleIo;
-use Cake\Console\ConsoleOptionParser;
+{{ element('Bake.file_header', {
+    namespace: "#{namespace}\\Command",
+    classImports: [
+        'Cake\\Command\\Command',
+        'Cake\\Console\\Arguments',
+        'Cake\\Console\\ConsoleIo',
+        'Cake\\Console\\ConsoleOptionParser',
+    ],
+}) }}
 
 /**
  * {{ name }} command.
--- a/templates/bake/Model/entity.twig
+++ b/templates/bake/Model/entity.twig
@@ -25,8 +25,10 @@
 {%- set accessible = Bake.getFieldAccessibility(fields, primaryKey) %}
 
 {%- set generatedProperties = [] %}
-{%- set uses = fileBuilder.uses(['Cake\\ORM\\Entity']) %}
-{{ element('Bake.file_header', {namespace: fileBuilder.namespace, uses: uses}) }}
+{{ element('Bake.file_header', {
+    namespace: fileBuilder.namespace,
+    classImports: fileBuilder.classImports(['Cake\\ORM\\Entity']),
+}) }}
 
 {{ DocBlock.classDescription(name, 'Entity', annotations)|raw }}
 class {{ name }} extends Entity
--- a/templates/bake/Model/table.twig
+++ b/templates/bake/Model/table.twig
@@ -15,8 +15,10 @@
 #}
 {% set annotations = DocBlock.buildTableAnnotations(associations, associationInfo, behaviors, entity, namespace) %}
 {% set generatedFunctions = ['initialize'] %}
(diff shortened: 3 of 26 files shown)


tests/TestCase/Command/PluginCommandTest.php
```diff
@@ -256,8 +256,8 @@ public function assertPluginContents($pluginName)
         );
 
         foreach ($comparisonFiles as $key => $file) {
-            $result = file_get_contents($file);
-            $this->assertSameAsFile($bakedFiles[$key], $result);
+            $result = file_get_contents($bakedFiles[$key]);
+            $this->assertSameAsFile($file, $result);
```
othercorey, line 260: Fixed the expected value so the diff looks correct.

templates/bake/element/file_header.twig
```diff
@@ -1,5 +1,4 @@
 <?php
 declare(strict_types=1);
-
-namespace {{ namespace }};
-{{-  Bake.concat('\n', [uses.class ?? [], uses.function ?? [], uses.const ?? []], '\n\n') }}
\ No newline at end of file
+{{- namespace ? "\n\nnamespace #{namespace};" : '' }}
+{{- Bake.concat('\n', [uses.class ?? [], uses.function ?? [], uses.const ?? []], '\n\n') }}
```
othercorey: This change was required to support bootstrap baking.

templates/bake/Command/command.twig
```diff
@@ -13,15 +13,18 @@
...
-
-use Cake\Command\Command;
-use Cake\Console\Arguments;
-use Cake\Console\ConsoleIo;
-use Cake\Console\ConsoleOptionParser;
+{{ element('Bake.file_header', {
+        namespace: "#{namespace}\\Command",
+        uses: {
+            class: [
+                'use Cake\\Command\\Command;',
```
ADmad: Having to prefix `use ` for all the classes seems a bit redundant.

othercorey: This ended up being the final iteration, but I tried others.

To avoid having to do awkward code-gen in the template, FileBuilder returns the code just like ClassBuilder does.

The class, function and const uses all have different prefixes, would you have had to code gen them separately.

I also need to allow users and non-model templates to add imports without the help of FileBuilder, so it ended up being arrays of staetments.

I could make them just lists of imports, but that requires a special array format and code-gen. It would also mean FileBuilder wouldn't return code. We bake imports with aliases, so that's not a hypothetical case.

othercorey: @markstory We should resolve this before beta

ADmad: Whether it's a class, function or const the `use` keyword will be the same for all. The `file_header` template could add that prefix instead of having to repeat it for each class/function/constant.

othercorey: That feels hacky. Then we'd have `function import` as the parameter which is neither code nor import.

We'd have to move back to arrays of imports/aliases.

templates/bake/Model/entity.twig
```diff
@@ -25,8 +25,11 @@
 {%- set accessible = Bake.getFieldAccessibility(fields, primaryKey) %}
 
 {%- set generatedProperties = [] %}
-{%- set uses = fileBuilder.uses(['Cake\\ORM\\Entity']) %}
-{{ element('Bake.file_header', {namespace: fileBuilder.namespace, uses: uses}) }}
+{{ element('Bake.file_header', {
+        namespace: fileBuilder.namespace,
+        uses: fileBuilder.uses(['Cake\\ORM\\Entity']),
```
markstory: Should this have `use`?

templates/bake/Model/table.twig
```diff
@@ -15,8 +15,11 @@
 #}
 {% set annotations = DocBlock.buildTableAnnotations(associations, associationInfo, behaviors, entity, namespace) %}
 {% set generatedFunctions = ['initialize'] %}
-{% set uses = fileBuilder.uses(['Cake\\ORM\\Query', 'Cake\\ORM\\RulesChecker', 'Cake\\ORM\\Table', 'Cake\\Validation\\Validator'], [], []) %}
-{{ element('Bake.file_header', {namespace: fileBuilder.namespace, uses: uses}) }}
+{{ element('Bake.file_header', {
+        namespace: fileBuilder.namespace,
+        uses: fileBuilder.uses(['Cake\\ORM\\Query', 'Cake\\ORM\\RulesChecker', 'Cake\\ORM\\Table', 'Cake\\Validation\\Validator'], [], []),
```
markstory: Should these have `use` on them?

templates/bake/Model/entity.twig
othercorey: No. This is passing an array of imports generated by code to FileBuilder which constructs  the use statements with the user-defined imports merged in.

Issue #709, opened 2020-09-22
Enhance cake bake to recognize unchanged files


lab-at-nohl:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.0
* Platform and Target: 

### What you did
rerun `bin/cake bake ...` after changing the schema

### What happened
cake asks for permission to overwrite existing files

### What you expected to happen
cake will overwrite files which have been baked before without confirmation

Possible solution: `cake bake` will store hashes of automatically created files. If file haven't been touched when running bake again, just overwrite it. Background: At least beginners experiment a lot, confusion about changed/unchanged files happens. A lot of work can get lost.

dereuromark:
I also disagree with this.
It adds a complexity and other pitfalls.

Instead, one should have committed the files (via Git for example), and then can quickly run -o (overwrite) to quickly regenerate, verify the diff (changes), apply certain undos (undo undesired hunks or reapply already done customizations from the history diff).
Thats how everyone pretty much does it quite efficiently these days.

markstory:
You can overwrite files using the `--force` option as well.

lab-at-nohl:
Hm, I still think bake should not ask to overwrite untouched files. At least I got used to overwrite all. I could go the other way around, how to see from outside if a file has been changed? One could go with commits, sure, but who does as many commits when starting something new?

I just returned to cake after some years (2.x or so). I still find the number of files confusing, you cannot look inside all before bake. Maybe I should quit using cake bake then...

dereuromark:
> using the --force option

oh, yes, not -o, but -f (I confuse them sometimes when not checking the CLI help)

> should not ask to overwrite untouched files

untouched is a very difficult word. What if the bake templates or helper code to generate those changed? Even a single white space change would automatically make your hash solution void.
In the end there is no alternative to what was already said - and using the force option is 2 char extra to have already full overwrite possibility

I don't think there is more to be done here at this point.

othercorey:
@lab-at-nohl You should probably only re-generate files if you never customize them. Only new files should be baked.

markstory:
> Hm, I still think bake should not ask to overwrite untouched files. At least I got used to overwrite all. I could go the other way around, how to see from outside if a file has been changed? One could go with commits, sure, but who does as many commits when starting something new?

[...]

PR #860, opened 2022-09-10
Skip prompting the user if entity or table file is unchanged

othercorey:
Refs https://github.com/cakephp/bake/issues/709

### DIFF
--- a/src/Command/BakeCommand.php
+++ b/src/Command/BakeCommand.php
@@ -212,4 +212,32 @@ abstract class BakeCommand extends Command
 
         return null;
     }
+
+    /**
+     * Write file contents out to path and prompt user with options with file exists.
+     *
+     * @param \Cake\Console\ConsoleIo $io Console io
+     * @param string $path The path to create the file at
+     * @param string $contents The contents to put into the file
+     * @param bool $forceOverwrite Whether the file should be overwritten without prompting the user
+     * @param bool $skipIfUnchnged Skip writing output if the contents match existing file
+     * @return bool True if successful, false otherwise
+     * @throws \Cake\Console\Exception\StopException When `q` is given as an answer
+     *   to whether a file should be overwritten.
+     */
+    protected function writeFile(
+        ConsoleIo $io,
+        string $path,
+        string $contents,
+        bool $forceOverwrite = false,
+        bool $skipIfUnchnged = true
+    ): bool {
+        if ($skipIfUnchnged && file_exists($path) && file_get_contents($path) === $contents) {
+            $io->info("Skipping update to `{$path}`. It already exists and would not change.");
+
+            return true;
+        }
+
+        return $io->createFile($path, $contents, $forceOverwrite);
+    }
 }
--- a/tests/TestCase/Command/ModelCommandTest.php
+++ b/tests/TestCase/Command/ModelCommandTest.php
@@ -1660,6 +1660,26 @@ class ModelCommandTest extends TestCase
     }
 
     /**
+     * Tests baking a file with no changes
+     *
+     * @return void
+     */
+    public function testBakeEntitySimpleUnchanged(): void
+    {
+        $this->generatedFile = APP . 'Model/Entity/User.php';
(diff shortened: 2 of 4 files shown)


src/Command/BakeCommand.php
```diff
@@ -196,4 +196,32 @@ protected function parseFile(string $path): ?ParsedFile
...
+     * @return bool True if successful, false otherwise
+     * @throws \Cake\Console\Exception\StopException When `q` is given as an answer
+     *   to whether a file should be overwritten.
+     */
+    protected function writeFile(
+        ConsoleIo $io,
+        string $path,
+        string $contents,
+        bool $forceOverwrite = false,
+        bool $skipSame = true,
```
othercorey: This could be an option to ConsoleIo::createFile()

othercorey:
We might need to support ignoring newline differences to support re-gen on windows with \n\r but the project saving as \n. This is currently why the windows test fails.

src/Command/BakeCommand.php
```diff
@@ -196,4 +196,32 @@ protected function parseFile(string $path): ?ParsedFile
...
+     * @return bool True if successful, false otherwise
+     * @throws \Cake\Console\Exception\StopException When `q` is given as an answer
+     *   to whether a file should be overwritten.
+     */
+    protected function writeFile(
+        ConsoleIo $io,
+        string $path,
+        string $contents,
+        bool $forceOverwrite = false,
+        bool $skipSame = true
```
ADmad:
```suggestion
        bool $skipIfSame = true
```
Sounds better

othercorey: I wasn't sure, but seeing that, I wonder if `skipIfUnchanged` is clearer. I used that name in the tests.

ADmad: `skipIfUnchanged` 👍

tests/TestCase/Command/ModelCommandTest.php
```diff
@@ -1659,6 +1659,26 @@ public function testBakeEntitySimple()
...
+    /**
+     * Tests baking a file with no changes
+     *
+     * @return void
+     */
+    public function testBakeEntitySimpleUnchanged(): void
+    {
+        $this->generatedFile = APP . 'Model/Entity/User.php';
+        $result = file_get_contents($this->_compareBasePath . __FUNCTION__ . '.php');
+        file_put_contents($this->generatedFile, str_replace("\r\n", "\n", $result));
```
othercorey, line 1671: This forces the newlines to match on windows.

src/Command/BakeCommand.php
```diff
@@ -212,4 +212,32 @@ protected function parseFile(string $path): ?ParsedFile
...
+     */
+    protected function writeFile(
+        ConsoleIo $io,
+        string $path,
+        string $contents,
+        bool $forceOverwrite = false,
+        bool $skipIfUnchnged = true
+    ): bool {
+        if ($skipIfUnchnged && file_exists($path) && file_get_contents($path) === $contents) {
+            $io->info("{$path} already exists but has no changes, skipping");
```
markstory:
```suggestion
            $io->info("Skipping update to `{$path}`. It already exists and would not change.");
```

tests/TestCase/Command/ModelCommandTest.php
```diff
@@ -1659,6 +1659,26 @@ public function testBakeEntitySimple()
...
+
+        $this->exec('bake model --no-test --no-fixture --no-table --no-fields --no-hidden users');
+
+        $this->assertExitCode(CommandInterface::CODE_SUCCESS);
+        $this->assertFileExists($this->generatedFile);
+        $this->assertSameAsFile(__FUNCTION__ . '.php', $result);
+
+        $this->assertOutputContains("{$this->generatedFile} already exists but has no changes, skipping");
```
markstory:
```suggestion
        $this->assertOutputContains("Skipping update to `{$this->path}`. It already exists and would not change.");
```

Issue #828, opened 2022-05-04
Bake Template: make fileextension configurable


fabian-mcfly:
Currently, the `php`-Extension for newly baked templates is hardcoded. Thus it's not possible to create other filetypes without overwriting the whole `TemplateCommand::bake` method.

https://github.com/cakephp/bake/blob/1c8a2d7b6a15f4bbc77bf15c5c780fbceb41de35/src/Command/TemplateCommand.php#L359

othercorey:
What other extensions would you use?

fabian-mcfly:
[Twig](https://github.com/cakephp/twig-view) for example.

othercorey:
This will be configurable with `TemplateCommand::$ext` in 2.8.

PR #861, opened 2022-09-10
Make bake template extension configurable

othercorey:
closes https://github.com/cakephp/bake/issues/828

Users would need to extend TemplateCommand and use it for their custom template baking.

Should this be a protected function instead or a config?

### DIFF
--- a/src/Command/TemplateCommand.php
+++ b/src/Command/TemplateCommand.php
@@ -86,6 +86,13 @@ class TemplateCommand extends BakeCommand
     public $path;
 
     /**
+     * Output extension
+     *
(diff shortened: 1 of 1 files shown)


src/Command/TemplateCommand.php
```diff
@@ -85,6 +85,13 @@ class TemplateCommand extends BakeCommand
      */
     public $path;
 
+    /**
+     * Output extension
+     *
+     * @var string
+     */
+    public $outputExt = 'php';
```
othercorey: Should this just be `$ext` even though it's not part of `$path` ?

Issue #721, opened 2020-11-04
Custom helpers for twig templates


skie:
Custom helpers for twig templates could be needed when user want to add more logic to templates, which is not possible to implement in twig templates.

We can consider two options:

1. Find way to pass custom helper list in TemplateRenderer layer.
2. Allow to define custom TemplateRenderer class which loaded in all commands.

Lets discuss what is the best way to add this feature.

ADmad:
We could add a `SimpleBakeCommand::templateRenderer()` which returns the renderer instance so people can use `$this->templateRenderer()->viewBuilder()`  in  `execute()` (or any other method) to setup the view builder.

Then `SimpleBakeCommand::bake()` would be also updated to use `templateRenderer()` instead of creating an instance itself.

othercorey:
You can handle the bake init event to add helpers, but it doesn't let you customize the view builder before creating.

PR #862, opened 2022-09-10
Add createTemplateRenderer() to BakeCommand

othercorey:
closes https://github.com/cakephp/bake/issues/721

This allows commands to add custom templates and other ViewBuilder options for each TemplateRenderer used.

### DIFF
--- a/src/Command/BakeCommand.php
+++ b/src/Command/BakeCommand.php
@@ -19,11 +19,14 @@ namespace Bake\Command;
 use Bake\CodeGen\CodeParser;
 use Bake\CodeGen\ParsedFile;
 use Bake\Utility\CommonOptionsTrait;
+use Bake\Utility\TemplateRenderer;
 use Cake\Command\Command;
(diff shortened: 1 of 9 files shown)


src/Command/BakeCommand.php
```diff
@@ -152,6 +155,19 @@ public function getTemplatePath(Arguments $args, ?string $container = null): str
         return str_replace('/', DIRECTORY_SEPARATOR, $path);
     }
 
+    /**
+     * Creates a new instance of TemplateRenderer with theme set.
+     *
+     * @return \Bake\Utility\TemplateRenderer
+     */
+    public function getTemplateRenderer(): TemplateRenderer
```
ADmad: `get` is not a suitable prefix if the method is going to build a new instance each time. `build`, `create` would be better.

othercorey: Sounds good. The code changed and I left the name.

PR #863, opened 2022-09-10
Add $force to CommonOptionsTrait

othercorey:


src/Utility/CommonOptionsTrait.php
```diff
@@ -45,6 +45,11 @@ trait CommonOptionsTrait
      */
     public $connection;
 
+    /**
+     * @var bool
+     */
+    public $force = false;
```
othercorey, line 51: The default value is required because we have some hacky unit tests that call part of the Command instead of baking the file.

PR #866, opened 2022-09-14
Fix wrong connection used when detecting belongsTo associations

othercorey:
This doesn't set the table name, but that doesn't seem to be part of the logic.


othercorey:
Doesn't look like we actually test with a custom connection. We probably need to drop the default connection to really test properly.

PR #867, opened 2022-09-19
Update tests to match cake5 generation

othercorey:


othercorey:
Tests no longer run due to a phpunit bug when mocking methods that return `never`.

markstory (approved):

PR #872, opened 2022-09-26
Update GA CI config.

ADmad:


othercorey:
We are just changing the CI config? If we aren't testing 7.2 then we should drop support.

othercorey:
> We are just changing the CI config? If we aren't testing 7.2 then we should drop support.

My mistake. The review cut off the php 7.2 config :)

PR #873, opened 2022-10-01
Add support for keeping custom class implements with --update

othercorey:
This reduces the code lost when baking entities that extend interfaces such as IdentityInterface.


othercorey:
I don't know if user tables ever implement an interface, but the change to table.twig is small.

ADmad (approved):

PR #876, opened 2022-10-10
Added `Middleware` to `TestCommand::classTypes`

### LINKED ISSUE #875: Baking Middleware should bake Middleware tests
### Description

I baked my first Middleware 🥳 .
I was surprised to see that, unlike other bakeable classes, Middleware do not bake their corresponding PHPUnit test stubs 😭 .
It would be cool if the test stubs were also generated as part of baking Middleware.

angelxmoreno:
resolves #875 

with this PR we can do
```shell
./bin/cake bake test middleware Sample1
```
Also, when baking a middleware, the test will automatically be generated

### DIFF
--- a/src/Command/TestCommand.php
+++ b/src/Command/TestCommand.php
@@ -57,6 +57,7 @@ class TestCommand extends BakeCommand
         'Mailer' => 'Mailer',
         'Command' => 'Command',
         'CommandHelper' => 'Command\Helper',
+        'Middleware' => 'Middleware',
     ];
 
     /**
@@ -79,6 +80,7 @@ class TestCommand extends BakeCommand
         'Mailer' => 'Mailer',
         'Command' => 'Command',
         'CommandHelper' => 'Helper',
+        'Middleware' => 'Middleware',
     ];
 
     /**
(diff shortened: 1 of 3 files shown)


angelxmoreno:
The baked test looks like this:
```php
<?php
declare(strict_types=1);

namespace App\Test\TestCase\Middleware;

use App\Middleware\Sample1Middleware;
use Cake\TestSuite\TestCase;

/**
 * App\Middleware\Sample1Middleware Test Case
 */
class Sample1MiddlewareTest extends TestCase
{
    /**
     * Test subject
     *
     * @var \App\Middleware\Sample1Middleware
     */
    protected $Sample1;

    /**
     * Test process method
     *
     * @return void
     * @uses \App\Middleware\Sample1Middleware::process()
     */
    public function testProcess(): void
    {
    }
}
```
Is this a sufficient stub?

angelxmoreno:
um , what is up with these tests? I'll work on fixing them

ADmad:
Now `--no-test` option should be added for the `MiddlewareCommand` as it's done for other commands which generate test classes.

PR #878, opened 2022-11-10
3.x: change baked index method to use query instance instead of table

LordSimal:
This is my proposal to change the default generated `paginate()` method to use a query instance instead of the table instance.

I never really used that approach because in the end I always created a query instance which then gets adjusted/modified/extended by other typical query methods (`where()`, `contain()`, `orderDesc()` etc.) and therefore this would just be a little more clearer for newbies as well instead of the "magic" that `paginate()` just uses the table instance and the `paginate` property of the current controller.

Sure one can easily create their own bake template where they can overwrite pretty much everything but that will only happen after spending quite some time with the framework.

Curious what you guys have to say about that 😄

### DIFF
--- a/templates/bake/element/Controller/index.twig
+++ b/templates/bake/element/Controller/index.twig
@@ -22,11 +22,12 @@
     {
 {% set belongsTo = Bake.aliasExtractor(modelObj, 'BelongsTo') %}
 {% if belongsTo %}
-        $this->paginate = [
-            'contain' => {{ Bake.exportArray(belongsTo)|raw }},
-        ];
+        $query = $this->{{ currentModelName }}->find()
+            ->contain({{ Bake.exportArray(belongsTo)|raw }});
+{% else %}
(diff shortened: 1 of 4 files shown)


dereuromark:
We should then find a way for https://github.com/dereuromark/cakephp-ide-helper/issues/292 or it decreases usability.
At least I always used the table if the model was the same as the controller so far.

LordSimal:
Sure, adding that functionality to the IDE-Helper is a nice to have but since the IDE-Helper is not pre-installed with cake it will not be used by most newcomers if they don't get it recommended by the support chat, online forums, their own desire to add advanced IDE support or something else.

My intend here is to remove a bit of the magic that is behind CakePHP and make it more clearer just by reading the code what is happening, where and why.

The `index` method is the only place where a query is really "used/needed" and expected by users so the fact that a baked controller doesn't even contain the word `query` is a bit strange to me.

ADmad (approved):

PR #879, opened 2022-11-15
Fix up closing of queries.

dereuromark:
I dont see the point in using all() for those array lists if it doesnt solve the issue of it being handled afterwards

> Cannot use object of type Cake\Datasource\ResultSetDecorator as array

Using toArray() is IMO expected here to close this instead to have an array to work with and easily append or prepend elements for the list.

### DIFF
--- a/templates/bake/element/Controller/add.twig
+++ b/templates/bake/element/Controller/add.twig
@@ -37,7 +37,7 @@
 {%- for assoc in associations %}
     {%- set otherName = Bake.getAssociatedTableAlias(modelObj, assoc) %}
     {%- set otherPlural = otherName|variable %}
-        ${{ otherPlural }} = $this->{{ currentModelName }}->{{ otherName }}->find('list', ['limit' => 200])->all();
+        ${{ otherPlural }} = $this->{{ currentModelName }}->{{ otherName }}->find('list', ['limit' => 200])->toArray();
         {{- "\n" }}
     {%- set compact = compact|merge(["'#{otherPlural}'"]) %}
 {% endfor %}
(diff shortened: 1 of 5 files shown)


ADmad:
This should go to `2.next`. There have been times when I have used the collection methods on result of `find('list')`.

dereuromark:
But wouldnt this merged just fix new code?
We dont even really overwrite existing, so it seems safe to fix up "new code to be generated", no?

ADmad:
Patch releases shouldn't be generating different output (unless we are fixing an actual bug) even if we are not overwriting anything.

dereuromark:
OK
Done :)
Seems it also merges 2.x into 2.next now.

dereuromark:
Are we OK with the merge? @ADmad

dereuromark:
I still think that bake code could easily go into current 2.x since it doesnt affect any existing code.
And for new code there is not an issue IMO.

ADmad:
@dereuromark I am okay with using `toArray()` instead of `all()` for `find('list')` if that's the consensus.

The PR seems to have additional commits unrelated to your changes.

dereuromark:
See my comment, the branch base change will just merge 2.x into 2.next as well

dereuromark:
Should we continue?

ADmad:
After working on a 4.x => 5.x upgrade I am not in favor of this change.

Since the ORM by default returns a resultset object we should use that wherever possible. Converting only `find('list')` results to array adds unnecessary cognitive load and if you need to perform additional conversion on the view variable you have to check the controller code to see whether it sets an object or array. Conversion to array should be done only when there's explicit need for it.

dereuromark:
Given the 5.x code this indeed is now less useful and needs more migrating.

PR #897, opened 2022-12-07
updates root folder links

Arhell:


othercorey:
We will just merge into 2.next

PR #898, opened 2023-01-06
Remove obsolete use Cake\ORM\Query;

cleptric:


ADmad:
This should target `3.x` which is the branch compatible with Cake 5.

cleptric:
I noticed that in a Cake 4 app, using bake 2

dereuromark:
To be fair: Additional use statements that are unused are usually quite easy to be cleaned up by the sniffer/fixer ruleset
And maybe someone has additional "custom" rules to add certain hooks that could need them for now.

So maybe this is indeed of more use for Cake 5.x

ADmad:
The additional unused `use` in 4.x are intentional to help newbies who often forget adding them after adding callback methods.

PR #902, opened 2023-01-10
Cleanup of signature docs

dereuromark:
Those always return a value (since they never render a template), so never void

### DIFF
--- a/templates/bake/element/Controller/delete.twig
+++ b/templates/bake/element/Controller/delete.twig
@@ -17,7 +17,7 @@
      * Delete method
      *
      * @param string|null $id {{ singularHumanName }} id.
-     * @return \Cake\Http\Response|null|void Redirects to index.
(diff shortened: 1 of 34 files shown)


templates/bake/element/Controller/delete.twig
```diff
@@ -17,7 +17,7 @@
      * Delete method
      *
      * @param string|null $id {{ singularHumanName }} id.
-     * @return \Cake\Http\Response|null|void Redirects to index.
+     * @return \Cake\Http\Response|null Redirects to index.
      * @throws \Cake\Datasource\Exception\RecordNotFoundException When record not found.
      */
     public function delete($id = null)
```
othercorey, line 23: We could probably generate the actual return type since `mixed|void` can be narrowed to `?Response`.

But, this works in general.

dereuromark:
Stan fails are unrelated

PR #905, opened 2023-02-14
3.x: narrow down baked controller return types

LordSimal:
This PR adds return types to baked controllers `index`, `view` and `delete` methods since they can be defined as such.

`add` and `edit` need to stay as they are (unless we want to change that)

Also: We shouldn't generate PHPDoc's if the generated code doesn't represent that.

### DIFF
--- a/tests/comparisons/Controller/testBakeActionsContent.php
+++ b/tests/comparisons/Controller/testBakeActionsContent.php
@@ -3,6 +3,8 @@ declare(strict_types=1);
 
 namespace Bake\Test\App\Controller;
 
+use Cake\Http\Response;
+
 /**
  * BakeArticles Controller
  *
@@ -13,9 +15,9 @@ class BakeArticlesController extends AppController
     /**
      * Index method
      *
-     * @return \Cake\Http\Response|null|void Renders view
+     * @return void Renders view
      */
-    public function index()
+    public function index(): void
     {
         $query = $this->BakeArticles->find()
             ->contain(['BakeUsers']);
@@ -28,10 +30,10 @@ class BakeArticlesController extends AppController
      * View method
      *
      * @param string|null $id Bake Article id.
-     * @return \Cake\Http\Response|null|void Renders view
+     * @return void Renders view
      * @throws \Cake\Datasource\Exception\RecordNotFoundException When record not found.
      */
-    public function view($id = null)
+    public function view($id = null): void
     {
         $bakeArticle = $this->BakeArticles->get($id, [
             'contain' => ['BakeUsers', 'BakeTags', 'BakeComments'],
@@ -95,7 +97,7 @@ class BakeArticlesController extends AppController
      * @return \Cake\Http\Response|null Redirects to index.
      * @throws \Cake\Datasource\Exception\RecordNotFoundException When record not found.
      */
-    public function delete($id = null)
+    public function delete($id = null): ?Response
     {
         $this->request->allowMethod(['post', 'delete']);
         $bakeArticle = $this->BakeArticles->get($id);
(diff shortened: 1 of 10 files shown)


tests/comparisons/Controller/testBakeActionsContent.php
```diff
@@ -13,9 +15,9 @@ class BakeArticlesController extends AppController
     /**
      * Index method
      *
-     * @return \Cake\Http\Response|null|void Renders view
+     * @return void Renders view
      */
-    public function index()
+    public function index(): void
```
markstory, line 20: I think we avoided adding hints to these methods before because `?Response` doesn't include null and there isn't a way to have a union between a type and `void`.

Should we these `void` actions start ending with `return $this->render()` and add the `?Response` return type to make it more obvious what is going on and set us up for a smoother upgrade in the future?

tests/comparisons/Controller/testBakeActionsContent.php
```diff
@@ -95,7 +97,7 @@ public function edit($id = null)
      * @return \Cake\Http\Response|null Redirects to index.
      * @throws \Cake\Datasource\Exception\RecordNotFoundException When record not found.
      */
-    public function delete($id = null)
+    public function delete($id = null): ?Response
```
markstory, line 100: This one seems good to me.

tests/comparisons/Controller/testBakeActionsContent.php
LordSimal: Indeed `void` isn't allowed be part of union return type (currently). Therefore we can't do `?Response|void`
I personally would prefer adding the `return $this->render();` at the end of each generated controller action (except `delete` of course) to enable us to have `?Response` for `add` and `edit` as well as just `Response` for `index` and `view` as a return type.
Users have the ability right now to do so - it just depends on if we want to enforce that in the future or not.

dereuromark: I personally would avoid setting the types here.
It is not uncommon to have to redirect in those actions (permissions, ...), and then you need to completely rewrite the types here.
It is easier IMO to leave the range open.

dereuromark: `return $this->render();` sounds like quite some noise and overhead for no real value gain?

ADmad: Auto rendering is one magic feature I am quite conflicted about 🙂. While having an explicit statement to render and return is cleaner/easier to grok, I am not too thrilled about adding those to all actions.

ADmad: I would simply leave out the explicit type instead of adding `void` as it doesn't add much value. There's no app code which calls action method and Controller::invokeAction() already throws an exception if the method returns anything besides ResponseInterface (or null).

LordSimal: My intention here was more like making the auto-generated code base a bit more uniform with how `5.x` has been treated till now. Especially since we just added return types and property types to basically everything in `5.x` it just looks weird to me to not have it for controller actions as well.

I do though understand that it doesn't really provide a benefit since (as you already said) controller actions don't get called by any other parts of the codebase beside `Controller::invokeAction()`.

dereuromark: Controller and command code, both.
They are the main glue of the layers and especially here it can be useful to allow a more lean approach in dev.
At least thats how we were used to it so far :)

LordSimal:
The fact that controller methods can't/shouldn't be called directly via any other code parts results in no real benefit adding proper return types to controller actions. Therefore this PR can be closed and we leave the system as it is.

PR #911, opened 2023-02-21
Fix error when trying to bake with non-existent plugins directory.

### LINKED ISSUE #909: Incorrect message when `plugins` directory not exists using bake
### Description

During creating new plugin using command `bin/cake bake plugin PluginName` with not existed `plugins` directory.

Bake:
- no returns information for required `plugins` directory;
- returns strange information:
```
[Cake\Core\Exception\MissingPluginException] Plugin  could not be found. in /var/www/html/vendor/cakephp/cakephp/src/Core/PluginCollection.php on line 143
Exception Attributes: array (
  'plugin' => '',
)
Stack Trace:
- /var/www/html/vendor/cakephp/cakephp/src/Core/PluginCollection.php:257
...
```

### Bake Version

2.8.2

### CakePHP Version

4.4.11

ADmad:
This was caused due to `Bake.theme` option defaulting to empty string instead of null and hence a plugin with empty name was being searched causing an exception.

Closes #909

### DIFF
--- a/src/Command/PluginCommand.php
+++ b/src/Command/PluginCommand.php
@@ -383,7 +383,7 @@ class PluginCommand extends BakeCommand
         ])->addOption('theme', [
             'short' => 't',
             'help' => 'The theme to use when baking code.',
-            'default' => Configure::read('Bake.theme') ?? '',
+            'default' => Configure::read('Bake.theme') ?: null,
             'choices' => $this->_getBakeThemes(),
         ]);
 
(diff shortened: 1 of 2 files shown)


LordSimal:
I don't understand why `Configure::read('Bake.theme') ?: null,` needs to be specified here.
If `Bake.theme` is not present `Configure::read` will return null anyways. Why would just `Configure::read('Bake.theme')` not work?
It seems I am missing something...

dereuromark:
The theme could also be empty string

ADmad:
@LordSimal It's a safe guard against someone setting the theme to an empty string.

LordSimal:
Just saying: Even with this fix I still get the error from the issue when trying to bake a plugin with a non existent `plugins` directory...

LordSimal:
Forget that... I had `$this->addPlugin('PluginName');` in my Application.php which I forgot to remove...

Issue #912, opened 2023-03-10
Column with *_id not generating a `belongsTo` relation anymore.


passchn:
### Description

I guess this is a feature which was introduced recently. So if you e.g. have a table `articles` and a field `image_id`, the generated `ArticlesTable` no longer has a relation to `Images`, if there is not actually an `images` table in the database. 

This "bug" was practical for my assets plugin which uses these "fake relations", as described here: 
https://github.com/brandcom/cakephp-assets/tree/master#use-alibi-names 

Is there any possibility to generate the relation code anyway, e.g. by using a flag when running the command? Could be something like `bin/cake bake model users --no-table-check` or  something like that. 

That would be great to opt in into the previous behaviour again. 

### Bake Version

2.8.2

### PHP Version

4.4.11, php 8.1.1

passchn:
I downgraded to the plugin version 2.6 and these are the code differences after running `bin/cake bake model articles` again: 

![Bildschirm­foto 2023-03-10 um 19 41 05](https://user-images.githubusercontent.com/77938819/224398762-39ee353d-7ccc-48b7-a491-119d93ecc390.png)

dereuromark:
I think you can make a PR that would allow for this.
It is unlikely someone else would come forward in a reasonable time frame for such specific requests.

PR #913, opened 2023-03-10
Add `allow-alias-relations` option to bake model command

passchn:
Since PR https://github.com/cakephp/bake/pull/842 which was merged into [2.7.1](https://github.com/cakephp/bake/releases/tag/2.7.1), the bake model command skips the generation of relation-code, if the table does not actually exist in the database. 

This might be unwanted behaviour for some more edge-case scenarios, e.g. when one table is used for multiple alias relations. 

E.g., if you have an `addresses` table, you might want to have a field `delivery_address_id` and `invoice_address_id` on your `customers` table, while using the same table and `Address` entity for both of them. You would just have to set the `className` option for the two relations, so that they both point to a field `addresses.id`. 

Since `2.7.1`, the relations are not baked anymore, because there is no actual `delivery_addresses` table in the database.

This PR introduces a new option `--allow-alias-relations` to skip the check for an existing table and to bake the relation code anyway. 

Related issue: 
https://github.com/cakephp/bake/issues/912

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -363,7 +363,7 @@ class ModelCommand extends BakeCommand
                     get_class($associationTable) === Table::class &&
                     !in_array(Inflector::tableize($tmpModelName), $tables, true)
                 ) {
-                    $allowAliasRelations = $args && $args->getOption('allow-alias-relations');
+                    $allowAliasRelations = $args && $args->getOption('skip-relation-check');
                     $found = $this->findTableReferencedBy($schema, $fieldName);
                     if ($found) {
                         $tmpModelName = Inflector::camelize($found);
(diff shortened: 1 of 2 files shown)


dereuromark:
I wonder if `--skip-relation-check` might be clearer.
But lets wait on further feedback from others first.

In general I like having it, since it has less change of failing if there is a unresolvable kind of relation in there.
So a good fallback before having to do things manually in such edge cases.

passchn:
I agree! I also have a lot of tables where I just save an id to, say, a youtube video. I at some point started to name the fields `youtubeid` instead of `youtube_id` to prevent bake from generating wrong code. So I appreciate the new behaviour to check tables first. 

What doesn't work with this PR though is e.g. if you have a `users` table with a `delivery_address_id`, `invoice_address_id` and, say, a `twitter_id` for whatever reason. Then you want to have two hasOne alibi-relations to `addresses`, but just a field for `twitter_id`.

passchn:
Just checked the documentation. The term `alias` or `table alias` is already used for the name of each relation, also for "real" relations with existing tables: <https://book.cakephp.org/4/en/orm/associations.html> 

So `--allow-alias-relations` is nonsensical.

markstory:
What about having the option name reflect that it will generate associations for tables that don't exist? Names like `include-missing` or `include-all-associations` might work better :shrug:

passchn:
Hmm.. if you save twitter user ids as `twitter_id`, that is not really a "missing" relation.

passchn:
I personally would prefer having a word describing that kind of relation. "Alibi" is already taken. Maybe "Pseudo-Relation" but that sounds odd. 

That name could also be used in [the docs](https://book.cakephp.org/4/en/orm/associations.html) for the examples `UnapprovedComments` and `SubCategories`.

dereuromark:
I am still for `--skip-relation-check` to just reflect that the check will not be done, and any kind of such relation, existing or not is being added.

passchn:
I renamed the option now.

markstory:
Thank you 🎉

PR #915, opened 2023-04-20
Fixes application Bake templates path in Document

nojimage:
In the documentation, there is a possibility that users may place files incorrectly, because the template path on the application side does not include "bake".


markstory:
Thank you 🎉

PR #918, opened 2023-04-26
Fix CS errors display on PRs

ADmad:


dereuromark:
Something weird is going on if u enable this

ADmad:
The `p` flag in the config file is breaking the output XML

dereuromark:
I guess we can remove the p then
Or we dont need the Display in PRs, since it kind of also distracts from actual human input

ADmad:
> Or we dont need the Display in PRs, since it kind of also distracts from actual human input

I certainly prefer having the errors displayed on the "Files Changed" tab of the PR instead of having to navigate to the action's log to see the errors. A lot of users might not even know how to navigate to the appropriate location in the logs.

dereuromark:
For CI we could just use custom command then
So normal usage via composer Script can keep progress

ADmad:
The problem is phpcs will always use the phpcs.xml file. Will have to figure out how to override the booelan `p` flag from the command line or make it ignore the config file.

dereuromark:
Or we remove it there, and just add it to the composer script part (`-p` I guess).

dereuromark:
I checked it for Migrations: https://github.com/cakephp/migrations/pull/612
Seems to work fine for manual run, but also now fine for CI

PR #927, opened 2023-06-05
3.x

vnt:
Hello,
In this new version, will we also have a new CSS framework?
Would it be nice to have bootstrap, tailwind or material desiging?


markstory:
>  Would it be nice to have bootstrap, tailwind or material desiging?

That isn't planned. We've been using milligram for a few years now as it doesn't require build tooling and is small enough that it can be removed and replaced by application developers that want to use something else.

vnt:
>  Would it be nice to have bootstrap, tailwind or material desiging?

That isn't planned. We've been using milligram for a few years now as it doesn't require build tooling and is small enough that it can be removed and replaced by application developers that want to use something else.

_Originally posted by @markstory in https://github.com/cakephp/bake/issues/927#issuecomment-1576939964_
            
Milligram development has been stopped for 3 years, maybe the project is dead
https://github.com/milligram/milligram/tags

It would be great to have CakeBake v.3  with native support to a CSS framework such as Bootstrap or Material Design.
Thanks a lot.

LordSimal:
@vnt there is https://github.com/FriendsOfCake/bootstrap-ui for your Bootstrap needs which also has bake templates so I'd recommend you take a look at that. I am not aware of any Material Design plugins for CakePHP which do the same thing though.

It may very well be the case that milligram has stopped development but its stable, its small, its easy to replace and its all we need for the moment.

Also I would recommend you discuss such things with us in our [slack or discord server ](https://cakephp.org/get-involved#getHelp)instead of providing a PR which merges 3.x into 2.x which will never happen 😉

PR #938, opened 2023-09-10
Try new dep

dereuromark:
Outdated dependency alert:

  - brick/varexporter (0.3.8) latest is 0.4.0
  
According to https://github.com/brick/varexporter/releases/tag/0.4.0 no real changes...
It was a bit nonsense to release a new major here

Anyway, we should probably also backport this to 2.x as `0.3 || 0.4`


ADmad:
> Anyway, we should probably also backport this to 2.x as 0.3 || 0.4

👍

PR #948, opened 2023-10-04
Update DocBlockHelper.php allow different date and time classes

markusramsak:
since CakePHP 5.0 the overwritten DateTime or Date class is not used anymore for annotation.
I can't get Checks to pass, but I think it is clear what I want to achieve. It worked in the previous major version.

### DIFF
--- a/src/View/Helper/DocBlockHelper.php
+++ b/src/View/Helper/DocBlockHelper.php
@@ -226,8 +226,8 @@ class DocBlockHelper extends Helper
 
             case 'time':
                 $dbType = TypeFactory::build($type);
-                if (method_exists($dbType, 'getDateTimeClassName')) {
-                    return '\\' . $dbType->getDateTimeClassName();
+                if (method_exists($dbType, 'getTimeClassName')) {
+                    return '\\' . $dbType->getTimeClassName();
                 }
 
                 return '\Cake\I18n\Time';
--- a/src/View/Helper/DocBlockHelper.php
+++ b/src/View/Helper/DocBlockHelper.php
@@ -204,6 +204,8 @@ class DocBlockHelper extends Helper
                 $dbType = TypeFactory::build($type);
                 if (method_exists($dbType, 'getDateClassName')) {
                     return '\\' . $dbType->getDateClassName();
+                } elseif (method_exists($dbType, 'getDateTimeClassName')) {
+                    return '\\' . $dbType->getDateTimeClassName();
                 }
 
                 return '\Cake\I18n\Date';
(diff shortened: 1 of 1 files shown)


markstory (commented):
It would be good to have a test for this. You could have the date/time types mapped to the PHP DateTime classes.

markusramsak:
sadly I am not able to write tests because of lack of knowledge

dereuromark:
@markusramsak 
The test docs should give you some insights
It is actually straight forward

You checkout your fork of this repo standalone, checkout your branch here, run

    composer update
    vendor/bin/phpunit YourTestFileClass.php

and adjust it with an additional test...() method that you want to test until it is green based
on your changes

dereuromark:
Also note that PHPStan still has an issue with the code it seems:

    Instanceof between mixed and Cake\Chronos\Chronos will always evaluate to false.

markusramsak:
I don't know how to solve it. It was changed for CakePHP 5 from @LordSimal but it is a regression because now it doesn't work anymore with overwritten DateTime oder Date class. I don't know why this change was made in the first place ...

dereuromark:
Ah ok, so this fail is unrelated to your PR and already in the main branch - then we can ignore this one for now :)

markusramsak:
I know I am missing the tests but the changes allow using custom Time/Date/DateTime classes which are used for annotations. Without this using only CakePHP default classes are used for annotation of entities which is regression to CakePHP 4.x and related bake version.
The change in https://github.com/cakephp/cakephp/pull/17370 is related to this change.

markusramsak:
How do I write a Test to confirm that my suggested code works? I want that Date, Time and DateTime classes could be overwritten (like before). How do I test this? It seems like a static validation to the Cake-Classes in the tests in DocBlockHelperTest.php at the moment ...

LordSimal:
This was removed in https://github.com/cakephp/bake/pull/929 because I thought this wasn't necessary anymore but clearly I was wrong 😅

markusramsak:
Please could help me someone to get the tests working for this pull request so this change can be merged. I cannot update my Entity-Docs anymore until this change is merged. This change is actually nothing new - it was like that before for CakePHP 4.x

LordSimal:
We can take care of a testcase later. I'll just do a new release to add this feature back

markusramsak:
> We can take care of a testcase later. I'll just do a new release to add this feature back

thanks for your help

LordSimal:
Sorry for taking so long to get this back in

markusramsak:
> Sorry for taking so long to get this back in

We are all doing this in our free time and have other obligations. I completely understand.
By the way I am from Austria too. I am from Kärnten / Klopeiner See.

dereuromark:
@LordSimal Do we also need a [IdeHelper](https://github.com/dereuromark/cakephp-ide-helper) update for this?

LordSimal:
will have to check later. My initial problem was just the fact, that `date` columns were mapped to the Time class which shouldn't happen with this change

PR #950, opened 2023-10-04
Add @phpcsSuppress to `Command::execute`

cleptric:
PHPCS will complain about missing return-type hints when you bake your first command. 

```php
Method \App\Command\AppCommand::execute() does not have native return type hint for its return value but it should be possible to add it based on @return
    |       |     annotation "int|null|void".
```

PHPCBF then adds `: int|null|null` 🤔

### DIFF
--- a/templates/bake/Command/command.twig
+++ b/templates/bake/Command/command.twig
@@ -48,6 +48,7 @@ class {{ name }}Command extends Command
      * @param \Cake\Console\Arguments $args The command arguments.
      * @param \Cake\Console\ConsoleIo $io The console io
      * @return null|void|int The exit code or null for success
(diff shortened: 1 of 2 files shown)


dereuromark:
Is there no way to fix the actual phpcs rule to not throw wrong error?

dereuromark:
I would recommend opening a Bug ticket or pr for their repo

ADmad:
We can just bake with `return Command::SUCCESS`.

othercorey:
Re-opened so we can follow up with the recomendation.

dereuromark:
That PHPCS is most likely fixed, this was a tooling issue on the CS ruleset.
Closing then for now, if it is still present and the sniffer not fixable we can rediscuss.

Issue #951, opened 2023-10-05
Cannot bake ControllerTest with model type other than Table


asgraf:
### Description

Steps to reproduce:
1) Create any controller with model type other than Table
```php
class NbpRatesController extends AppController
{
    public $modelClass = 'NbpApi.NbpRates';
    public $_modelType = 'Endpoint';
  
   //some other code
}
```
2) bake unit test for that controller

```
application@d6af9a97ebb9:/application$ bin/cake bake test Controller --prefix=Admin --plugin=SuperAdmin NbpRates
Bake is detecting possible fixtures...

2023-10-05 16:33:22 error: [TypeError] Bake\Command\TestCommand::_processModel(): Argument cakephp/cakephp#1 ($subject) must be of type Cake\ORM\Table, NbpApi\Model\Endpoint\NbpRatesEndpoint given, called in /application/vendor/cakephp/bake/src/Command/TestCommand.php on line 502 in /application/vendor/cakephp/bake/src/Command/TestCommand.php on line 465
Stack Trace:
- /application/vendor/cakephp/bake/src/Command/TestCommand.php:502
- /application/vendor/cakephp/bake/src/Command/TestCommand.php:452
- /application/vendor/cakephp/bake/src/Command/TestCommand.php:258
- /application/vendor/cakephp/bake/src/Command/TestCommand.php:132
- /application/vendor/cakephp/cakephp/src/Console/BaseCommand.php:189
- /application/vendor/cakephp/cakephp/src/Console/CommandRunner.php:334
- /application/vendor/cakephp/cakephp/src/Console/CommandRunner.php:172
- /application/bin/cake.php:12

[TypeError] Bake\Command\TestCommand::_processModel(): Argument cakephp/cakephp#1 ($subject) must be of type Cake\ORM\Table, NbpApi\Model\Endpoint\NbpRatesEndpoint given, called in /application/vendor/cakephp/bake/src/Command/TestCommand.php on line 502 in /application/vendor/cakephp/bake/src/Command/TestCommand.php on line 465

Stack Trace:

#0 /application/vendor/cakephp/bake/src/Command/TestCommand.php(502): Bake\Command\TestCommand->_processModel(Object(NbpApi\Model\Endpoint\NbpRatesEndpoint))
cakephp/cakephp#1 /application/vendor/cakephp/bake/src/Command/TestCommand.php(452): Bake\Command\TestCommand->_processController(Object(SuperAdmin\Controller\Admin\NbpRatesController))
[...]

PR #952, opened 2023-10-10
#951 Support for non-table model types

asgraf:
Support for non-table model types


dereuromark:
The phpstan errors seem related

PR #953, opened 2023-10-10
Support for non-table model types (2.x)

asgraf:
#951

### DIFF
--- a/src/Command/TestCommand.php
+++ b/src/Command/TestCommand.php
@@ -24,6 +24,7 @@ use Cake\Controller\Controller;
 use Cake\Core\Configure;
 use Cake\Core\Exception\CakeException;
 use Cake\Core\Plugin;
+use Cake\Datasource\RepositoryInterface;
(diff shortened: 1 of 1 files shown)


src/Command/TestCommand.php
```diff
@@ -459,12 +460,15 @@ public function generateFixtureList($subject): array
     /**
      * Process a model, pull out model name + associations converted to fixture names.
      *
-     * @param \Cake\ORM\Table $subject A Model class to scan for associations and pull fixtures off of.
+     * @param @param \Cake\Datasource\RepositoryInterface $subject A Model class to scan for associations and pull fixtures off of.
```
dereuromark: Phpstan found the double param here invalid

markstory:
Would be good to have test coverage for generating non-ORM models if that's going to be a feature.

dereuromark:
ping @asgraf

dereuromark:
Ping @cakephp/core

PR #954, opened 2023-10-16
3.x: adjust annotations

### LINKED ISSUE #946: RFC: generics for array collections?
### Description

When running IdeHelper over the baked code, it refactors it as:
```diff
  * @method \App\Model\Entity\Apple newEntity(array $data, array $options = [])
- * @method \App\Model\Entity\Apple[] newEntities(array $data, array $options = [])
+ * @method array<\App\Model\Entity\Apple> newEntities(array $data, array $options = [])
  * @method \App\Model\Entity\Apple get(mixed $primaryKey, array|string $finder = 'all', \Psr\SimpleCache\CacheInterface|string|null $cache = null, \Closure|string|null $cacheKey = null, mixed ...$args)
  * @method \App\Model\Entity\Apple findOrCreate($search, ?callable $callback = null, $options = [])
  * @method \App\Model\Entity\Apple patchEntity(\Cake\Datasource\EntityInterface $entity, array $data, array $options = [])
- * @method \App\Model\Entity\Apple[] patchEntities(iterable $entities, array $data, array $options = [])
+ * @method array<\App\Model\Entity\Apple> patchEntities(iterable $entities, array $data, array $options = [])
 * @method \App\Model\Entity\Apple|false save(\Cake\Datasource\EntityInterface $entity, $options = [])
 * @method \App\Model\Entity\Apple saveOrFail(\Cake\Datasource\EntityInterface $entity, $options = [])
- * @method \App\Model\Entity\Apple[]|\Cake\Datasource\ResultSetInterface|false saveMany(iterable $entities, $options = [])
- * @method \App\Model\Entity\Apple[]|\Cake\Datasource\ResultSetInterface saveManyOrFail(iterable $entities, $options = [])
- * @method \App\Model\Entity\Apple[]|\Cake\Datasource\ResultSetInterface|false deleteMany(iterable $entities, $options = [])
- * @method \App\Model\Entity\Apple[]|\Cake\Datasource\ResultSetInterface deleteManyOrFail(iterable $entities, $options = [])
+ * @method \Cake\Datasource\ResultSetInterface<\App\Model\Entity\Apple>|false saveMany(iterable $entities, $options = [])
+ * @method \Cake\Datasource\ResultSetInterface<\App\Model\Entity\Apple> saveManyOrFail(iterable $entities, $options = [])
+ * @method \Cake\Datasource\ResultSetInterface<\App\Model\Entity\Apple>|false deleteMany(iterable $entities, $options = [])
+ * @method \Cake\Datasource\ResultSetInterface<\App\Model\Entity\Apple> deleteManyOrFail(iterable $entities, $options = [])
```

I wonder, should we switch over baked code in general to new generic style?
Or do we want to keep the legacy syntax of `Collection|Entity[]` etc?
Any pros/cons?

LordSimal:
closes https://github.com/cakephp/bake/issues/946

### DIFF
--- a/src/View/Helper/DocBlockHelper.php
+++ b/src/View/Helper/DocBlockHelper.php
@@ -265,17 +265,17 @@ class DocBlockHelper extends Helper
         // phpcs:disable
         $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity} newEmptyEntity()";
         $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity} newEntity(array \$data, array \$options = [])";
-        $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity}[] newEntities(array \$data, array \$options = [])";
+        $annotations[] = "@method array<\\{$namespace}\\Model\\Entity\\{$entity}> newEntities(array \$data, array \$options = [])";
         $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity} get(mixed \$primaryKey, array|string \$finder = 'all', \\Psr\\SimpleCache\\CacheInterface|string|null \$cache = null, \Closure|string|null \$cacheKey = null, mixed ...\$args)";
-        $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity} findOrCreate(\$search, ?callable \$callback = null, \$options = [])";
+        $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity} findOrCreate(\$search, ?callable \$callback = null, array \$options = [])";
         $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity} patchEntity(\\Cake\\Datasource\\EntityInterface \$entity, array \$data, array \$options = [])";
-        $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity}[] patchEntities(iterable \$entities, array \$data, array \$options = [])";
-        $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity}|false save(\\Cake\\Datasource\\EntityInterface \$entity, \$options = [])";
-        $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity} saveOrFail(\\Cake\\Datasource\\EntityInterface \$entity, \$options = [])";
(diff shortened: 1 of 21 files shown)


LordSimal:
Had to ignore that stan issue because as you can see in https://github.com/cakephp/bake/actions/runs/6530648862/job/17730339754?pr=954#step:7:67 this will get generated with the lowest chronos version due to the fact, that it doesn't extend Datetime in that specific version

src/View/Helper/DocBlockHelper.php
```diff
@@ -265,17 +265,17 @@ public function buildTableAnnotations(
...
+        $annotations[] = "@method array<\\{$namespace}\\Model\\Entity\\{$entity}> patchEntities(iterable \$entities, array \$data, array \$options = [])";
+        $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity}|false save(\\Cake\\Datasource\\EntityInterface \$entity, array \$options = [])";
+        $annotations[] = "@method \\{$namespace}\\Model\\Entity\\{$entity} saveOrFail(\\Cake\\Datasource\\EntityInterface \$entity, array \$options = [])";
+        $annotations[] = "@method \Cake\Datasource\ResultSetInterface<\\{$namespace}\\Model\\Entity\\{$entity}>|false saveMany(iterable \$entities, array \$options = [])";
+        $annotations[] = "@method \Cake\Datasource\ResultSetInterface<\\{$namespace}\\Model\\Entity\\{$entity}> saveManyOrFail(iterable \$entities, array \$options = [])";
+        $annotations[] = "@method \Cake\Datasource\ResultSetInterface<\\{$namespace}\\Model\\Entity\\{$entity}>|false deleteMany(iterable \$entities, array \$options = [])";
+        $annotations[] = "@method \Cake\Datasource\ResultSetInterface<\\{$namespace}\\Model\\Entity\\{$entity}> deleteManyOrFail(iterable \$entities, array \$options = [])";
```
ndm2, line 278: That doesn't look right to me. While those methods _could_ return `\Cake\Datasource\ResultSetInterface`, they generally return what they were fed with in the `$entities` argument, which would be `iterable<\Cake\Datasource\EntityInterface>`.

I think the point of additionally having `\Cake\Datasource\ResultSetInterface`, was to account for the use-case of passing (modified) query results into these methods.

dereuromark: Indeed, we could use `iterable<\Cake\Datasource\EntityInterface>|\Cake\Datasource\ResultSetInterface` here

PR #959, opened 2023-11-02
Fix annotation for entity field mapped to enum type.

### LINKED ISSUE #958: Baked enum docblock wrong
### Description

Following https://github.com/cakephp/cakephp/pull/17365

When you have a type mapped as
```php
$this->getSchema()->setColumnType('status', EnumType::from(UserStatus::class));
```

one would expect the baked entity to contain
```php
@property \Sandbox\Model\Enum\UserStatus $status
```
which it actually is when debugging it.

It currently is baked as
```
@property string $status
```
which doesnt work as it doesnt implement Stringable.
```php
echo h($user->status);
```
output is 
```
(object)Sandbox\Model\Enum\UserStatus
```

The string might be okish/correct if we check for Stringable being implemented, as then the label could be auto-printed. But even then we should stick to the object, as otherwise in the templates we cannot have autohinting on the available methods or properties, e.g.
```php
echo h($user->status->value);
```

Note:
In my case I used a backed enum as tinyint(1) in DB. But should probably be the same from the display/usage perspective.

cc @ADmad 

### Bake Version

3.x

### PHP Version

8.2

ADmad:
Closes #958

### DIFF
--- a/src/View/Helper/DocBlockHelper.php
+++ b/src/View/Helper/DocBlockHelper.php
@@ -5,6 +5,7 @@ namespace Bake\View\Helper;
 
 use Cake\Collection\Collection;
 use Cake\Core\App;
+use Cake\Database\Type\EnumType;
 use Cake\Database\TypeFactory;
 use Cake\ORM\Association;
 use Cake\Utility\Inflector;
@@ -230,6 +231,14 @@ class DocBlockHelper extends Helper
                 }
 
                 return '\Cake\I18n\Time';
+
+            default:
+                if (str_starts_with($type, 'enum-')) {
+                    $dbType = TypeFactory::build($type);
+                    if ($dbType instanceof EnumType) {
+                        return '\\' . $dbType->getEnumClassName();
+                    }
+                }
         }
 
         // Any unique or custom types will have a `string` type hint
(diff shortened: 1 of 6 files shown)


othercorey:
Looks good, but will let @LordSimal confirm the type is built properly.

LordSimal:
LGTM 👍🏻

src/View/Helper/DocBlockHelper.php
```diff
@@ -230,6 +231,14 @@ public function columnTypeToHintType(string $type): ?string
                 }
 
                 return '\Cake\I18n\Time';
+
+            default:
+                if (str_starts_with($type, 'enum-')) {
+                    $dbType = TypeFactory::build($type);
```
othercorey, line 237: @LordSimal Won't this throw an exception if the type isn't found? Is that ok for a catch-all?

dereuromark: Indeed, we might wanna use try catch

LordSimal: I am not really sure this situation is possible.
Getting to this case where the cake type string starts with `enum-` (basically) requires it being generated by
```
EnumType::from(MyEnum::class)
```
So that DBType needs to be present in the TypeFactory

dereuromark:
For IdeHelper I did https://github.com/dereuromark/cakephp-ide-helper/pull/333
Just waiting for a release here to constraint to it.

Issue #957, opened 2023-10-25
Application/Validation rule issues


ndm2:
### Description

I don't have time right now to take care of it, but I wanted to note it down so that I don't forget about it too easily, or maybe someone else can take a look.

# Fields incorrectly identified as file fields

When a field contains the strings `file` or `image`, as in for example `profile_id`, bake will generate a `*EmptyFile()` validation rule for it.

https://github.com/cakephp/bake/blob/d0dd499d199b6e0c0adfede2267b6ab8c0589a7d/src/Command/ModelCommand.php#L966-L968

Maybe it would be better to instead test for the end of the column name, and require a separator, something along the lines of this:

```
/(^|\s|_|-)(attachment|file|image)$/i
```

# Rules for unique constraints getting lost

When generating the map with rules, the indexes are based on column names, 

The code for generating `isUnique()` rules uses only the first column of the constraint for the index in the rules map, resulting in it being overwritten when multiple unique constraints exists where the first column is identical.

Furthermore the code for generating `existsIn()` rules can overwrite the unique rules in case that first column in the unique constraints is a foreign key column.

https://github.com/cakephp/bake/blob/d0dd499d199b6e0c0adfede2267b6ab8c0589a7d/src/Command/ModelCommand.php#L998-L1028

Relatively easy fix might be to not use explicit indexes, but just use the `fields` option, and figure things from that in the templates. Being able to specify multiple error fields for a single rule could be helpful here too.

### Bake Version

3.0.2

### PHP Version

5.0

ADmad:
> Maybe it would be better to instead test for the end of the column name, and require a separator,

I am sure there are people who name their fields like `image_profile` or just `image` / `file` (I have done the latter).

ndm2:
> > Maybe it would be better to instead test for the end of the column name, and require a separator,
> 
> I am sure there are people who name their fields like `image_profile` ...

I'd reckon that must be  the same people that read "_Don't Dead, Open Inside_" and walk right through the door, it's just natural selection at that point 😉 

>... or just `image` / `file` (I have done the latter).

That's perfectly reasonable, I have such fields too, the example regex there would've catched that.

ADmad:
> I'd reckon that must be the same people that read "Don't Dead, Open Inside" and walk right through the door, it's just natural selection at that point 😉

I am going to give humanity the benefit of doubt here and assume there are languages where `image_foo` feels more grammatically correct or maybe for RTL languages 🙂.

[...]

PR #967, opened 2023-12-22
3.next - Fix unique constraints getting lost.

ndm2:
When using the first column of the constraints and foreign keys to index the rules array, constraints and foreign keys that share the same first column in the composite constraint will overwrite previously set rules.

Now that is kind of a breaking change for people that are using custom table templates and/or custom model commands, but I don't really see a good way to fix this in a backwards compatible manner without awkward workarounds that will eventually still have pitfalls. Maybe this should rather go in a not yet existent 3.next?

Ideally there would be support for multiple error fields, as setting the error on the first field only is kinda awkward. Any opinions on that?

refs #957

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -984,22 +984,14 @@ class ModelCommand extends BakeCommand
             return [];
         }
         $schema = $model->getSchema();
-        $fields = $schema->columns();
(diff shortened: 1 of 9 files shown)


othercorey:
3.next does make sense for this and some other feature PRs

PR #968, opened 2023-12-22
Allow baking enums.

dereuromark:
> bin/cake bake enum FooBar

or 

> bin/cake bake enum FooBar -i

Next step could be to use conventions to auto map/detect enums for table fields:
- `{EntityName}{FieldName}` enum

E.g. `Users` table with field `status`: `User` entity and `UserStatus` enum would map and get added to forms etc.

--- 

the fails are unrelated and will be green once https://github.com/cakephp/cakephp/pull/17482 is released with 5.0.4


tests/comparisons/Model/testBakeEnum.php
```diff
@@ -0,0 +1,18 @@
+<?php
+declare(strict_types=1);
+
+namespace Bake\Test\App\Model\Enum;
+
+/**
+ * FooBar Enum
+ */
+enum FooBar: int
```
ADmad: `string` return type might be better here. Personally I prefer my raw db values to be readable too. We could default to`string ` and have an option to specify the return type if required.

tests/comparisons/Model/testBakeEnum.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    /**
+     * @return string
+     */
+    public function label(): string
+    {
+        return mb_strtolower($this->name);
```
ADmad: "Humanized" inflection would be better here as these are the values displayed to the user.

othercorey: This also implies a utf8 name. Is that something that exists?

tests/comparisons/Model/testBakeEnum.php
markstory: :+1: to defaulting to string enums. They are far simpler to operate over strings. You can always optimize the storage of stable enums by using concrete enum columns as well.

tests/comparisons/Model/testBakeEnum.php
markstory: Entertainingly the following is valid PHP

```php
<?php

enum Foo : string {
    case FIRST = 'first';
    case 😃 = 'unicode';
}
echo Foo::FIRST->value;
echo Foo::😃->value;
```

templates/bake/Model/enum.twig
```diff
@@ -0,0 +1,30 @@
...
+ * @since         3.1.0
+ * @license       https://www.opensource.org/licenses/mit-license.php MIT License
+ */
+#}
+{{ element('Bake.file_header', {
+    namespace: "#{namespace}\\Model\\Enum",
+}) }}
+
+{{ DocBlock.classDescription(name, 'Enum', [])|raw }}
+enum {{ name }}: int
```
markstory: Wasn't there an interface we needed enums to implement to have their label method used?

tests/comparisons/Model/testBakeEnum.php
dereuromark: Can we add a flag --backed ?
I prefer them for various reasons

othercorey: I think an option is needed here. `--backing string` makes the most sense

templates/bake/Model/enum.twig
LordSimal: `EnumLabelInterface` is what you are looking for

tests/comparisons/Model/testBakeEnum.php
LordSimal: I'd image people will be so confused and mix up `bake` and `backing`

templates/bake/Model/enum.twig
dereuromark: I added that myself
When whipping up my ideas yesterday I totally forget about that one^^

tests/comparisons/Model/testBakeEnum.php
dereuromark: --backed-enum ?
There will also be a description

ADmad: > I'd image people will be so confused and mix up bake and backing

You need to be really bad a English to mix up those.

> --backed-enum

The "enum" sounds redundant

How about `--backing-type`?

dereuromark:
Given the description I think `--backed=string|int` should suffice as option.
```
$ bin/cake bake enum FooBar -h
Bake (backed) enums for use in models.

Usage:
cake bake enum [options]

Options:

--backed, -b      If using backed enums. Set to `string` or `int`.
--connection, -c  The datasource connection to get data from.
```

dereuromark:
In a 2nd step we could think of allowing argument/option to dictate the enum values, similar to bake migrations etc:

Backed int
> 0: inactive, 1:active, ...

Backed string
> nope: inactive, 'yep yep':active, ...

Default
> inactive, active, ...

In a 3rd step: If those would be part of the db field comment, it could be used to auto-generate the whole enum based on that, making it super RAD.

src/Command/EnumCommand.php
```diff
@@ -51,4 +55,45 @@ public function template(): string
...
+     * @return \Cake\Console\ConsoleOptionParser
+     */
+    public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
+    {
+        $parser = $this->_setCommonOptions($parser);
+
+        $parser->setDescription(
+            'Bake (backed) enums for use in models.'
+        )->addOption('backed', [
+            'help' => 'If using backed enums. Set to `string` or `int`.',
```
ADmad: For the db at least you can only use backed enums. So we can skip the "If using...".

ADmad: Also since there are only 2 valid values we can use the `choices` config as done [here](https://github.com/cakephp/bake/blob/3.x/src/Command/TestCommand.php#L717) with default as string.

dereuromark: That only works for arguments, not options though

dereuromark: Rephrased the help sentence.

tests/comparisons/Model/testBakeEnumBacked.php
```diff
@@ -0,0 +1,21 @@
...
+ * FooBar Enum
+ */
+enum FooBar: string implements EnumLabelInterface
+{
+    /**
+     * @return string
+     */
+    public function label(): string
+    {
+        return Inflector::humanize(mb_strtolower($this->name));
```
ADmad, line 19:
```suggestion
        return Inflector::humanize(Inflector::underscore($this->name));
```

This will ensure name like "FooBar" gets inflicted to "Foo Bar".

templates/bake/Model/enum.twig
```diff
@@ -0,0 +1,34 @@
...
+
+{{ DocBlock.classDescription(name, 'Enum', [])|raw }}
+enum {{ name }}{{ backed ? ': ' ~ backed : ''}} implements EnumLabelInterface
+{
+    /**
+     * @return string
+     */
+    public function label(): string
+    {
+        return Inflector::humanize(mb_strtolower($this->name));
```
ADmad:
```suggestion
        return Inflector::humanize(Inflector::underscore($this->name));
```

This will ensure name like "FooBar" gets inflected to "Foo Bar".

src/Command/EnumCommand.php
ADmad: >  That only works for arguments, not options though

Options also have choices https://github.com/cakephp/cakephp/blob/5.x/src/Console/ConsoleInputOption.php#L101

dereuromark: If anyone has time  feel free to add on too
I am now offline for a bit.

src/Command/EnumCommand.php
```diff
@@ -0,0 +1,99 @@
...
+     * @param \Cake\Console\ConsoleOptionParser $parser The option parser to update.
+     * @return \Cake\Console\ConsoleOptionParser
+     */
+    public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
+    {
+        $parser = $this->_setCommonOptions($parser);
+
+        $parser->setDescription(
+            'Bake (backed) enums for use in models.'
+        )->addOption('backed', [
```
othercorey: Think this should be `backing` not` backed` since it will always be backed. The `backing` type is `string`.

At least, that's what makes sense to me in english. Maybe it's harder for non-english speakers to understand the difference though.

ADmad: I agree, hence I suggested `backing-type` earlier or shorter `backing` works too.

othercorey: `backing-type` is clearer.

dereuromark: Wait, but without that Option the enum is not backed. So for me the original Name still seems to make sense. 
I am fine changing it to your suggestion if u think it makes more sense.

ADmad: > Wait, but without that Option the enum is not backed

Why do we need to bake non-backed enums? The db layer only supports backed enums, so it's only a question of deciding whether to have `int` return type or `string`, defaulting to the latter.

dereuromark: True, the db layer sure does
But maybe people want to use enums in their (model) layer or otherwise and those wouldnt necessarily need to be backed.
So I thought:
- normal (default)
- backing-type int or string

ADmad: Then I don't see why such enums would need a `label()` method too and without that such classes can be easily created by hand.

I have opened https://github.com/cakephp/bake/pull/970 which my suggested changes on top of this PR.

ADmad: Making `string` backed enums would be the most common use case and having to specify `-b string` each time would be an unnecessary chore.

dereuromark:
Looks like we can bake enums now for xmas:)

dereuromark:
Given that we only now have backed, and string vs int
Couldnt we just add boolean `--int` flag? `-b ...` seems not necessary in that case.

ADmad:
`--int` has just 1 character less than `-b int` :)

I would prefer if we keep the more descriptive name.

dereuromark:
Should it be just `type`/`t` though? Since it is clear it is a backing type on backed enums :)

ADmad:
I can live with `type` if others are in favor.

dereuromark:
Never mind, t is theme :)

That said: Shouldnt it throw an error if you define a short twice?
```
--theme, -t       The theme to use when baking code. (choices:
                  Bake|BootstrapUI|Migrations|Queue|Search|Setup|StateMachine)
--type, -t        The return type for the backed enum class
```
No error here..

ADmad:
> Shouldnt it throw an error if you define a short twice?

Guess we are missing duplicate naming checks.

ADmad (approved):

dereuromark:
@ADmad it is now simpler, only one char for int

So all good?

PR #969, opened 2023-12-23
Bake table with enums mapped.

dereuromark:
Follow https://github.com/cakephp/bake/pull/968

Currently finds the enums per fields, e.g.
```
[
  'status' => 'Bake\Test\App\Model\Enum\BakeArticleStatus'
]
```

This already works, but needs some cleanup
How do we avoid the fqcn for example? Use statements should be used instead.


othercorey:
So any field that shares the name of an enum is automatically converted? Not sure how I feel about that.

tests/comparisons/Model/testBakeTableWithEnum.php
```diff
@@ -0,0 +1,116 @@
+<?php
+declare(strict_types=1);
+
+namespace Bake\Test\App\Model\Table;
+
+use Bake\Test\App\Model\Enum\BakeArticleStatus;
```
markstory: I like this convention. It helps ensure that enums are scoped to the entity so that we don't easily get conflicts.

src/Command/ModelCommand.php
```diff
@@ -1382,4 +1383,28 @@ public function bakeTest(string $className, Arguments $args, ConsoleIo $io): voi
...
+     * @param string $namespace
+     * @return array<string, class-string>
+     */
+    protected function enums(array $data, string $entity, string $namespace): array
+    {
+        $fields = $data['fields'];
+        $enumClassNamespace = $namespace . '\Model\Enum\\';
+
+        $enums = [];
+        foreach ($fields as $field) {
```
ADmad, line 1399: Don't we have a way to iterate through only sting and integer fields?

dereuromark: I found a way, see enumLikeFields()

But should be also allow integer, or only string and tinyinteger? I feel like someone with integer doesnt really want enums^^

src/Command/ModelCommand.php
```diff
@@ -1382,4 +1383,48 @@ public function bakeTest(string $className, Arguments $args, ConsoleIo $io): voi
...
+        }
+
+        return $enums;
+    }
+
+    /**
+     * @param \Cake\Database\Schema\TableSchemaInterface $schema
+     * @return array<string>
+     */
+    protected function enumLikeFields(TableSchemaInterface $schema): array
```
dereuromark: Could also be named `possibleEnumFields`

src/Command/ModelCommand.php
ADmad: Being pedantic about the db column type will just increase support questions :) I am sure lot of people don't bother and just use integer field.

dereuromark: Fair enough

dereuromark:
~~Any idea what to do with the postgres topic?~~
Found it, apparently there it is 'smallinteger'...

Also, how can we cleanup the FQCN here inside the code?

PR #971, opened 2023-12-25
Make backing type a bool, since it defaults anyway on 1 of 2 options.

dereuromark:
https://github.com/cakephp/bake/pull/968

I would still like to at least propose this for consideration.
Given that there are now only 2 - and string by default, it seems kind of overkill to have a full option for it.
It seems `-i` could suffice?

### DIFF
--- a/src/Command/EnumCommand.php
+++ b/src/Command/EnumCommand.php
@@ -18,7 +18,6 @@ namespace Bake\Command;
 
 use Cake\Console\Arguments;
 use Cake\Console\ConsoleOptionParser;
-use InvalidArgumentException;
 
 /**
  * Enum code generator.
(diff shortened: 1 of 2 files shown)


othercorey:
Sorry, but I think this is a confusing option. Maybe if it was `--int` or `--string` that would be clear.

dereuromark:
So just --int ?

src/Command/EnumCommand.php
```diff
@@ -83,11 +82,10 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
...
             'Bake backed enums for use in models.'
-        )->addOption('backing-type', [
-            'help' => 'The return type for the backed enum class',
-            'default' => 'string',
-            'choices' => ['string', 'int'],
-            'short' => 'b',
+        )->addOption('int', [
+            'help' => 'Using backed enums with int instead of string as return type',
+            'boolean' => true,
+            'short' => 'i',
```
markstory, line 88: I don't know if we need a short option for this as `int` isn't that long.

dereuromark: Well, it can still be convinient :)

markstory: Fair enough.

PR #972, opened 2023-12-28
Complete baking of enums.

dereuromark:
Completes https://github.com/cakephp/bake/pull/968 and https://github.com/cakephp/bake/pull/969

Now fully supports baking enums including cases.

### Examples
implicit cases for string:
> bin/cake bake enum UserStatus inactive,active

explicit cases for string:
> bin/cake bake enum UserStatus inactive:i,active:a

implicit cases for int:
> bin/cake bake enum UserStatus inactive,active -i

explicit cases for int:
> bin/cake bake enum UserStatus active:1,inactive:0

On top  of that now also bake all works completely out of the box when you set a `[enum]` marker + optional case config as comment:
```php
            ->addColumn('status', 'tinyinteger', [
                'default' => 0,
                'limit' => 2,
                'null' => false,
                'comment' => '[enum] inactive:0,active:1'
            ])
            ->addColumn('gender', 'string', [
                'default' => null,
                'limit' => 10,
                'null' => true,
                'comment' => '[enum] male,female,diverse'
            ])
```

> bin cake bake all Users

will generate
- both enums incl the cases if given
- map those to the baked table

```php
enum UserStatus: int implements EnumLabelInterface
{
    case INACTIVE = 0;
    case ACTIVE = 1;
    ...
    
enum UserGender: string implements EnumLabelInterface
{
    case MALE = 'male';
    case FEMALE = 'female';
    case DIVERSE = 'diverse';
    ...
```

```php
$this->getSchema()->setColumnType('status', \Cake\Database\Type\EnumType::from(\App\Model\Enum\UserStatus::class));
$this->getSchema()->setColumnType('gender', \Cake\Database\Type\EnumType::from(\App\Model\Enum\UserGender::class));
```

If this sounds fine as is, I will add a few more test cases.
Just wanted to get some feedback on it first.

Also, open tasks:
- [x] Bake entity correctly in terms of annotations (currently still int and string every 2nd time for some reason)
- [x] Bake index and view correctly, as those also are using format() and h() instead of `->label()`.

### DIFF
--- a/src/Command/EnumCommand.php
+++ b/src/Command/EnumCommand.php
@@ -18,6 +18,8 @@ namespace Bake\Command;
 
 use Cake\Console\Arguments;
 use Cake\Console\ConsoleOptionParser;
+use Cake\Utility\Inflector;
+use InvalidArgumentException;
 
 /**
  * Enum code generator.
@@ -64,8 +66,20 @@ class EnumCommand extends SimpleBakeCommand
      */
     public function templateData(Arguments $arguments): array
     {
+        $cases = $this->parseCases($arguments->getArgument('cases'), (bool)$arguments->getOption('int'));
+        $isOfTypeInt = $this->isOfTypeInt($cases);
+        $backingType = $isOfTypeInt ? 'int' : 'string';
+        if ($arguments->getOption('int')) {
+            if ($cases && !$isOfTypeInt) {
+                throw new InvalidArgumentException('The cases provided do not seem to match the int type you want to bake');
+            }
+
+            $backingType = 'int';
+        }
+
         $data = parent::templateData($arguments);
-        $data['backingType'] = $arguments->getOption('int') ? 'int' : 'string';
+        $data['backingType'] = $backingType;
+        $data['cases'] = $this->formatCases($cases);
 
         return $data;
     }
@@ -82,12 +96,82 @@ class EnumCommand extends SimpleBakeCommand
 
         $parser->setDescription(
             'Bake backed enums for use in models.'
-        )->addOption('int', [
-            'help' => 'Using backed enums with int instead of string as return type',
+        )->addArgument('name', [
+            'help' => 'Name of the enum to bake. You can use Plugin.name to bake plugin enums.',
+            'required' => true,
+        ])->addArgument('cases', [
+            'help' => 'List of either `one,two` for string or `0:foo,1:bar` for int type.',
+        ])->addOption('int', [
+            'help' => 'Using backed enums with int instead of string as return type.',
             'boolean' => true,
             'short' => 'i',
         ]);
 
         return $parser;
+    }
+
+    /**
+     * @param string|null $casesString
+     * @return array<int|string, string>
+     */
+    protected function parseCases(?string $casesString, bool $int): array
+    {
+        if ($casesString === null) {
+            return [];
+        }
+
+        $enumCases = explode(',', $casesString);
+
+        $definition = [];
+        foreach ($enumCases as $k => $enumCase) {
+            $key = $value = trim($enumCase);
+            if (str_contains($key, ':')) {
+                $value = trim(mb_substr($key, strpos($key, ':') + 1));
+                $key = mb_substr($key, 0, strpos($key, ':'));
+            } elseif ($int) {
+                $key = $k;
+            }
+
+            $definition[$key] = $value;
+        }
+
+        return $definition;
+    }
+
+    /**
+     * @param array<int|string, string> $definition
+     * @return bool
+     */
+    protected function isOfTypeInt(array $definition): bool
+    {
+        if (!$definition) {
+            return false;
+        }
+
+        foreach ($definition as $key => $value) {
+            if (!is_int($key)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     * @param array<int|string, string> $cases
+     * @return array<string>
+     */
+    protected function formatCases(array $cases): array
+    {
+        $formatted = [];
+        foreach ($cases as $case => $alias) {
+            $alias = mb_strtoupper(Inflector::underscore($alias));
+            if (is_string($case)) {
+                $case = '\'' . $case . '\'';
+            }
+            $formatted[] = 'case ' . $alias . ' = ' . $case . ';';
+        }
+
+        return $formatted;
     }
 }
--- a/src/Command/EnumCommand.php
+++ b/src/Command/EnumCommand.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 namespace Bake\Command;
 
 use Cake\Console\Arguments;
+use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Utility\Inflector;
 use InvalidArgumentException;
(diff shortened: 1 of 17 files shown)


src/Command/EnumCommand.php
```diff
@@ -64,8 +66,20 @@ public function template(): string
      */
     public function templateData(Arguments $arguments): array
     {
+        $cases = $this->parseCases($arguments->getArgument('cases'), (bool)$arguments->getOption('int'));
+        $isOfTypeInt = $this->isOfTypeInt($cases);
+        $backingType = $isOfTypeInt ? 'int' : 'string';
+        if ($arguments->getOption('int')) {
+            if ($cases && !$isOfTypeInt) {
+                throw new InvalidArgumentException('The cases provided do not seem to match the int type you want to bake');
```
othercorey:
```suggestion
                throw new InvalidArgumentException('Cases do not match requested `int` backing type.');
```

ADmad:
> bin/cake bake enum UserStatus i:inactive,a:active

IMO it should be `bin/cake bake enum UserStatus inactive:i,active:a` as that's how named args work in PHP, so  `name`:`value` feels more natural.

ADmad:
Also instead of using uppercase case names I would use capitalized case names.

templates/bake/Model/enum.twig
```diff
@@ -24,11 +24,15 @@
...
+    {{ Bake.concat('\n    ', cases) }}
+
+{% endif %}
     /**
      * @return string
      */
     public function label(): string
     {
-        return Inflector::humanize(Inflector::underscore($this->name));
+        return Inflector::humanize(mb_strtolower($this->name));
```
ADmad: Why this change again?

dereuromark:
> Also instead of using uppercase case names I would use capitalized case names.

Arent those kind of constants?
I orientied myself on core files here actually, according to blame also the ones you added:
https://github.com/cakephp/cakephp/blame/5.x/src/Http/Cookie/SameSiteEnum.php#L20-L22
Same for
- DriverFeatureEnum
- as well as the test enums in core

So I guessed that we are now using UPPER_CASE instead of CamelCase here as convention.

Whatever we decide on, we probably want to have a code sniffer rule for it.

templates/bake/Model/enum.twig
dereuromark: Well, because it doesnt work otherwise with UPPER_CASE, as it would become U P P E R C A S E for display field in form. See my other comment and referenced core code.

dereuromark:
> IMO it should be bin/cake bake enum UserStatus inactive:i,active:a as that's how named args work in PHP, so name:value feels more natural.

I was more thinking of how we all know arrays work with `"key/case" => "value/alias"` transferring here into those.
But in the end I am also fine with another interpretation of it.

src/Command/EnumCommand.php
```diff
@@ -82,12 +97,104 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
...
+
+    /**
+     * @param array<int|string, string> $cases
+     * @return array<string>
+     */
+    protected function formatCases(array $cases): array
+    {
+        $formatted = [];
+        foreach ($cases as $case => $alias) {
+            $alias = mb_strtoupper(Inflector::underscore($alias));
```
markstory: Do we need to underscore? Developers could input their choices as `snake_case` or `UPPER_CASE`. Do we also need to support `mixedCase`?

src/Command/EnumCommand.php
```diff
@@ -82,12 +97,104 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
...
+            if (is_string($case)) {
+                $case = '\'' . $case . '\'';
```
markstory: Do we need to escape any `'` or raise an error should the `case` contain quotes or other characters not allowed in PHP constant names?

src/Command/ModelCommand.php
```diff
@@ -1444,4 +1454,94 @@ protected function possibleEnumFields(TableSchemaInterface $schema): array
 
         return $fields;
     }
+
+    /**
+     * @param \Cake\Database\Schema\TableSchemaInterface $schema
+     * @return array<string, mixed>
+     */
+    protected function getEnumDefinitions(TableSchemaInterface $schema): array
```
markstory, line 1466: Should we make the schema parsing logic in `EnumCommand` public methods so that we can re-use it here? Having two implementations of it could diverge.

ADmad:
> I orientied myself on core files here actually, according to blame also the ones you added

I have always preferred CamelCased names but probably went with the consensus opinion :). Since the couple of core enums use upper case I guess that's what we have settled on.

othercorey:
> > I orientied myself on core files here actually, according to blame also the ones you added
> 
> I have always preferred CamelCased names but probably went with the consensus opinion :). Since the couple of core enums use upper case I guess that's what we have settled on.

UPPERCASE enum cases usually come from other languages where convention is to use underscores to separate words in the case:

`MY_CASE` while php would typically use `MyCase` with no underscore.

dereuromark:
The current understanding is mainly that those are close to constants. Thus the uppercase.
We could still change the convention maybe.

othercorey:
> The current understanding is mainly that those are close to constants. Thus the uppercase. We could still change the convention maybe.

I'm happy to treat enums as constants if that is the convention.

markstory:
My personal preference would be to use CamelCase for enum values. But I also recognize that uppercase is more aligned with conventions for constants.

Are there other projects we should be looking at to share conventions with?

dereuromark:
Good question
The official docs list them as camel: https://www.php.net/manual/en/language.enumerations.backed.php

But then again, it seems a bit harder to read what is the enum (class), what is the case and what is the property..

src/Command/EnumCommand.php
dereuromark: According to PHP constants are

    ^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$

but is the same true for enum cases though?

It would still be wise to limit our bake to those IMO, as the other characters seem pointless here to support.

ADmad:
Enums can have constants. Can't think of a use case for them but if someone did have them then they wouldn't be able to easily distinguish between the constants and case.

ADmad:
> Constants in Enumerations MAY use either PascalCase or UPPER_CASE capitalization. PascalCase is RECOMMENDED, so that it is consistent with case declarations.

https://github.com/php-fig/per-coding-style/blob/master/spec.md#9-enumerations

ADmad:
> Use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case) for enumeration cases (e.g. InputArgumentMode::IsArray);

https://symfony.com/doc/current/contributing/code/standards.html

dereuromark:
Hmm then would recommend switching core first
And then add a code sniffer
In the meantime I can address in this PR:

- CamelBacked
- key value switching as requested

templates/bake/Model/enum.twig
dereuromark: reverted now

src/Command/EnumCommand.php
dereuromark: All cases should be possible now, as only the case is touched, not the value

src/Command/ModelCommand.php
dereuromark: I moved it into own utility method (static) for reuse

othercorey:
> Enums can have constants. Can't think of a use case for them but if someone did have them then they wouldn't be able to easily distinguish between the constants and case.
> > Constants in Enumerations MAY use either PascalCase or UPPER_CASE capitalization. PascalCase is RECOMMENDED, so that it is consistent with case declarations.

Funny they don't try to distinguish them either.

src/Command/ModelCommand.php
```diff
@@ -1444,4 +1455,95 @@ protected function possibleEnumFields(TableSchemaInterface $schema): array
...
+            if (!in_array($columnSchema['type'], ['string', 'integer', 'tinyinteger', 'smallinteger'], true)) {
+                continue;
+            }
+
+            if (empty($columnSchema['comment']) || strpos($columnSchema['comment'], '[enum]') === false) {
+                continue;
+            }
+
+            $enumsDefinitionString = trim(mb_substr($columnSchema['comment'], strpos($columnSchema['comment'], '[enum]') + 6));
+            $isInt = in_array($columnSchema['type'], ['integer', 'tinyinteger', 'smallinteger', true]);
```
```suggestion
            $isInt = in_array($columnSchema['type'], ['integer', 'tinyinteger', 'smallinteger'], true);
```

dereuromark:
I also tested in in real life with int + string + nullable each.
Seems to work all fine, incl rebake

ADmad (approved):

PR #973, opened 2024-01-04
Update phpunit.xml.dist.twig

dereuromark:


markstory:
Seems ok, tests need to be adjusted though.

PR #982, opened 2024-03-07
Fix up bake value defaults for fixtures with enums

### LINKED ISSUE #981: Enum bake fixture bug
### Description

When baking a table with enum, everything works out automatically, except for the fixture

                'status' => '',

leads to
```
TypeError: App\Model\Enum\OfferStatus::tryFrom(): Argument #1 ($value) must be of type int, string given
```

It is a int backed enum, so the default should either be null or numeric

The field itself is not null (required input).
I checked and it is the same issue for both default value present or default null.

Expected behavior:
- default null: null
- default some int: that int

Looking into bake schema collection, I can see those being passed into _generateRecords():
```
[
  'type' => 'enum-app-model-enum-offeraccuracy',
  'length' => null,
  'unsigned' => true,
  'null' => false,
  'default' => null,
  'comment' => '[enum] exact:0, approximate:1, month:2',
  'precision' => null,
  'autoIncrement' => null
]

[
  'type' => 'enum-app-model-enum-offerstatus',
  'length' => null,
  'unsigned' => true,
  'null' => false,
  'default' => '1',
  'comment' => '[enum] inactive:0, active:1, public:2',
  'precision' => null,
  'autoIncrement' => null
]
```
which looks correct, if we ignore the string vs int part on the default value.

But apparently, they are not being used.

### Bake Version

5.x

### PHP Version

8.2

dereuromark:
Resolves https://github.com/cakephp/bake/issues/981

both for required fields and default values present

### DIFF
--- a/src/Command/FixtureCommand.php
+++ b/src/Command/FixtureCommand.php
@@ -26,10 +26,13 @@ use Cake\Console\ConsoleOptionParser;
 use Cake\Core\Configure;
 use Cake\Core\Exception\CakeException;
 use Cake\Database\Schema\TableSchemaInterface;
+use Cake\Database\Type\EnumType;
+use Cake\Database\TypeFactory;
 use Cake\Datasource\ConnectionManager;
 use Cake\Utility\Inflector;
 use Cake\Utility\Text;
 use DateTimeInterface;
+use ReflectionEnum;
 
 /**
  * Task class for creating and updating fixtures files.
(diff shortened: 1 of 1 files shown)


src/Command/FixtureCommand.php
```diff
@@ -419,6 +422,31 @@ protected function _generateRecords(TableSchemaInterface $table, int $recordCoun
...
+                            if ($fieldInfo['default'] !== null) {
+                                $insert = $fieldInfo['default'];
+                                if ($backingType === 'int') {
+                                    $insert = (int)$insert;
+                                }
+                            } else {
+                                $cases = $reflectionEnum->getCases();
+                                $firstCase = array_shift($cases);
+                                /** @var \BackedEnum $firstValue */
+                                $firstValue = $firstCase->getValue();
```
markstory: Could we hit a runtime error here if the enum has no cases? I don't think there is a practical scenario for no cases, but it could happen.

dereuromark: Wrapped it.

PR #983, opened 2024-04-18
Wrap addPlugin() call in has() check.

ADmad:
Cake 5.1 throws exception for duplicate plugin loading and this avoids exception if the TwigView plugin is already loaded from the application.


ADmad:
I have created a `3.next` branch but this could safely go in `3.x` too.

dereuromark (approved):

Issue #991, opened 2024-06-07
Bake should support core plugins better


markstory:
### Description

Bake could be expanded to include behavior for other core plugins. We could enable this behavior in a generic way so that users could use it in any custom templates they may have. 

## Controller base - actions with plugin logic

When you're using the `authorization` plugin, all of your controller actions will require a call to `authorize()`. That is a perfect task for bake to take care of. For example, controller actions could have a block like:

```
{% if Bake.hasPlugin('Authorization') %}
    $this->Authorize->authorize(${{ singularVar }});
{% endif %}
```
This would allow plugins to have deeper integration with bake and get people further faster.

We can also use this to generate a simple login form, and `beforeFilter` method on a `UsersController` if the Authentication plugin is present. While I would like to be able to modify the `Application` methods as well, we would need bake to be able to parse and manipulate files. This feels similar to what rector does, but I'm not sure rector was intended to be used this way.

## Model bake - Add plugin behaviors

We could also explore giving plugins ways to hook into table generation. For example plugins could receive the table, inspect schema and add additional behaviors. An example usecase would be a Sluggable behavior being added to models with `slug` in their schema.

ADmad:
> When you're using the authorization plugin, all of your controller actions will require a call to authorize()

I updated my bake templates to do just that recently :)

> We could also explore giving plugins ways to hook into table generation.

That would be great, then I wouldn't have to override methods of the ModelCommand as I do now.

ADmad:
I don't think generating the `<plugin>/tests/bootstrap.php` and `<plugin>/tests/schema.sql` files makes sense either for in-app plugins.

dereuromark:
With https://github.com/cakephp/bake/pull/1007 merged, what else might still be needed?

PR #992, opened 2024-06-17
Add a method to BakeHelper to check for other plugins.

markstory:
First step to adding authentication and authorization hooks to bake templates.

Part of #991


othercorey (approved):
Check looks ok

PR #993, opened 2024-06-19
Detect Authorization plugin and include it in bake output

markstory:
If the host application has Authorization plugin installed, we can autodetect that and include component checks in generated code from bake.


.editorconfig
```diff
@@ -16,6 +16,3 @@ indent_size = 2
 
 [phars.xml]
 indent_size = 2
-
-[*.twig]
-insert_final_newline = false
```
othercorey: There was some file that needed this, but I can't remember which one.

templates/bake/element/Controller/index.twig
```diff
@@ -26,6 +26,9 @@
             ->contain({{ Bake.exportArray(belongsTo)|raw }});
 {% else %}
         $query = $this->{{ currentModelName }}->find();
+{% endif %}
+{% if Bake.hasPlugin('Authorization') %}
+        $query = $this->Authorization->applyScope($query);
```
othercorey, line 31: Do we need to bake stub code for these?

markstory: I think it is helpful. I usually use this to build in permissions like 'view all the projects I have membership in'.

othercorey: I mean, should we bake/generate the authorization code stubs to match this?

markstory: > I mean, should we bake/generate the authorization code stubs to match this?

That can be done already with `bake policy --type table Articles`. If we wanted to get fancy we could make the bake code generation respect whether or not you have a table/entity policy generated.

templates/bake/element/Controller/add.twig
```diff
@@ -22,6 +22,9 @@
     public function add()
     {
         ${{ singularName }} = $this->{{ currentModelName }}->newEmptyEntity();
+{% if Bake.hasPlugin('Authorization') %}
```
ADmad, line 25: I am curious about the use case for authorizing using an empty entity. Personally I have only authorized the "add" action.

markstory: Applications could have roles or rules on who can create what. I agree that most of the time add() will have a policy of `return true`.

.editorconfig
ADmad: Yeah whitespace changes in templates can affect the display, so I wouldn't auto add/remove it.

markstory:
Anything else I should take care of with this? I'd like to get this merged so I can move forwards with improving the authentication integration.

.editorconfig
ADmad: @markstory Can you revert this please.

ADmad (approved):
LGTM except the change to `.editorconfig`.

PR #994, opened 2024-06-20
Update brick/var-exporter to 0.5.0 and fix deprecations.

nicosp:
I took the liberty to update the PHP version for the code generation to 8.1 because it's also the minimum version for CakePHP 5.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "require": {
         "php": ">=8.1",
-        "brick/varexporter": "^0.4.0",
+        "brick/varexporter": "^0.5.0",
         "cakephp/cakephp": "^5.0.3",
         "cakephp/twig-view": "^2.0.0",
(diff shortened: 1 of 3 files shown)


dereuromark:
I does block apps from updating that still require/use v4 (even as non primary dependency)
But I think blocking isn't as bad as BC breaking and as such this change seems like a pragmatic way of moving the dependency (usage) forward.
:+1:

ADmad:
Only potential issue I see is someone using the removed VarExporter constants in their custom bake code. We can target this to `3.next` to be safe.

ADmad:
@nicosp Please re-target this to `3.next` branch.

dereuromark:
Also

    brick/varexporter         0.5.0      requires         nikic/php-parser (^5.0) 
    
Currently 4.x afaik

So better to have that clearly released as such :)

PR #996, opened 2024-07-05
Make adding twig non-blocking

markstory:
Bake is mostly interested that twig is set, and since #17628 addPlugin() will throw on duplicates. This also fixes the failing test with migrations + 5.1


ADmad:
This change will conflict the change in `3.next` https://github.com/cakephp/bake/pull/983

markstory:
Closing as this has been addressed in 3.next already.

PR #997, opened 2024-07-10
Improve bake for Users::login action

markstory:
If the application is using Authentication we can generate a more complete Users::login action that works with the authentication plugin.

### DIFF
--- a/templates/bake/element/Controller/login.twig
+++ b/templates/bake/element/Controller/login.twig
@@ -1,0 +1,42 @@
+{#
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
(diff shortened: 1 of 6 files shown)


templates/bake/element/Controller/login.twig
```diff
@@ -0,0 +1,42 @@
...
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.2.0
+ * @license       https://www.opensource.org/licenses/mit-license.php MIT License
+ */
+#}
+    /**
```
othercorey, line 16: Is this supposed to be indented?

PR #999, opened 2024-08-08
Bump up CakePHP and PHPUnit.

ADmad:
I couldn't figure out the proper attributes to use for the `@covers` annotations. So would be nice if someone could address that else we can drop the annotations.

### DIFF
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -111,7 +111,8 @@ class AllCommand extends BakeCommand
             }
 
             foreach ($tables as $table) {
-                $subArgs = new Arguments([$table], $options, ['name']);
+                $parser = $command->getOptionParser();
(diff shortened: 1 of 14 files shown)


src/Command/AllCommand.php
```diff
@@ -111,7 +111,8 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
             }
 
             foreach ($tables as $table) {
-                $subArgs = new Arguments([$table], $options, ['name']);
+                $parser = $command->getOptionParser();
+                $subArgs = new Arguments([$table], $options, $parser->argumentNames());
```
ADmad, line 115: This change is necessary to the avoid the `..argument not defined..` exception added in Cake 5.1. We need to ensure the command is executed with all it's defined args and not just `name`.

PR #1000, opened 2024-08-08
update stan

LordSimal:


phpstan.neon
```diff
@@ -3,8 +3,9 @@ includes:
 
 parameters:
     level: 6
```
dereuromark, line 5: We could try to reach 8 at some point maybe

LordSimal: Adjusting it to level 7 results in 37 errors, most of them in our commands and codegen parts.

PR #1003, opened 2024-09-13
merge 2.x => 3.x

LordSimal:


tests/TestCase/Command/TemplateCommandTest.php
```diff
@@ -443,6 +443,11 @@ public function testGetContentWithRoutingPrefix()
      */
     public function testBakeView()
     {
+        if (version_compare(Configure::version(), '4.5.0', '<')) {
```
ADmad: You'll have to removed this check after merging as Bake 3 isn't going to be used for 4.x :)

PR #1006, opened 2024-10-06
change command template to return actual exit code

LordSimal:
Refs: https://github.com/cakephp/bake/pull/950

### DIFF
--- a/templates/bake/Command/command.twig
+++ b/templates/bake/Command/command.twig
@@ -47,9 +47,10 @@ class {{ name }}Command extends Command
      *
      * @param \Cake\Console\Arguments $args The command arguments.
      * @param \Cake\Console\ConsoleIo $io The console io
-     * @return int|null|void The exit code or null for success
+     * @return int The exit code
      */
-    public function execute(Arguments $args, ConsoleIo $io)
+    public function execute(Arguments $args, ConsoleIo $io): int
     {
+        return Command::CODE_SUCCESS;
     }
 }
--- a/tests/comparisons/Command/testBakePlugin.php
+++ b/tests/comparisons/Command/testBakePlugin.php
@@ -32,9 +32,10 @@ class ExampleCommand extends Command
      *
      * @param \Cake\Console\Arguments $args The command arguments.
      * @param \Cake\Console\ConsoleIo $io The console io
-     * @return int|null|void The exit code or null for success
+     * @return int The exit code
      */
-    public function execute(Arguments $args, ConsoleIo $io)
+    public function execute(Arguments $args, ConsoleIo $io): int
     {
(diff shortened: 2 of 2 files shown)


dereuromark:
I don't like how this is now not in sync with controllers.
Both communication layer classes.
Just to please a wrongly configured cs tool.
Many of us don't have that issue.

It also doesn't reflect reality IMO

templates/bake/Command/command.twig
```diff
@@ -47,9 +47,10 @@ class {{ name }}Command extends Command
      *
      * @param \Cake\Console\Arguments $args The command arguments.
      * @param \Cake\Console\ConsoleIo $io The console io
-     * @return int|null|void The exit code or null for success
+     * @return int The exit code
      */
-    public function execute(Arguments $args, ConsoleIo $io)
+    public function execute(Arguments $args, ConsoleIo $io): int
```
dereuromark, line 52:
```suggestion
    public function execute(Arguments $args, ConsoleIo $io): ?int
```
Wouldn't this be more correct?

LordSimal:
> It also doesn't reflect reality IMO

[We return success on null as a fallback](https://github.com/cakephp/cakephp/blob/5.x/src/Console/CommandRunner.php#L168-L177), sure, but this doesn't mean this doesn't reflect reality.

I just don't like having 3 return types for a command which should always return an exit code since its a CLI tool.

templates/bake/Command/command.twig
ADmad: I would rather we keep it `int` only so that in a future release we can add `int` type to our base class `Command::execute()` itself.

ADmad (approved):

templates/bake/Command/command.twig
dereuromark: Wouldnt this start with core though
Baked Code by no means would be a reliable way of going into that direction

LordSimal: This kind of change can only be enforced in core in 6.x anyway.
With this change in bake it at least encourages new commands being generated to use this kind of setup.

markstory: Is requiring an explicit return value for commands going to make userland code better? I can see how it makes the framework code simpler, but does it also benefit end users?

dereuromark: I very much doubt that.
That said: Symfony went into that direction and the new versions require a non void (maybe even integer) return code.

My point here is that bake is the end of the chain - and narrowing it here to just int seems not like the best idea if there was no broader discussion on core changes yet.

LordSimal:
since there is not enough common agreement, this will not change in current Cake5. 

maybe something to consider for Cake6

Issue #1002, opened 2024-08-30
In PHPStan 1.12.0 with bleeding edge generated annotations @method for tables are reported


Harfusha:
### Description

https://github.com/dereuromark/cakephp-ide-helper/issues/353

`\src\Model\Table\AccountsTable.php:55:Class App\Model\Table\AccountsTable has PHPDoc tag @method for method newEntity() parameter #1 $data with no value type specified in iterable type array.`
`\src\Model\Table\AccountsTable.php:55:Class App\Model\Table\AccountsTable has PHPDoc tag @method for method newEntity() parameter #2 $options with no value type specified in iterable type array.`

This problem is with array type in params and with iterable type in params

`@method \App\Model\Entity\Account newEntity(array $data, array $options = [])`
This should be updated to this
`@method \App\Model\Entity\Account newEntity(array<mixed> $data, array<mixed> $options = [])`

`@method \Cake\Datasource\ResultSetInterface<\App\Model\Entity\Account>|false saveMany(iterable $entities, $options = [])`
This should be updated to this
`@method \Cake\Datasource\ResultSetInterface<\App\Model\Entity\Account>|false saveMany(iterable<mixed> $entities, $options = [])`
Or better to
`@method \Cake\Datasource\ResultSetInterface<\App\Model\Entity\Account>|false saveMany(iterable<\App\Model\Entity\Account> $entities, $options = [])`

PHPStan issue: https://github.com/phpstan/phpstan/issues/11593

Edit more issues found:
`@method \App\Model\Entity\Service[]|\Cake\Datasource\ResultSetInterface paginate($object = null, array $settings = [])`
`C:\Workplace\many\many\src\Controller\ServicesController.php:60:PHPDoc tag @method for method App\Controller\ServicesController::paginate() return type contains generic interface Cake\Datasource\ResultSetInterface but does not specify its types: T`

(Used annotations are changed by dereuromark/cakephp-ide-helper, but the issue is same)

### Bake Version

*

### PHP Version

8.3

LordSimal:
In the meantime you can just use https://github.com/dereuromark/cakephp-ide-helper to automatically generate all the necessary annotations so static analysis is happy.

dereuromark:
Both Bake and IdeHelper are currently not adding `<mixed>` to iterable here.

@Harfusha Do u want to make a PR here?

Harfusha:
And how should i implement it? Should i add <mixed> or the specific entity? I would rather add specific entity.

[...]

PR #1010, opened 2024-10-14
Make method annotations more strict

Harfusha:
https://github.com/cakephp/bake/issues/1002


dereuromark:
The original issue reported was

```
> newEntity() parameter #1 $data with no value type specified in iterable type array.
```
None of the concrete entity changes is addressing it.
The opposite: this is usually rather creating more problems for App level.

Can we just fix the mixed in iterable instead?

dereuromark:
I would expect
```
iterable<EntityInterface>
```
For input args (param) annotation 

And concrete entity for output (return) annotation

Harfusha:
Oh, so you want to implement the concete entities in IdeHelper?
Going to fix the issues.

dereuromark (dismissed):
Looks good

src/View/Helper/DocBlockHelper.php
```diff
@@ -290,20 +290,23 @@ public function buildTableAnnotations(
...
-        $annotations[] = "@method iterable<\\{$namespace}\\Model\\Entity\\{$entity}>|\Cake\Datasource\ResultSetInterface<\\{$namespace}\\Model\\Entity\\{$entity}>|false deleteMany(iterable \$entities, array \$options = [])";
-        $annotations[] = "@method iterable<\\{$namespace}\\Model\\Entity\\{$entity}>|\Cake\Datasource\ResultSetInterface<\\{$namespace}\\Model\\Entity\\{$entity}> deleteManyOrFail(iterable \$entities, array \$options = [])";
+        $annotations[] = "@method $entityPath newEmptyEntity()";
+        $annotations[] = "@method $entityPath newEntity(array<mixed> \$data, array<string, mixed> \$options = [])";
+        $annotations[] = "@method array<$entityPath> newEntities(array<mixed> \$data, array<string, mixed> \$options = [])";
+        $annotations[] = "@method $entityPath get(mixed \$primaryKey, array<mixed>|string \$finder = 'all', \\Psr\\SimpleCache\\CacheInterface|string|null \$cache = null, \Closure|string|null \$cacheKey = null, mixed ...\$args)";
```
ADmad: Passing an array for the `$finder` arg is deprecated and kept just for BC, so I would suggest just going with `string $finder`

Issue #1012, opened 2024-11-15
Error bake  while code generation


luizcmarin:
### Description

I am installing the "Authentication" plugin using the command "bin/cake bake model Users" (https://book.cakephp.org/authentication/3/en/index.html#building-a-login-action)

It returns the error "error: [Cake\Core\Exception\CakeException] Association alias `Countries` is already set. in...."

It turns out that I need these relationships. 

This is a zero-scale application with only one table created so far, "tests". So I imported all the other tables. I installed the plugin and then ran the command mentioned. There is no call anywhere for the 'Users' table.

Suppot say: "i thought you get that error after you baked the code. but it happens while code generation."

Could you please check?

[kancats.txt](https://github.com/user-attachments/files/17779543/kancats.txt)
[TestesTable.txt](https://github.com/user-attachments/files/17779545/TestesTable.txt)
[Testis.txt](https://github.com/user-attachments/files/17779544/Testis.txt)
[composer.txt](https://github.com/user-attachments/files/17779502/composer.txt)
![Image](https://github.com/user-attachments/assets/cc3865ed-65a7-43c4-92a8-be0c30d55973)
![Image](https://github.com/user-attachments/assets/48bcacd4-dacd-49d5-8a08-98082e689474)
![Image](https://github.com/user-attachments/assets/cc7622db-7f66-40fc-96d4-87fbc0562e3a)

### Bake Version

actual (see composer.txt)

### PHP Version

833

ADmad:
You have multiple foreign keys referencing the same table `countries`. That would require creating associations with a different alias than the default `Countries`. Bake is currently unable to handle such a case.

luizcmarin:
But is it possible to implement this?

dereuromark:
Yes
We could detect non default relation and prefix automatically 
That is if there is already a country relation detectable.

other_country_id becomes OtherCountries alias.
Etc

Want to try making a PR?

luizcmarin:
Of course!

How should I write this?


> Yes We could detect non default relation and prefix automatically That is if there is already a country relation detectable.
> 
> other_country_id becomes OtherCountries alias. Etc
> 
> Want to try making a PR?

luizcmarin:
> Yes We could detect non default relation and prefix automatically That is if there is already a country relation detectable.
> 
> other_country_id becomes OtherCountries alias. Etc
> 
> Want to try making a PR?


**I was unable to open a PR. Could you do it for me?
I would be very grateful.**

[...]

PR #1015, opened 2024-11-21
Make sure custom relations work out with new alias strictness.

dereuromark:
Refs https://github.com/cakephp/bake/issues/1012

So basically 5.1 and the more strict relation-handling ("Association alias `Countries` is already set") kinda broke the non documented feature of custom relationship baking.
Even though it was just baked wrongly with always the same Countries relation.

This PR adds the expected aliasing (back) ensuring there is no collision, and having the correct name in the process for most cases:
```
[
  'belongsTo' => [
    (int) 0 => [
      'alias' => 'Countries',
      'foreignKey' => 'country_id',
      'className' => 'Countries'
    ],
    (int) 1 => [
      'alias' => 'BillingCountries',
      'foreignKey' => 'billing_country_id',
      'className' => 'Countries'
    ],
    (int) 2 => [
      'alias' => 'ShippingCountries',
      'foreignKey' => 'shipping_country_id',
      'className' => 'Countries'
    ]
  ],
  'hasOne' => [],
  'hasMany' => [],
  'belongsToMany' => []
]
```
as example for

Countries
- id

Users
- id
- country_id
- billing_country_id
- shipping_country_id

### DIFF
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -377,8 +377,9 @@ class ModelCommand extends BakeCommand
                 ) {
                     $allowAliasRelations = $args && $args->getOption('skip-relation-check');
                     $found = $this->findTableReferencedBy($schema, $fieldName);
+                    $className = null;
                     if ($found) {
-                        $tmpModelName = Inflector::camelize($found);
+                        $className = ($this->plugin ? $this->plugin . '.' : '') . Inflector::camelize($found);
                     } elseif (!$allowAliasRelations) {
                         continue;
                     }
@@ -387,6 +388,9 @@ class ModelCommand extends BakeCommand
                     'alias' => $tmpModelName,
                     'foreignKey' => $fieldName,
                 ];
+                if ($className && $className !== $tmpModelName) {
+                    $assoc['className'] = $className;
+                }
                 if ($schema->getColumn($fieldName)['null'] === false) {
                     $assoc['joinType'] = 'INNER';
                 }
@@ -395,6 +399,7 @@ class ModelCommand extends BakeCommand
             if ($this->plugin && empty($assoc['className'])) {
                 $assoc['className'] = $this->plugin . '.' . $assoc['alias'];
             }
+
             $associations['belongsTo'][] = $assoc;
         }
 
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -346,13 +346,13 @@ class ModelCommand extends BakeCommand
      */
     public function findBelongsTo(Table $model, array $associations, ?Arguments $args = null): array
     {
-        $className = null;
         $schema = $model->getSchema();
         foreach ($schema->columns() as $fieldName) {
             if (!preg_match('/^.+_id$/', $fieldName) || ($schema->getPrimaryKey() === [$fieldName])) {
                 continue;
             }
 
+            $className = null;
             if ($fieldName === 'parent_id') {
                 $className = $this->plugin ? $this->plugin . '.' . $model->getAlias() : $model->getAlias();
                 $assoc = [
--- a/src/Command/ModelCommand.php
+++ b/src/Command/ModelCommand.php
@@ -256,6 +256,8 @@ class ModelCommand extends BakeCommand
         $associations = $this->findHasMany($table, $associations);
         $associations = $this->findBelongsToMany($table, $associations);
 
+        $associations = $this->ensureAliasUniqueness($associations);
+
         return $associations;
     }
 
@@ -275,8 +277,6 @@ class ModelCommand extends BakeCommand
         if (get_class($model) !== Table::class) {
             return;
         }
-
-        $this->ensureAliasUniqueness($associations);
 
         foreach ($associations as $type => $assocs) {
             foreach ($assocs as $assoc) {
@@ -1561,7 +1561,7 @@ class ModelCommand extends BakeCommand
             foreach ($associationsPerType as $k => $association) {
                 $alias = $association['alias'];
                 if (in_array($alias, $existing, true)) {
-                    $alias = $this->alias($association);
+                    $alias = $this->createAssociationAlias($association);
                 }
                 $existing[] = $alias;
                 if (empty($association['className'])) {
@@ -1580,7 +1580,7 @@ class ModelCommand extends BakeCommand
      * @param array<string, mixed> $association
      * @return string
      */
-    protected function alias(array $association): string
+    protected function createAssociationAlias(array $association): string
     {
         $foreignKey = $association['foreignKey'];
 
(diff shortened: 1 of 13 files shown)


src/Command/ModelCommand.php
```diff
@@ -1546,4 +1549,41 @@ protected function bakeEnums(Table $model, array $data, Arguments $args, Console
...
+        }
+
+        return $associations;
+    }
+
+    /**
+     * @param array<string, mixed> $association
+     * @return string
+     */
+    protected function alias(array $association): string
```
```suggestion
    protected function createAssociationAlias(array $association): string
```

`alias` is a pretty general term, I think a more specific name could be helpful for future us.

src/Command/ModelCommand.php
```diff
@@ -1546,4 +1549,41 @@ protected function bakeEnums(Table $model, array $data, Arguments $args, Console
...
+     * @return array<string, array<string, mixed>>
+     */
+    protected function ensureAliasUniqueness(array $associations): array
+    {
+        $existing = [];
+        foreach ($associations as $type => $associationsPerType) {
+            foreach ($associationsPerType as $k => $association) {
+                $alias = $association['alias'];
+                if (in_array($alias, $existing, true)) {
+                    $alias = $this->alias($association);
```
```suggestion
                    $alias = $this->createAssociationAlias($association);
```

dereuromark:
It actually works:
```
        $this->belongsTo('Countries', [
            'foreignKey' => 'country_id',
            'className' => 'Countries',
        ]);
        $this->belongsTo('BillingCountries', [
            'foreignKey' => 'billing_country_id',
            'className' => 'Countries',
        ]);
        $this->belongsTo('ShippingCountries', [
            'foreignKey' => 'shipping_country_id',
            'className' => 'Countries',
        ]);
```

Maybe someone wants to add a test case for it? E.g. @luizcmarin ?
You can also test it and confirm

dereuromark:
Turns out there is already an existing test case that so far was already invalid (exception would have happened), and now working:
https://github.com/cakephp/bake/pull/1015/commits/9631dd049a4e63e5274b8a52c757d01c7ea96057

tests/TestCase/Command/ModelCommandTest.php
```diff
@@ -602,9 +602,10 @@ public function testGetAssociationsConstraints()
                 'joinType' => 'INNER',
             ],
             [
-                'alias' => 'Users',
+                'alias' => 'Receivers',
```
ADmad, line 606: Would be nice if the aliases were always created based on the foreign key field, so `Senders` for `sender_id` too instead of `Users`.

dereuromark: It seems to be that way - until table lookup.
The problem seems to be that

                    $found = $this->findTableReferencedBy($schema, $fieldName);
                    if ($found) {
                        $tmpModelName = Inflector::camelize($found);
                    }
                    
finds the users table, and as such then switches it back to Users

dereuromark:
@ADmad I found a way to make it work.

luizcmarin:
> Maybe someone wants to add a test case for it? E.g. @luizcmarin ? You can also test it and confirm

#1018

luizcmarin:
I created a small test with some tables where errors occur. You may want to take note of this inconvenience.

[invoices.txt](https://github.com/user-attachments/files/17932993/invoices.txt)
image

dereuromark:
I wonder if thats directly related to the issue here (was working before and now not)
It rather looks like a different topic and probably should be checked out independently (e.g. after merge of this).

luizcmarin:
> I wonder if thats directly related to the issue here (was working before and now not) It rather looks like a different topic and probably should be checked out independently (e.g. after merge of this).

In fact, bake never finished generating the files due to this problem. This also happened with other tables. I isolated these attached tables for you to test, if you wish. But it happens with others too. Does this happen to you?

I'll send you the full script, in case you want to try it.

[kancats.txt](https://github.com/user-attachments/files/17933707/kancats.txt)

dereuromark:
You need to actually share the commands you use
because the SQL table itself does not provide the info to replicate

For me both "bake all Countries" and then "bake all Users" works just fine now with the upgrade here.
Same for "Proposals" and "ProposalItems" of this schema.

So whats the concrete command that breaks, and what have been generated before, or if manually adjusted, what does that file look like before running bake.

Again: I am fairly certain this is outside the scope of this specific issue here, and should not block the relation fix for the current bug in the system.

luizcmarin:
I do not rule out the possibility of errors here. Please, if everything is ok in your tests, continue. I will do new tests after updating.

The command I use is simply: " bin\cake bake all --everything". Nothing more than that.

dereuromark:
I can partically reproduce it
The issue you are having is probably due to some invalid setup in your DB schema

        $this->hasMany('Proposals', [
            'foreignKey' => 'proposal_id',
            'className' => 'Invoices',
        ]);
        
This wrongly is a proposal_id, but should be something else at least to make clear this is from invoices table referenced, and also must always be different from the actual proposal_id since Proposals is their own table.
So you are creating on purpose a setup that bake everything cannot solve for you.
You need to fix your table setup here, e.g. use `invoice_proposal_id` or alike to make it unique.

luizcmarin:
Thank you very much for the tip. This means more work for me and less for you :) :(...
I'll review my scheme. I think this solves everything.

dereuromark:
Well, https://github.com/cakephp/bake/pull/1019 will make it easier for you once merged :)

dereuromark:
All good? Can we squash merge?

src/Command/ModelCommand.php
```diff
@@ -1546,4 +1554,43 @@ protected function bakeEnums(Table $model, array $data, Arguments $args, Console
             $enumCommand->execute($args, $io);
         }
     }
+
+    /**
+     * @param array<string, array<string, mixed>> $associations
+     * @return array<string, array<string, mixed>>
+     */
+    protected function ensureAliasUniqueness(array $associations): array
```
ADmad, line 1562:
```suggestion
    protected function ensureUniqueAlias(array $associations): array
```
This rolls a bit easier off the tongue :)

ADmad (commented):
Looks good apart from the minor nitpick of function name.

src/Command/ModelCommand.php
dereuromark: But it does it for more than one, so we could debate `ensureUniqueAliases()` etc.

dereuromark: I must say I kinda like the naming ensureAliasUniqueness() that @markstory proposed.

ADmad: Okay, we can keep the same then.

dereuromark: Oh createAssociationAlias() was from him, not this one.
Either way, if all are fine, we can keep it.

PR #1020, opened 2024-11-28
ignore shadow translations tables *_translations

### LINKED ISSUE #1011: Skip table class generation for `*_translations` table.
### Description

The `TableScanner` class needs to be updated to ignore shadow translations tables `*_translations`, we don't to generate table classes for them.

dereuromark:
Resolves https://github.com/cakephp/bake/issues/1011

We probably want to document that we cannot bake any actual tables anymore that would end this way.
I have for example
https://github.com/dereuromark/cakephp-translate/blob/734d8610e6de7b81878052546890a55662e52eda/config/Migrations/20170414190437_TranslateCache.php#L11
which would then not work anymore.

Is there a way we could be more specific maybe?
Intospecting the table schema to see if it matches a shadow table schema might be overkill, no?

### DIFF
--- a/src/Command/AllCommand.php
+++ b/src/Command/AllCommand.php
@@ -84,17 +84,17 @@ class AllCommand extends BakeCommand
         /** @var \Cake\Database\Connection $connection */
         $connection = ConnectionManager::get($this->connection);
         $scanner = new TableScanner($connection);
+        $tables = $scanner->removeShadowTranslationTables($scanner->listUnskipped());
+
         if (!$name && !$args->getOption('everything')) {
             $io->out('Choose a table to generate from the following:');
-            foreach ($scanner->listUnskipped() as $table) {
+            foreach ($tables as $table) {
                 $io->out('- ' . $this->_camelize($table));
             }
 
             return static::CODE_SUCCESS;
         }
-        if ($args->getOption('everything')) {
-            $tables = $scanner->listUnskipped();
-        } else {
+        if (!$args->getOption('everything')) {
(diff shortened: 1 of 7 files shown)


ADmad:
> Intospecting the table schema to see if it matches a shadow table schema might be overkill, no?

It would be a foolproof way though as shadow tables have a multicolumn PK with a `locale` column.

Another simpler option would be to check if a matching table without the `_translations` suffix also exists.

dereuromark:
Either way, then the regexp and $ignore seems not to cover it anymore here.
I made its own method now.

ADmad (approved):

ADmad (changes requested):
I just realized something. Skipping the shadow tables should be done in the `ModelAllCommand`, not the `TableScanner`.

I have my own customized model command where I check if the tables list returned by the `TableScanner` has a matching shadow table for a table and add the `Translate` behavior for it. So removing these tables from the tables list itself will break that feature.

dereuromark:
Better?

dereuromark:
However, it would still bake
- Table
- Entity
- Controller
- ...

when running through "bake all --everything" etc
so this then needs to be duplicated across mutliple ones :/

ADmad (approved):

ADmad:
In that case keep the `removeShadowTranslationTables()` in `TableScanner` itself but call in from `ModelAllCommand`, `ControllerAllCommand` and `AllCommand`?

dereuromark:
Now it works properly :tada:

PR #1025, opened 2024-12-28
3.x: remove unnecessary parent call in table class

LordSimal:
There is no need to call `parent::initialize($config);` in a baked Table class since [we don't actually do anything in there.](https://github.com/cakephp/cakephp/blob/5.x/src/ORM/Table.php#L390-L392)

### DIFF
--- a/templates/bake/Model/table.twig
+++ b/templates/bake/Model/table.twig
@@ -41,8 +41,6 @@ class {{ name }}Table extends Table{{ fileBuilder.classBuilder.implements ? ' im
      */
     public function initialize(array $config): void
     {
-        parent::initialize($config);
-
 {% if table %}
         $this->setTable('{{ table }}');
 {% endif %}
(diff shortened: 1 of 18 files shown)


dereuromark:
But if we, like me, use a different base class after baking you suddenly need it.
I would be rather having it consistently call it and have it no-op by default than having to wry about adding it.

LordSimal:
> But if we, like me, use a different base class after baking you suddenly need it.
> I would be rather having it consistently call it and have it no-op by default than having to wry about adding it.

If you introduce your own inheritance logic then I'd say its logical to add that parent call yourself, not rely on the Framework to do the thinking for you.

I think its just misleading that sometimes the `parent` call does something for our code (like `Application::bootstrap`) and sometimes it doesn't because "maybe you need it later"

ADmad:
I too would prefer keeping the parent call for consistency. We have it for the controller class too.

PR #1028, opened 2025-01-17
Set the pseudo method to "delete" for "Delete" links.

ADmad:
This sets the `_method` hidden field in the post link form to `delete`.

### DIFF
--- a/templates/bake/Template/index.twig
+++ b/templates/bake/Template/index.twig
@@ -65,7 +65,14 @@
                     <td class="actions">
                         <?= $this->Html->link(__('View'), ['action' => 'view', {{ pk|raw }}]) ?>
                         <?= $this->Html->link(__('Edit'), ['action' => 'edit', {{ pk|raw }}]) ?>
-                        <?= $this->Form->postLink(__('Delete'), ['action' => 'delete', {{ pk|raw }}], ['confirm' => __('Are you sure you want to delete # {0}?', {{ pk|raw }})]) ?>
+                        <?= $this->Form->postLink(
+                            __('Delete'),
+                            ['action' => 'delete', {{ pk|raw }}],
+                            [
+                                'method' => 'delete',
+                                'confirm' => __('Are you sure you want to delete # {0}?', {{ pk|raw }}),
+                            ]
+                        ) ?>
                     </td>
                 </tr>
                 <?php endforeach; ?>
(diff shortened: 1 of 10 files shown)


ADmad:
Should we target this to `3.next`?

LordSimal:
if so, then we should change the generated delete method in the controller to only allow the DELETE HTTP Method as well.

ADmad:
It's very easy to miss specifying the `method` as evident here. I think it is quite likely users might be using their own templates and the baked controller code. So I would keep the controller template unchanged.

Allowing `post` in `delete()` won't cause any problem in majority of the apps. I made this change mainly so that new users can be made aware of the delete simulation.

templates/bake/Template/index.twig
```diff
@@ -65,7 +65,14 @@
                     <td class="actions">
                         <?= $this->Html->link(__('View'), ['action' => 'view', {{ pk|raw }}]) ?>
                         <?= $this->Html->link(__('Edit'), ['action' => 'edit', {{ pk|raw }}]) ?>
-                        <?= $this->Form->postLink(__('Delete'), ['action' => 'delete', {{ pk|raw }}], ['confirm' => __('Are you sure you want to delete # {0}?', {{ pk|raw }})]) ?>
+                        <?= $this->Form->postLink(
```
dereuromark, line 68: Wouldnt deleteLink() be cleaner to avoid the assoc array config overhead?

dereuromark: Been using this for the last decade in my Form helper(s):
```php
public function deleteLink(string|HtmlStringable $title, array|string|null $url = null, array $options = []): string {
	$options['method'] = 'delete';
	if (!isset($options['class'])) {
		$options['class'] = 'delete-link';
	}

	return $this->postLink($title, $url, $options);
}
```
I tend to avoid too much array config stuff in the templates themselves, and prefer speaking methods.

markstory: We don't have a `deleteLink` method. We could add one though.

markstory (approved):

templates/bake/Template/index.twig
ADmad: I too was thinking of adding `deleteLink()`.

dereuromark: https://github.com/cakephp/cakephp/pull/18144

PR #1037, opened 2025-04-16
Fix up getDisplayField() in relation to Cake 6.0

dereuromark:
Given that Bake is a debug/local tool only, and that this method would rarely be overwritten/extended, this should be safe to do in a satellite plugin like this.
Refs https://github.com/cakephp/cakephp/pull/18338

It is impossible to be null anyway.


markstory:
What's going on with the failing tests?

dereuromark:
Thats us waiting for your core Patch:)

Issue #1038, opened 2025-04-17
Add support for __d() when baking code in a plugin context using -p option


AvelinoJavier:
#### Description

When using `bake` to generate code within a plugin using the `-p` (or `--plugin`) option, all translatable strings are currently generated with the `__()` function, regardless of the plugin context.

Example:

```bash
bin/cake bake all Posts -p MyPlugin
```

Generates code like:
__('Edit Post')

However, the recommended approach for translations inside plugins—as per CakePHP documentation—is to use __d('plugin_name', ...) with the plugin name as the translation domain. So the expected output should be:
__d('my_plugin', 'Edit Post')

#### Why this matters
It aligns with CakePHP’s i18n best practices.

It avoids the need to manually replace all __() calls with __d() after baking.

It improves plugin encapsulation by keeping translations scoped to the plugin.

#### Proposed behavior
When the -p or --plugin option is used with bake, the generated templates, controllers, and views should automatically use:
__d('{plugin_name}', 'Text')

#### Optional enhancement
Allow a new flag like --i18n-domain=plugin or --use-domain to make this behavior explicit and flexible.

Thanks for your consideration!

dereuromark:
@AvelinoJavier Do you want to make a PR here?
--use-domain opt int flag sounds reasonable.

dereuromark:
What do you think of https://github.com/cakephp/bake/pull/1039 ?
Maybe there is a better solution even?

PR #1039, opened 2025-06-07
Allow using __d() in plugin context.

dereuromark:
Possible soution to https://github.com/cakephp/bake/issues/1038

I wonder if we could make a helper wrapper for translations, to keep the templates more readable?


LordSimal:
The base idea is pretty cool, but I don't see any reason why one should use `__()` inside a plugin (so what we currently have 😁)

I'd rather always default to `__d()` if `-p` aka a Plugin file is being generated.

dereuromark:
That would simplify it a bit I guess :)

src/Command/ControllerCommand.php
```diff
@@ -112,14 +112,15 @@ public function bake(string $controllerName, Arguments $args, ConsoleIo $io): vo
 
         $currentModelName = $controllerName;
         $plugin = $this->plugin;
-        if ($plugin) {
-            $plugin .= '.';
+        $pluginPath = $plugin;
```
dereuromark, line 115: In the process of simplifying I also found a bug: A dot added to plugin is passed along to the templates.
This is now also fixed as a side effect.

PR #1041, opened 2025-06-22
update stan

LordSimal:


LordSimal:
something weird is going on in php 8.3 with sqlite here...

markstory:
That could be a regression in schema reflection, let me take a look.

PR #1045, opened 2025-07-09
Fix up wrapper usage for io.

dereuromark:
I noticed that IO is not used via wrappers, probably legacy wise.
Also some errors are written into stdout instead of stderr

I propose to consolidate this using the wrappers as intended to be used, also simplifies the code.

Added PHPUnit 12


markstory (approved):
Looks good once the stan error is addressed.

dereuromark:
Done.

PR #1046, opened 2025-07-14
Use internal table locator with fallback classes

nicosp:


src/Command/BakeCommand.php
```diff
@@ -67,6 +68,18 @@ public static function defaultName(): string
...
 
+    /**
+     * @inheritDoc
+     */
+    public function initialize(): void
+    {
+        parent::initialize();
+        // Use our own table locator with fallback classes
+        $locator = new TableLocator();
+        $locator->allowFallbackClass(true);
```
dereuromark: I wonder, could we check if there is already one locator defined that has fallback activated?
Then we could skip adding another one.

nicosp: We don't have a way to check unless I am missing something.

nicosp: nvm. I figured it out. Not ideal but it works for now.

dereuromark: Yeah, this looks like the safest BC way once we remove it from the app skeleton.

ADmad:
Why the hacky try/catch and new instance creation instead of a single line `$this->getTableLocator()->allowFallbackClass(true);` ?

dereuromark:
If we remove the fallback in app skeleton, we might still have existing apps already declare it.
So declaring it twice seems overkill, no?

ADmad:
How is simply resetting a flag, even if already set, an overall, instead of recreating a locator instance?

ADmad:
Also replacing the existing locator instance with a new one would lose and customization/config. For e.g. one can set additional locations for the location using the `$locations` constructor argument or a custom `QueryFactory`.

dereuromark:
Yeah there should be no replacement. Only adding if not there yet

dereuromark:
@ADmad you might have misread. It does not replace.

ADmad:
> you might have misread. It does not replace.

```
            $locator = new TableLocator();
            $locator->allowFallbackClass(true);
            $this->setTableLocator($locator);
```

These lines do replace the locator instance. All that needs to be done is toggle on the fallback behavior.

ADmad:
Closing in favor of #1047

PR #1047, opened 2025-07-15
Ensure fallback class is enabled for the table locator.

ADmad:
It's recommended to disable fallback class during normal functioning of the app but fallback needs to be enabled when baking as the required table classes might still not have been created.


ADmad:
Ideally enabling fallback should only be required for the `ModelCommand` but that would generate errors if for e.g. someone tries to generate a controller before generating it's model class. So the safest option is to enable it for all commands.

Issue #998, opened 2024-07-24
Plugin generation shouldn't include composer.json & phpunit.xml by default


markstory:
### Description

The default behavior of `bake plugin` is to create an application plugin. However, `plugin bake` also generates `phpunit.xml.dist` and `composer.json` which are redundant for in-app plugins.

Adding a `--standalone` option to `bake plugin` would allow us to skip generating composer+phpunit configuration for in-app plugins but retain that behavior for standalone plugins.

Related to #991

dereuromark:
The Alternative would be to add --vendor to put it into a vendor folder directly and adding those files. And otherwise skip them.
The flag --standalone could also be understood as vendor standalone compared to plugin and App related and might be a bit unclear.

LordSimal:
Wit the addition of https://github.com/cakephp/plugin-installer/pull/64 the composer.json is NOT only needed for publishing the plugin on packagist but it instead moves the autoload namespace to the plugin itself.

Therefore removing it from the default template is not needed anymore.

`phpunit.xml.dist` on the other hand could easily be removed imho. 
especially since the only parts special inside that file are the testsuite definition and the PHPUnitExtension for the fixtures.

ADmad:
> ..but it instead moves the autoload namespace to the plugin itself.

No it doesn't. That change is related to composer config of the app, it has nothing to do with the composer.json within the plugin itself.

So generating composer.json is still redundant for in-app plugins.

LordSimal:
I thought you used the namespaces configured inside those composer.json but looking at that PR again it indeed just uses the folder name inside `plugins` as the root namespace for that plugin.

With that I can agree, that the composer.json is actually doing nothing out of the box.

As long as that plugin name doesn't colide with another vendor namespace this seems fine by me, but it would still be a bit more configurable to actually define the plugin namesapce inside the composer.json of th plugin.

PR #1054, opened 2025-09-03
fix composer.json not being generated for private plugins

LordSimal:
I'd love to see users move the namespace definition of their private plugins inside the plugin where its actually used, not in the root composer.json like in the old days.

This is supported since CakePHP 5 / Plugin Installver v2 as you can see [here](https://github.com/cakephp/plugin-installer/pull/64)

Currently when baking a plugin it doesn't generate a composer.json AND doesn't adjust the composer.json of the root app, so the generated namespace in the plugin is actually not hooked up at all.

### DIFF
--- a/src/Command/PluginCommand.php
+++ b/src/Command/PluginCommand.php
@@ -238,7 +238,7 @@ class PluginCommand extends BakeCommand
 
                 if (!$this->isVendor) {
                     $vendorFiles = [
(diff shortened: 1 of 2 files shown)


LordSimal:
I am stupid and forgot about this: https://github.com/cakephp/bake/issues/998#issuecomment-2395427797

PR #1060, opened 2025-09-29
Allow bake update to keep custom column types.

### LINKED ISSUE #1058: Improve --update to retain column schema overrides
### Description

I have a table that re-maps a column to an enum type.

```php
    /**
     * Initialize method
     *
     * @param array<string, mixed> $config The configuration for the Table.
     * @return void
     */
    public function initialize(array $config): void
    {
        parent::initialize($config);

        $this->setTable('issue_activities');

        $this->setDisplayField('activity_type');

        $this->setPrimaryKey('id');

        $this->addBehavior('Timestamp');

        $this->belongsTo('Issues', [
            'foreignKey' => 'issue_id',
            'joinType' => 'INNER',
        ]);
        $this->getSchema()->setColumnType('activity_type', EnumType::from(IssueActivityTypes::class));
    }
```

If I use `bin/cake model --no-test --no-fixture --update IssueActivities` the enum mapping will be removed. Ideally we can retain custom schema mappings. I recognize this is a complicated change and I see this as an improvement.

dereuromark:
Might resolve https://github.com/cakephp/bake/issues/1058

@markstory Maybe you can test it.

### DIFF
--- a/src/CodeGen/ColumnTypeExtractor.php
+++ b/src/CodeGen/ColumnTypeExtractor.php
@@ -1,0 +1,225 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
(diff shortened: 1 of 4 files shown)


markstory:
Thanks! This worked for me.

PR #1061, opened 2025-10-22
Bump twig-view constraint

josbeir:
Bumping twig-view constraint from 2.0.0 to 2.0.2 as it requires a higher twig version.

Fixes failing twig compilation tests on prefer-lowest builds.

reference: https://github.com/cakephp/queue/pull/174


josbeir:
@dereuromark or @markstory can we do a 3.5.1 release?

ADmad:
> can we do a 3.5.1 release?

Done

PR #1069, opened 2026-01-11
Fix deprecations from 5.3

markstory:


dereuromark:
I tackled most of it in my PR at the same time it seems.
You can pull out some of it in this one here if it helps.

I can then simplify mine again.

PR #1070, opened 2026-02-04
Fix checking references int non-foreign constraints

othercorey:
Non-foreign constraints such as a check don't return `columns`. This simply defaults the columns array to fail the check.


othercorey:
@markstory I don't know if this changed with the new schema classes

markstory:
@othercorey I suspect this is related to the introduction of check constraints like you said.

othercorey:
> @othercorey I suspect this is related to the introduction of check constraints like you said.

Could you do a patch release with this? I keep having to patch vendor when I run update.

LordSimal:
Done https://github.com/cakephp/bake/releases/tag/3.6.1

PR #1073, opened 2026-02-24
Bump up phpstan to level 8

ADmad:


src/Utility/SubsetSchemaCollection.php
```diff
@@ -33,7 +33,7 @@ class SubsetSchemaCollection implements CollectionInterface
     protected CollectionInterface $collection;
 
     /**
-     * @var array<string>
+     * @var list<string>
```
dereuromark: I think we mainly avoid list<> in general, as it would also complain about writing to this dynamically from the code in many cases.

dereuromark (approved):

PR #1077, opened 2026-03-29
add rector

LordSimal:


src/CodeGen/ColumnTypeExtractor.php
```diff
@@ -136,37 +130,25 @@ public function leaveNode(Node $node)
...
-
-                    // Get the type expression as a string
-                    $typeExpression = $this->getTypeExpression($typeArg);
-                    if ($typeExpression !== null) {
-                        $this->columnTypes[$columnName] = $typeExpression;
-                    }
-                }
+        // Check if it's called on getSchema()
+        // Extract the column name and type expression
+        if ($methodCall->name instanceof Node\Identifier && $methodCall->name->name === 'setColumnType' && ($methodCall->var instanceof MethodCall && $methodCall->var->name instanceof Node\Identifier && $methodCall->var->name->name === 'getSchema' && $methodCall->var->var instanceof Variable && $methodCall->var->var->name === 'this') && count($methodCall->args) >= 2) {
```
LordSimal: WTF is this....

src/View/Helper/DocBlockHelper.php
```diff
@@ -288,29 +288,29 @@ public function buildTableAnnotations(
...
+        $annotations[] = sprintf('@method \%s\Model\Entity\%s saveOrFail(\Cake\Datasource\EntityInterface $entity, array $options = [])', $namespace, $entity);
+        $annotations[] = sprintf('@method iterable<\%s\Model\Entity\%s>|\Cake\Datasource\ResultSetInterface<\%s\Model\Entity\%s>|false saveMany(iterable $entities, array $options = [])', $namespace, $entity, $namespace, $entity);
+        $annotations[] = sprintf('@method iterable<\%s\Model\Entity\%s>|\Cake\Datasource\ResultSetInterface<\%s\Model\Entity\%s> saveManyOrFail(iterable $entities, array $options = [])', $namespace, $entity, $namespace, $entity);
+        $annotations[] = sprintf('@method iterable<\%s\Model\Entity\%s>|\Cake\Datasource\ResultSetInterface<\%s\Model\Entity\%s>|false deleteMany(iterable $entities, array $options = [])', $namespace, $entity, $namespace, $entity);
+        $annotations[] = sprintf('@method iterable<\%s\Model\Entity\%s>|\Cake\Datasource\ResultSetInterface<\%s\Model\Entity\%s> deleteManyOrFail(iterable $entities, array $options = [])', $namespace, $entity, $namespace, $entity);
```
LordSimal: should we disable the sprintf rule for this repo?

src/CodeGen/ColumnTypeExtractor.php
LordSimal: I have now refactored this method to use the early return pattern which is common for AST based logic

src/Utility/TemplateRenderer.php
```diff
@@ -64,7 +60,7 @@ public function __construct(?string $theme = null)
      */
     public function getView(): View
     {
-        if ($this->view) {
+        if ($this->view instanceof BakeView) {
```
dereuromark, line 63: I wonder why rector wouldnt just do the faster !== null here.
Since it is technically clear that it can only be that object or null.

dereuromark: I also would find the current way `if ($this->view)` "clean", since we always shortcut the bool way with if checks if its clear about the scope (no ambiguity).

dereuromark (approved):

src/Utility/TemplateRenderer.php
LordSimal:
```
Applied rules:
 * FlipTypeControlToUseExclusiveTypeRector
 * NullableCompareToNullRector
 ```

File: src/Command/AllCommand.php [hunk 4]
```diff
@@ -45,7 +45,7 @@ class AllCommand extends BakeCommand
      * @param \Cake\Console\ConsoleOptionParser $parser Option parser to update.
      * @return \Cake\Console\ConsoleOptionParser
      */
-    public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
+    protected function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
```
dereuromark, line 48: Note: Changing the signature here in a patch (3.6.4) creates a potential BC break downstream.
For PHPStan requires e.g. https://github.com/cakephp/authorization/pull/333 to have the same fixed in a patch, and without require constraint (only require-dev), this is not enforcable.

~~We are forcing users to use latest of everything here. Probably fair enough for now for "dev only" tooling.~~
We should probably just silence the PHPStan issue downstream to avoid any fallout.

PR #1081, opened 2026-05-02
remove defaultName method from baked command

LordSimal:
It was added in https://github.com/cakephp/bake/pull/1030 but its not really needed, as it gets statically inflected by the classname anyways.

This makes the user generated code clearer as there is only 1 place where the command name is defined, not 2.


LordSimal:
According to @ADmad the difference between those 2 is as follows:

> the static method provides the default name used for command list without having to instantiate the command. The prop holds the name used when explicitly adding commands using CommandCollection::add() from App's console() hook

